PR #1, opened 2016-11-09
Write repmgr conf and pgpass when configuring a primary database server

carbonin:
When a primary server fails and a failover is executed it is typical to reintroduce the failed primary server as a standby if the failure is recoverable.

Some of the configuration edits made when configuring a standby also should be done when configuring a primary server to deal with this situation.

These config changes are safe to make when the server is running as a primary and will allow us to reintroduce the server as a standby without having to make the changes manually.

@gtanzillo @jvlcek please review

Was https://github.com/ManageIQ/manageiq/pull/12422

### DIFF
--- a/lib/gems/pending/appliance_console/database_replication.rb
+++ b/lib/gems/pending/appliance_console/database_replication.rb
@@ -8,6 +8,7 @@ module ApplianceConsole
     include ApplianceConsole::Logging
 
     REPMGR_CONFIG = '/etc/repmgr.conf'.freeze
+    REPMGR_LOG    = '/var/log/repmgr/repmgrd.log'.freeze
 
     attr_accessor :cluster_name, :node_number, :database_name, :database_user,
                   :database_password, :primary_host
@@ -52,6 +53,10 @@ Replication Server Configuration
         f.puts("conninfo='host=#{host} user=#{database_user} dbname=#{database_name}'")
         f.puts("use_replication_slots=1")
         f.puts("pg_basebackup_options='--xlog-method=stream'")
+        f.puts("failover=automatic\n")
+        f.puts("promote_command='repmgr standby promote'\n")
+        f.puts("follow_command='repmgr standby follow'\n")
+        f.puts("logfile=#{REPMGR_LOG}\n")
       end
       true
     end
--- a/lib/gems/pending/appliance_console/database_replication.rb
+++ b/lib/gems/pending/appliance_console/database_replication.rb
@@ -47,19 +47,23 @@ Replication Server Configuration
     end
 
     def create_config_file(host)
-      File.open(REPMGR_CONFIG, "w") do |f|
-        f.puts("cluster=#{cluster_name}")
-        f.puts("node=#{node_number}")
-        f.puts("node_name=#{host}")
-        f.puts("conninfo='host=#{host} user=#{database_user} dbname=#{database_name}'")
-        f.puts("use_replication_slots=1")
-        f.puts("pg_basebackup_options='--xlog-method=stream'")
-        f.puts("failover=automatic\n")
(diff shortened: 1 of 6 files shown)


carbonin:
Also ....

FIRST!!!1!!1

lib/gems/pending/appliance_console/database_replication.rb
```diff
@@ -70,6 +76,16 @@ def generate_cluster_name
...
     end
 
+    def write_pgpass_file
+      File.open(PGPASS_FILE, "w") do |f|
+        f.write("*:*:#{database_name}:#{database_user}:#{database_password}\n")
+        f.write("*:*:replication:#{database_user}:#{database_password}\n")
+      end
+
+      FileUtils.chmod(0600, PGPASS_FILE)
+      FileUtils.chown("postgres", "postgres", PGPASS_FILE)
```
jvlcek: @carbonin Looks really good.

Just one Suggestion:
Although this will work it might be best to return true here. It would avoid having a string literal in the condition at the caller.

Fryguy:
Now that Travis is enabled on the repo, please rebase.

Fryguy:
Please rebase now that the repo has Travis support.

lib/gems/pending/appliance_console/database_replication.rb
```diff
@@ -52,6 +53,10 @@ def create_config_file(host)
         f.puts("conninfo='host=#{host} user=#{database_user} dbname=#{database_name}'")
         f.puts("use_replication_slots=1")
         f.puts("pg_basebackup_options='--xlog-method=stream'")
+        f.puts("failover=automatic\n")
```
bdunne: Should the content generation be moved into a method then just `File.write` that?  Then you can just test that the method generates the correct content and not worry about a test that writes a file.

jvlcek: 👍  I like this idea. Wish I would have thought of it. ;) I could have used this approach on some of my appliance_console PRs.

spec/appliance_console/database_replication_spec.rb
```diff
@@ -115,4 +115,33 @@
       expect(subject.generate_cluster_name).to be_falsey
     end
   end
+
+  context "#write_pgpass_file" do
+    let(:pgpass_path) { Tempfile.new("pgpass").path }
+
+    before do
+      stub_const("#{described_class}::PGPASS_FILE", pgpass_path)
```
bdunne: You could do the same here as my other suggestion to avoid writing a file for tests.

carbonin: Ah I misread your comment before. I only changed the method, not the spec. Let me fix this up ...

carbonin: Okay, I changed how we test the repmgr.conf file, but I kind of like testing this one as it is because it allows us to test the permission changes in a nice way.

I would rather lay a file down and test the permissions with a stat than stub those calls.

carbonin:
Created https://bugzilla.redhat.com/show_bug.cgi?id=1450492 for euwe backport

simaishi:

```
$ git log -1
commit 73909f2f71722ab102f4b60d1f6c6a7fa44a7593
Author: Gregg Tanzillo <gtanzill@redhat.com>
Date:   Mon Dec 5 15:29:08 2016 -0500

    Merge pull request #1 from carbonin/move_some_ha_config_to_base_class
    
    Write repmgr conf and pgpass when configuring a primary database server
    (cherry picked from commit 3f676e09f62766050302ebd27c6e2c623e58bfbc)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1450508
```

PR #2, opened 2016-11-09
Add some error reporting when we start evmserverd

carbonin:
This will hopefully help pin down #10300

Because we fork to run the systemd commands, it may be that some errors are not getting reported anywhere. This change will make sure we can see any errors in the `appliance_console.log`

@jvlcek please review

Was: https://github.com/ManageIQ/manageiq/pull/12202


jvlcek:
@carbonin 👍 The only thing I could suggest is updating the spec to test messages are logged if LinuxAdmin::Service.new("evmserverd") raises an error

PR #9, opened 2016-11-14
Add Gnocchi metric service to Openstack handle

aufi:
In order to support metrics in OSP10, we need add support methods for
Gnocchi service.

This PR contains backend methods for https://github.com/ManageIQ/manageiq/pull/12396


tzumainn:
Do you have to update manageiq-gems-pending.gemspec?

s.add_runtime_dependency "fog-openstack",           "=0.1.15"

tzumainn:
Duh, I see it now, sorry!

tzumainn:
:+1: looks good to me!

blomquisg:
Ugh, I was gonna ask why we would have a `metric` *and* a `metering` delegate, but it looks like that's what OpenStack is actually calling the Gnocchi and Ceilometer services, respectively.

aufi:
@blomquisg Yeah, Metric in a new service in Openstack (but was extracted from Metering), for some time we need support both

blomquisg:
@aufi gotcha

PR #10, opened 2016-11-15
Use KEY_ROOT over RAILS_ROOT for the certs

carbonin:
This is defined in our environment, is more applicable, and should be more reliable than RAILS_ROOT.

Currently this is failing because rails root isn't defined and the file is no longer in the manageiq code tree.
This means that `Pathname.new(__dir__).join("../../..")` is no longer valid.

@Fryguy please review.


bdunne:
:+1: LGTM

PR #15, opened 2016-11-16
Update azure-armrest gem to 0.3.12

djberg96:
This updates the azure-armrest gem to 0.3.12 and replaces "=" with "~>".


manageiq-gems-pending.gemspec
```diff
@@ -23,7 +23,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "activesupport",           "~> 5.0.0"
   s.add_runtime_dependency "addressable",             "~> 2.4"
   s.add_runtime_dependency "awesome_spawn",           "~> 1.4"
-  s.add_runtime_dependency "azure-armrest",           "=0.3.9"
+  s.add_runtime_dependency "azure-armrest",           ">= 0.3.12"
```
djberg96: @bdunne Originally I was going to do that. I keep flip-flopping on how I want to do it. I'll change it back.

PR #17, opened 2016-11-17
Remove forced GC.start from every API call

agrare:
Average timings for `collect_inventory_for_targets` over 5 runs:
`With GC.start: 62.57s`
`Without GC.start: 38.06s`

So 38% less time to collect inventory.

The steady state memory usage (after the first full refresh completes) was:
`With GC.start: 549MB`
`Without GC.start: 632MB`

So an increase of 15% memory usage.

https://bugzilla.redhat.com/show_bug.cgi?id=1378998
https://bugzilla.redhat.com/show_bug.cgi?id=1341402
https://bugzilla.redhat.com/show_bug.cgi?id=1353707

### DIFF
--- a/lib/gems/pending/VMwareWebService/VimService.rb
+++ b/lib/gems/pending/VMwareWebService/VimService.rb
@@ -1254,11 +1254,11 @@ class VimService < Handsoap::Service
   end
 
   def handle_memory_and_gc(response)
+    xml_len = response.instance_variable_get(:@http_body).length
+
     # At this point we don't need the internal raw XML content since we
     #   have the parsed XML document, so we can free it for the GC.
     response.instance_variable_set(:@http_body, nil)
-
-    xml_len = response.document.to_s.length
 
     @xml_payload_lock.synchronize do
       @xml_payload_len += xml_len
--- a/lib/gems/pending/VMwareWebService/VimService.rb
+++ b/lib/gems/pending/VMwareWebService/VimService.rb
@@ -14,6 +14,10 @@ class VimService < Handsoap::Service
     @serviceInstanceMor = VimString.new("ServiceInstance", "ServiceInstance")
     @session_cookie     = nil
 
+    @xml_payload_len  = 0
+    @xml_payload_max  = 10.megabytes
+    @xml_payload_lock = Mutex.new
+
     @sic = retrieveServiceContent
 
     @about           = @sic.about
@@ -1249,16 +1253,32 @@ class VimService < Handsoap::Service
     end
   end
 
+  def handle_memory_and_gc(response)
+    # At this point we don't need the internal raw XML content since we
+    #   have the parsed XML document, so we can free it for the GC.
+    response.instance_variable_set(:@http_body, nil)
+
+    xml_len = response.document.to_s.length
+
+    @xml_payload_lock.synchronize do
+      @xml_payload_len += xml_len
+
+      if @xml_payload_len > @xml_payload_max
+        @xml_payload_len = 0
+
+        # Force a GC, because Ruby's GC is triggered on number of objects without
+        #   regard to size.  The object we just freed may not be released right away.
+        GC.start
+      end
+    end
+  end
+
   def parse_response(response, rType)
     doc  = response.document
     raise "Response <#{response.inspect}> has no XML document" if doc.nil?
 
-    # At this point we don't need the internal raw XML content since we
-    #   have the parsed XML document, so we can free it for the GC.
-    response.instance_variable_set(:@http_body, nil)
-    # Force a GC, because Ruby's GC is triggered on number of objects without
-    #   regard to size.  The object we just freed may not be released right away.
-    GC.start
+    # Cleanup the raw XML document from the response and kick a GC
+    handle_memory_and_gc(response)
 
     search_path = "//n1:#{rType}"
     node = doc.xpath(search_path, @ns).first


agrare:
cc @Fryguy @roliveri 

Without a clear winner I'm leaning towards making this configurable, the customer in https://bugzilla.redhat.com/show_bug.cgi?id=1378998 cannot do a full refresh in any reasonable time without removing the `GC.start` and they're okay with a bit more memory usage but I'm not sure everyone would make the same choice...thoughts?

Fryguy:
@agrare I'm not a fan of the configuration option if we can avoid one.  Can we discuss?  I feel like there's an answer here.

agrare:
@Fryguy definitely, what did you have in mind?

lib/gems/pending/VMwareWebService/VimService.rb
```diff
@@ -1249,16 +1253,32 @@ def marshalObj(node, obj)
     end
   end
 
+  def handle_memory_and_gc(response)
+    # At this point we don't need the internal raw XML content since we
+    #   have the parsed XML document, so we can free it for the GC.
+    response.instance_variable_set(:@http_body, nil)
+
+    xml_len = response.document.to_s.length
```
Fryguy: Does `response.document` just return an internal reference, or does it actually create something?  If the latter, we may want to pass it in because it's being created on line 1277 as well.

Fryguy: Perhaps instead of .to_s on the document, we can use the `@http_body` directly...

```ruby
  def handle_memory_and_gc(response)
    xml_len = response.instance_variable_get(:@http_body).length

    # At this point we don't need the internal raw XML content since we
    #   have the parsed XML document, so we can free it for the GC.
    response.instance_variable_set(:@http_body, nil)
```

lib/gems/pending/VMwareWebService/VimService.rb
```diff
@@ -1249,16 +1253,32 @@ def marshalObj(node, obj)
     end
   end
 
-  def parse_response(response, rType)
-    doc  = response.document
-    raise "Response <#{response.inspect}> has no XML document" if doc.nil?
+  def handle_memory_and_gc(response)
```
Fryguy: Minor, but just ensure this is a private method.

Fryguy:
@agrare This looks good to me...what are you measurements now?

agrare:
@Fryguy average of `38.5 sec` so a negligible increase over no GC
```
39.66774034500122
38.96231555938721
37.94510054588318
37.87278461456299
38.10516047477722
Average: 38.510620307922366
```

lib/gems/pending/VMwareWebService/VimService.rb
agrare: It is under https://github.com/agrare/manageiq-gems-pending/blob/b69d853254d43b4ccd771df9568c1e1fe582a980/lib/gems/pending/VMwareWebService/VimService.rb#L1218, https://github.com/agrare/manageiq-gems-pending/blob/b69d853254d43b4ccd771df9568c1e1fe582a980/lib/gems/pending/VMwareWebService/VimService.rb#L1256

Fryguy:
How's the steady state memory usage?

agrare:
So at 200MB threshold its the same as no GC because I never actually hit it.
Playing around with some other limits now.

agrare:
For reference this VC has 2048 VMs and 64 Hosts

agrare:
So having the threshold at 10MB balances the memory usage in the middle, 601MB but I'm only getting 20MB total XML back so I'm not sure how much we want to tune for this specific VC

agrare:
@Fryguy a full refresh of a large VC (8192 VMs) received 87MB of XML docs, what do you think of 50MB as the GC threshold?

Fryguy:
Sounds good to me.

agrare:
@Fryguy running GC every `10MB` added 2% overhead, every `1MB` a `16%` overhead so `10MB` seems like a good compromise.

kbrock:
I'm +1 w/ no `Gc.start` and no configuration.
Keep in mind that every point release of ruby could change these numbers. (which would encourage a configuration option :( )

Wonder if it is possible to stream this output instead of allocating so much memory in the first place.
Or put into an array and join at the end.  The way we are building up these strings is probably causing the extra bloat.

agrare:
@kbrock I'll look at using a stream parser instead of the standard parser.
@Fryguy would you prefer periodic forced GC or no GC?

Fryguy:
@agrare I like it how it is in this PR.  Stream (specifically a SAX) parser was attempted in https://github.com/ManageIQ/manageiq/pull/5135 , but since we added paging it was less useful.

agrare:
Euwe BZ https://bugzilla.redhat.com/show_bug.cgi?id=1399803
Darga BZ https://bugzilla.redhat.com/show_bug.cgi?id=1399805

chessbyte:
@kbrock see https://github.com/ManageIQ/manageiq/pull/5135

chessbyte:
Euwe Backport details:
```sh
$ git log -1
commit 4c4aaeb29a4933a331cd3294e4d83520fe9b3154
Date:   Tue Nov 29 13:00:33 2016 -0500

    Merge pull request #17 from agrare/vmware_broker_gc
    
    Remove forced GC.start from every API call
    (cherry picked from commit ManageIQ/manageiq-gems-pending@725926cf8bdce06cd2610383be61c8526ae88195)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1399803
```

simaishi:
Backported to Darga via https://github.com/ManageIQ/manageiq/pull/13625

PR #19, opened 2016-11-17
VixDiskLib Fixes Required by Creation of manageiq-gems-pending Gem

jerryk55:
2 types of fixes to the VixDiskLib code path are needed due to the
creation of the manageiq-gems-pending gem.

1) All requires of VixDiskLib require a relative path - require "VixDiskLib/VixDiskLib".
2) Since the VixDiskLib server does not use the Rails Root some acrobatics
   were used to get the log directory.  These must change (and are probably better here
   than previously used)

@bdunne @Fryguy @roliveri please review and merge (and then do whatever gem magic is required if any)

### DIFF
--- a/lib/gems/pending/VixDiskLib/VixDiskLib.rb
+++ b/lib/gems/pending/VixDiskLib/VixDiskLib.rb
@@ -15,8 +15,6 @@ class VixDiskLibError < RuntimeError
 end
 
 SERVER_PATH = File.expand_path(__dir__)
-LOG_DIR     = ENV["LOG"]
-LOG_FILE    = File.join(LOG_DIR, "vim.log")
 
 class VixDiskLib
   VIXDISKLIB_FLAG_OPEN_READ_ONLY = FFI::VixDiskLib::API::VIXDISKLIB_FLAG_OPEN_READ_ONLY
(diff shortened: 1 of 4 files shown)


lib/gems/pending/VixDiskLib/VixDiskLib.rb
```diff
@@ -15,8 +15,7 @@ class VixDiskLibError < RuntimeError
 end
 
 SERVER_PATH = File.expand_path(__dir__)
-MIQ_ROOT    = File.expand_path(File.join(SERVER_PATH, "../../.."))
-LOG_DIR     = File.join(MIQ_ROOT, "log")
+LOG_DIR     = ENV["LOG"]
```
bdunne: What is setting `ENV["LOG"]`?

jerryk55:
@bdunne oh. duh.  I think that's my own environment.  That won't work.  Thanks for catching. Lemme go back to the drawing board.

Fryguy:
> Since the VixDiskLib server does not use the Rails Root some acrobatics were used to get the log directory.

I would expect the code to use the known $log variables.  Those would be available in a Rails context and can be created in a script/test contest.

jerryk55:
@Fryguy as long as the $log vars are available in the VixDiskLib client code before it nukes the Rails environment the relevant one can be passed along to VixDiskLibServer.  I will look at that.

jerryk55:
@Fryguy @bdunne now passing value of $vim_log via the environment in the spawn.  Tested ok.

Fryguy:
Please rebase now that the repo has Travis support.

jerryk55:
Done.

jerryk55:
@miq-bot assign @bdunne

jerryk55:
@miq-bot add_label bug

PR #22, opened 2016-11-18
Remove warnings

Fryguy:
As of rake 11, warnings are on by default, as this is why our output is flooded.  To go back to the old way we can set `t.warning = false` in the Rake task, but I want to remove a few before disabling.  There are still more, but this trims out a lot.

@bdunne Please review.

### DIFF
--- a/lib/gems/pending/VolumeManager/MiqLdm.rb
+++ b/lib/gems/pending/VolumeManager/MiqLdm.rb
@@ -475,7 +475,7 @@ if __FILE__ == $0
     exit(1)
   end
 
-  parts = disk.getPartitions
+  # parts = disk.getPartitions
 
   puts "Disk type: #{disk.diskType}"
   puts "Disk partition type: #{disk.partType}"
--- a/lib/gems/pending/metadata/VmConfig/VmConfig.rb
+++ b/lib/gems/pending/metadata/VmConfig/VmConfig.rb
@@ -106,7 +106,7 @@ class VmConfig
       filename = @cfgHash[dk + ".filename"]
       @diskFileHash[dk] = filename
       if @direct_file_access
-        ds, dir, name = split_filename(filename)
+        ds, _dir, _name = split_filename(filename)
         if ds.nil? && !Pathname.new(filename).absolute?
           @diskFileHash[dk] = File.expand_path(File.join(@configPath, filename))
         end
@@ -141,7 +141,7 @@ class VmConfig
       disk_path = filename
 
       if @direct_file_access
-        ds, dir, name = split_filename(filename)
+        ds, _dir, _name = split_filename(filename)
         if ds.nil? && !Pathname.new(@cfgHash[dk + ".filename"]).absolute?
           disk_path = File.join(@configPath, filename)
         end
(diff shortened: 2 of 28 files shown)


lib/gems/pending/VolumeManager/MiqLdm.rb
```diff
@@ -475,7 +475,7 @@ def format(event)
     exit(1)
   end
 
-  parts = disk.getPartitions
+  # parts = disk.getPartitions
```
bdunne: Why comment rather than delete?

lib/gems/pending/metadata/VmConfig/VmConfig.rb
```diff
@@ -665,13 +664,13 @@ def getDevicePath(name, endType)
...
     # Initialize variables
-    full_pos, maj_pos, device, pos_idx = split_data(name)
+    full_pos, _maj_pos, _device, pos_idx = split_data(name)
 
     return [["vm"], ["snapshots"]] if pos_idx.nil?
     ret = [["vm"], ["snapshots"], [endType, {"id" => full_pos}]]
 
     if disk
-      a, b, subDevice, c = split_data(disk)
+      _, _, subDevice, _ = split_data(disk)
```
bdunne: `subDevice = split_data(disk)[2]`?

lib/gems/pending/metadata/util/md5deep.rb
```diff
@@ -244,7 +242,7 @@ def to_xml
 
   # Mount VM Image to a real drive letter
   #  mountNative, startPath = false, "M:/WINDOWS/system32/mui"
-  startPath = "c:/windows/system32"
+  #  startPath = "c:/windows/system32"
```
bdunne: Looks like this `startPath` comment could be merged with the one above.  Also makes me wonder if there is a reason for these comments.

lib/gems/pending/util/miq-logger.rb
```diff
@@ -136,8 +135,8 @@ def addPostRollEvent(name, run_now = false, &blk)
 
       def miqLogHeader(fileName = nil, logData = true)
         # Log an opening message framed in "*" chars
-        @miqFileName = fileName if fileName
-        if logData == true && @miqFileName.nil? == false
+        @miqFileName = fileName
```
bdunne: This changes behavior a bit...

lib/gems/pending/metadata/VmConfig/VmConfig.rb
Fryguy: Oh yeah I like that better.

lib/gems/pending/VolumeManager/MiqLdm.rb
Fryguy: It's part of a `if __FILE__ == $0` section, and I didn't want to mess with the intentions of that (and since it's in there it's relatively benign).  I believe its purpose is to "test" that particular method...I felt removing it hid that.

I'm not sure...what do you think given my reasoning above?

lib/gems/pending/metadata/util/md5deep.rb
Fryguy: Yeah I agree...I'll knock out this line and just leave the comment above.

lib/gems/pending/util/miq-logger.rb
Fryguy: Yeah you're right...If this gets called more than once it will override it the second time...

The Ruby was killing my eyes...I shouldn't have overstepped :smile:

Fryguy:
@bdunne Updated.

PR #24, opened 2016-11-28
Add Datastore maintenanceMode to VMwareWebService

agrare:
Datastore maintenance mode was added in vSphere 5.0 and allows the administrator to mark a datastore that is part of a datastore cluster as being in maintenance mode.  The intention is that this can be used to remove datastores that are in maintenance mode from the list of eligible targets for provisioning.

Datastore not in maintenance mode:
screenshot from 2016-11-28 12-29-31

Datastore in maintenance mode:
screenshot from 2016-11-28 12-29-23

VC SDK link: http://pubs.vmware.com/vsphere-50/index.jsp?topic=%2Fcom.vmware.wssdk.apiref.doc_50%2Fvim.Datastore.Summary.html
```
maintenanceMode*	xsd:string	
The current maintenance mode state of the datastore. The set of possible values is described in DatastoreSummaryMaintenanceModeState.
Since vSphere API 5.0
```

https://bugzilla.redhat.com/show_bug.cgi?id=1394909

### DIFF
--- a/lib/gems/pending/VMwareWebService/MiqVimInventory.rb
+++ b/lib/gems/pending/VMwareWebService/MiqVimInventory.rb
@@ -47,6 +47,7 @@ class MiqVimInventory < MiqVimClientBase
       deleteProperty(:HostSystem, "capability.vmotionWithStorageVMotionSupported")
 
       deleteProperty(:Datastore, "summary.uncommitted")
+      deleteProperty(:Datastore, "summary.maintenanceMode")
 
       deleteProperty(:VirtualMachine, "config.cpuHotAddEnabled")
       deleteProperty(:VirtualMachine, "config.cpuHotRemoveEnabled")
@@ -64,11 +65,17 @@ class MiqVimInventory < MiqVimClientBase
         deleteProperty(:HostSystem, "config.dateTimeInfo")
       end
     else
-      if @v4 && cacheScope != :cache_scope_event_monitor
+      if @v4
         @propMap = dupProps(@propMap)
-        addProperty(:VirtualMachine, "runtime.memoryOverhead")
-        deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")
+
+        deleteProperty(:Datastore, "summary.maintenanceMode")
+
+        if cacheScope != :cache_scope_event_monitor
+          addProperty(:VirtualMachine, "runtime.memoryOverhead")
+          deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")
+        end
       end
+
       @propMap = @propMap.merge(PropMap4)
     end
 
--- a/lib/gems/pending/VMwareWebService/VimPropMaps.rb
+++ b/lib/gems/pending/VMwareWebService/VimPropMaps.rb
@@ -344,6 +344,7 @@ module VimPropMaps
         "summary.capacity",
         "summary.datastore",
         "summary.freeSpace",
+        "summary.maintenanceMode",
         "summary.multipleHostAccess",
         "summary.name",
         "summary.type",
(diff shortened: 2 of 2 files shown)


agrare:
cc @roliveri

agrare:
@roliveri can you review?

lib/gems/pending/VMwareWebService/MiqVimInventory.rb
```diff
@@ -64,11 +65,17 @@ def initialize(server, username, password, cacheScope = nil)
         deleteProperty(:HostSystem, "config.dateTimeInfo")
       end
     else
-      if @v4 && cacheScope != :cache_scope_event_monitor
+      if @v4
```
roliveri: Is the removal of the other conditions related to this PR? Is there no longer a need to check the cache scope?

agrare: @roliveri I just moved that conditional down [here](https://github.com/ManageIQ/manageiq-gems-pending/pull/24/files/7aab7da6d2624d0b984281c0b3b529d1233efc2d#diff-618aa8640f55186534c8e2be1b5b38b4R73l) so those two add/delete properties are the same but we're removing the datastore maintenance mode for all 4.x VCs

lib/gems/pending/VMwareWebService/MiqVimInventory.rb
```diff
@@ -64,11 +65,17 @@ def initialize(server, username, password, cacheScope = nil)
...
         @propMap = dupProps(@propMap)
-        addProperty(:VirtualMachine, "runtime.memoryOverhead")
-        deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")
+
+        deleteProperty(:Datastore, "summary.maintenanceMode")
+
+        if cacheScope != :cache_scope_event_monitor
+          addProperty(:VirtualMachine, "runtime.memoryOverhead")
+          deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")
```
agrare: @roliveri actually I don't fully understand this block, why is the `deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")` in `if cacheScope != :cache_scope_event_monitor` for `@v4` but not [here](https://github.com/agrare/manageiq-gems-pending/blob/7aab7da6d2624d0b984281c0b3b529d1233efc2d/lib/gems/pending/VMwareWebService/MiqVimInventory.rb#L54) in `@v2`?  I get not wanting to do the `addProperty` since that wouldn't be in the `EventMonitorPropMap` anyway but is that the only one that should be in the `if cacheScope != :cache_scope_event_monitor` conditional?

roliveri: I think those properties didn't exist in v2 so they needed to be added for v4, but the event monitor didn't need them.

agrare: Yeah I get the `addProperty` needing to be in there, but why is this [`deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")` in `@v2`](https://github.com/ManageIQ/manageiq-gems-pending/blob/ba8878680bb41af9ca22e17673316e0aa0ac3de9/lib/gems/pending/VMwareWebService/MiqVimInventory.rb#L55) different than this [`deleteProperty(:VirtualMachine, "config.hardware.numCoresPerSocket")` in `@v4`](https://github.com/ManageIQ/manageiq-gems-pending/blob/ba8878680bb41af9ca22e17673316e0aa0ac3de9/lib/gems/pending/VMwareWebService/MiqVimInventory.rb#L72)  one is in `if cacheScope != :cache_scope_event_monitor` but the other is not.  Maybe it was kind of an optimization since at that point nothing had to be done for `@v4` if it was an event monitor?  Just is kind of confusing and the deleteProperty should work if there is nothing in the prop map anyway...

roliveri: @agrare 
Hmmm, it's been so long, I can't remember. It could be those `deleteProperty` calls were added at different times, not realizing that a call already existed in another location - but I'm not sure.

PR #29, opened 2016-12-06
Add a port prompt to database setup

carbonin:
This PR makes changes that allow a user of the appliance_console or appliance_console_cli to specify a non-default (other than 5432) port to use when connecting to the database.

This only really applies to external databases as we use a socket when connecting locally.

We default the port to 5432 when configuring an external database and do not prompt for a port when configuring an internal database.

@jvlcek @gtanzillo please review

https://bugzilla.redhat.com/show_bug.cgi?id=1386331

### DIFF
--- a/lib/gems/pending/appliance_console/database_configuration.rb
+++ b/lib/gems/pending/appliance_console/database_configuration.rb
@@ -28,6 +28,7 @@ module ApplianceConsole
     # 9223372 won't be a full region though, so we're not including it.
     # TODO: This information should be shared outside of appliance console code and MiqRegion.
     REGION_RANGE = 0..9223371
+    DEFAULT_PORT = 5432
 
     include ApplianceConsole::Logging
 
@@ -114,7 +115,7 @@ module ApplianceConsole
 
     def ask_for_database_credentials
       self.host     = ask_for_ip_or_hostname("database hostname or IP address", host) if host.blank? || !local?
-      self.port     = ask_for_integer("port number", nil, default_port) unless local?
+      self.port     = ask_for_integer("port number", nil, port) unless local?
       self.database = just_ask("name of the database on #{host}", database) unless local?
       self.username = just_ask("username", username) unless local?
       count = 0
@@ -273,10 +274,6 @@ FRIENDLY
           raise ArgumentError, "Invalid argument: #{k}"
         end
       end
-    end
-
-    def default_port
-      5432
     end
   end
 end
--- a/lib/gems/pending/appliance_console/database_configuration.rb
+++ b/lib/gems/pending/appliance_console/database_configuration.rb
@@ -141,6 +141,7 @@ Host:     #{host}
 Username: #{username}
 Database: #{database}
 FRIENDLY
+      output << "Port:     #{port}\n" if port
       output << "Region:   #{region}\n" if region
       output
     end
--- a/lib/gems/pending/appliance_console/database_configuration.rb
+++ b/lib/gems/pending/appliance_console/database_configuration.rb
@@ -114,6 +114,7 @@ module ApplianceConsole
 
     def ask_for_database_credentials
       self.host     = ask_for_ip_or_hostname("database hostname or IP address", host) if host.blank? || !local?
+      self.port     = ask_for_integer("port number", nil, default_port) unless local?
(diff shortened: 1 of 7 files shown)


lib/gems/pending/appliance_console/database_configuration.rb
```diff
@@ -140,6 +141,7 @@ def friendly_inspect
 Username: #{username}
 Database: #{database}
 FRIENDLY
+      output << "Port:     #{port}\n" if port
```
kbrock: do we want a port if they chose the default?
```ruby
if port && port != default_port
```

lib/gems/pending/appliance_console/prompts.rb
```diff
@@ -126,8 +126,8 @@ def ask_for_string(prompt, default = nil)
       just_ask(prompt, default)
     end
 
-    def ask_for_integer(prompt, range = nil)
-      just_ask(prompt, nil, INT_REGEXP, "an integer", Integer) { |q| q.in = range if range }
+    def ask_for_integer(prompt, range = nil, default = nil)
+      just_ask(prompt, default, INT_REGEXP, "an integer", Integer) { |q| q.in = range if range }
```
kbrock: very very nice

lib/gems/pending/appliance_console/database_configuration.rb
carbonin: I would think we should still display it right?

lib/gems/pending/appliance_console/internal_database_configuration.rb
```diff
@@ -25,6 +25,7 @@ def initialize(hash = {})
 
     def set_defaults
       self.host              = "127.0.0.1"
+      self.port              = default_port
```
carbonin: I think this should not be set here and defaulted in external database configuration ... oops.

lib/gems/pending/appliance_console/database_configuration.rb
kbrock: aah. I was thinking we would only display it if we wanted to change something.

same with writing to the yaml file. only write it if it is non default.

leaving it in sounds good

jvlcek: I think we should leave it. Display the default if being used as the default value.

lib/gems/pending/appliance_console/database_configuration.rb
```diff
@@ -272,5 +274,9 @@ def initialize_from_hash(hash)
         end
       end
     end
+
+    def default_port
```
jvlcek: Why not have an `attr_reader :default_port` and set it in `DatabaseConfiguration#initialize`

lib/gems/pending/appliance_console/internal_database_configuration.rb
jvlcek: 👍

lib/gems/pending/appliance_console/prompts.rb
jvlcek: 👍

lib/gems/pending/appliance_console/database_configuration.rb
carbonin: I think `attr_reader` implies something that can or might be changed. If I were to change this, I might actually make it a constant. Thoughts?

carbonin: You can always override the port which is an `attr_accessor`, but I don't see any need to override the default. Kinda would defeat the purpose of a default ...

jvlcek: @carbonin OK then a constant would be a good choice.

lib/gems/pending/appliance_console/database_configuration.rb
```diff
@@ -28,6 +28,7 @@ class ModelWithNoBackingTable < ActiveRecord::Base
     # 9223372 won't be a full region though, so we're not including it.
     # TODO: This information should be shared outside of appliance console code and MiqRegion.
     REGION_RANGE = 0..9223371
+    DEFAULT_PORT = 5432
```
jvlcek: 👍

jvlcek:
@carbonin Looks great! 👍

PR #31, opened 2016-12-19
Add Azure SSA rake tasks

djberg96:
This adds the SSA related tasks for Azure back into the available rake tasks.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -3,6 +3,8 @@ require 'rspec/core/rake_task'
 require 'rake'
 require 'rake/testtask'
 
+Rake.add_rakelib 'lib/gems/pending/lib/tasks'
+
 RSpec::Core::RakeTask.new(:spec) do |t|
   t.rspec_opts = "--options #{File.expand_path(".rspec_ci", __dir__)}" if ENV['CI']
 end


Fryguy:
Not sure why this is in gems pending...I would expect this in the azure provider plugin.

djberg96:
@Fryguy These tasks are for the SSA related stuff, regenerating cassettes and so on.

Fryguy:
There is a way to include rake tasks within the azure plugin.  If tasks are placed within lib/tasks of the plugin gem, then they will be available from the plugin and from manageiq.  If they are put in lib/tasks_private, they will only be available within the gem.  IMO, these tasks should be moved from here into the gem.

@durandom @bdunne Thoughts?

Rakefile
```diff
@@ -3,6 +3,8 @@ require 'rspec/core/rake_task'
 require 'rake'
 require 'rake/testtask'
 
+Rake.add_rakelib 'lib/gems/pending/lib/tasks'
```
bdunne: Why is this file in `lib/gems/pending/lib/tasks` rather than `lib/tasks`?

bdunne:
@Fryguy I guess I don't know enough about it to decide where it should live, but I don't like where it currently is. :)

durandom:
Shouldnt those specs and cassettes move to the azure provider?

djberg96:
@durandom Eventually the SSA stuff is going to get split into its own gem. I think that's the plan anyway.

chessbyte:
@djberg96 Eventually is almost here: https://github.com/ManageIQ/manageiq-gems-pending/pull/164

djberg96:
@chessbyte Alright, I'll move them over when it's ready.

Fryguy:
@miq-bot move_issue ManageIQ/manageiq-smartstate

Fryguy:
Derp, of course not :)

Fryguy:
@djberg96 Please recreate this PR in the new repo https://github.com/ManageIQ/manageiq-smartstate

PR #32, opened 2016-12-21
Add concurrent jobs parameter to pg_dump and pg_restore for faster exec

dtrieu80:
Out of the box cloudforms appliances are already a 2x6, but most database are larger (4x8 or greater). We should be able to take advantage of this with our database backups and restores using concurrency.


carbonin:
This option can't be used with the custom output format.

From the [postgres docs](https://www.postgresql.org/docs/9.5/static/app-pgdump.html):
>You can only use this option with the directory output format because this is the only output format where multiple processes can write their data at the same time.

PR #34, opened 2017-01-05
Do not stop and start evmserverd if it was not running.

jvlcek:
This PR simply ensures the evmserverd service is not manipulated after logfile
configuration changes are made, if the evmserverd service was not running.

**To test this upstream:**

- add an extra harddisk to the test appliance,
- disable evmserverd
- attempt to use the appliance_console to add a new logfile disk volume.

To disable the evmserverd service upstream do:

```
  systemctl stop evmserverd && systemctl disable evmserverd
  systemctl stop $APPLIANCE_PG_SERVICE && systemctl disable $APPLIANCE_PG_SERVICE
  echo "rm PG data"
  rm -rf $APPLIANCE_PG_DATA/*
  vmdb
  rm -f REGION config/database.yml GUID
```

After executing the above commands use the appliance_console to configure a new
logfile disk volume, ensuring no traceback is produced. 

https://bugzilla.redhat.com/show_bug.cgi?id=1393349

### DIFF
--- a/lib/gems/pending/appliance_console/logfile_configuration.rb
+++ b/lib/gems/pending/appliance_console/logfile_configuration.rb
@@ -11,7 +11,7 @@ module ApplianceConsole
     LOGFILE_NAME = "miq_logs".freeze
     MIQ_LOGS_CONF = Pathname.new("/etc/logrotate.d/miq_logs.conf").freeze
 
-    attr_accessor :size, :disk, :current_logrotate_count, :new_logrotate_count
+    attr_accessor :size, :disk, :current_logrotate_count, :new_logrotate_count, :evm_was_running
 
     include ApplianceConsole::Logging
 
@@ -21,6 +21,7 @@ module ApplianceConsole
 
       self.size = MiqSystem.disk_usage(LOGFILE_DIRECTORY)[0][:total_bytes]
       self.current_logrotate_count = /rotate\s+(\d+)/.match(File.read(MIQ_LOGS_CONF))[1]
+      self.evm_was_running = LinuxAdmin::Service.new("evmserverd").running?
     end
 
     def activate
@@ -102,12 +103,14 @@ module ApplianceConsole
     end
 
     def start_evm
+      return true unless evm_was_running
       say 'Starting EVM'
       LinuxAdmin::Service.new("evmserverd").enable.start
       LinuxAdmin::Service.new("httpd").enable.start
     end
 
     def stop_evm
+      return true unless evm_was_running
       say 'Stopping EVM'
       LinuxAdmin::Service.new("evmserverd").stop
       LinuxAdmin::Service.new("httpd").stop
--- a/lib/gems/pending/appliance_console/logfile_configuration.rb
+++ b/lib/gems/pending/appliance_console/logfile_configuration.rb
@@ -86,9 +86,9 @@ module ApplianceConsole
 
     def activate_new_disk
       return true unless disk
-      stop_evm
+      stop_evm if evm_was_running
       initialize_logfile_disk
-      start_evm
+      start_evm if evm_was_running
       true
     end
 
@@ -103,14 +103,12 @@ module ApplianceConsole
     end
 
     def start_evm
-      return true unless evm_was_running
       say 'Starting EVM'
       LinuxAdmin::Service.new("evmserverd").enable.start
       LinuxAdmin::Service.new("httpd").enable.start
     end
 
     def stop_evm
-      return true unless evm_was_running
       say 'Stopping EVM'
       LinuxAdmin::Service.new("evmserverd").stop
       LinuxAdmin::Service.new("httpd").stop
--- a/spec/appliance_console/logfile_configuration_spec.rb
+++ b/spec/appliance_console/logfile_configuration_spec.rb
@@ -25,6 +25,7 @@ describe ApplianceConsole::LogfileConfiguration do
   let(:miq_logs_conf) { Tempfile.new(@spec_name.downcase) }
 
   before do
+    allow(LinuxAdmin::Service).to receive(:new).and_return(double(@spec_name, :running? => true))
     @spec_name = File.basename(__FILE__).split(".rb").first.freeze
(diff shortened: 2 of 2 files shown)


jvlcek:
@carbonin Please review

jvlcek:
@miq-bot add_label appliance/console

jvlcek:
@miq-bot add_label bug

lib/gems/pending/appliance_console/logfile_configuration.rb
```diff
@@ -102,12 +103,14 @@ def initialize_logfile_disk
     end
 
     def start_evm
+      return true unless evm_was_running
```
carbonin: I think this would read better if we moved this check to the caller.
So in `#activate_new_disk` you would get:
```ruby
start_evm if evm_was_running
```

lib/gems/pending/appliance_console/logfile_configuration.rb
```diff
@@ -102,12 +103,14 @@ def initialize_logfile_disk
...
       say 'Starting EVM'
       LinuxAdmin::Service.new("evmserverd").enable.start
       LinuxAdmin::Service.new("httpd").enable.start
     end
 
     def stop_evm
+      return true unless evm_was_running
```
carbonin: Same as above

spec/appliance_console/logfile_configuration_spec.rb
```diff
@@ -89,22 +90,47 @@
       EOT
     end
 
-    it "stops and starts evm and configures the logfile disk" do
-      subject.new_logrotate_count = 3
-      subject.disk = double(@spec_name, :size => "9999999", :path => "fake disk")
+    context "evm was running" do
```
carbonin: If there's only one spec in the `context` I think you can just add the context description at the end of the spec name. In this case it would be:
`it "stops and starts evm and configures the logfile disk when evm was running" do`

spec/appliance_console/logfile_configuration_spec.rb
```diff
@@ -89,22 +90,47 @@
...
+          .to receive(:new).and_return(double(@spec_name, :enable => double(@spec_name, :start => true))).twice
 
-      expect(subject.activate).to be true
-      expect(File.read(miq_logs_conf)).to eq(expected_miq_logs_conf)
+        expect(subject.activate).to be true
+        expect(File.read(miq_logs_conf)).to eq(expected_miq_logs_conf)
+      end
+    end
+
+    context "evm was not running" do
```
carbonin: Same here, no need to add a context when there is only one spec.

spec/appliance_console/logfile_configuration_spec.rb
```diff
@@ -94,17 +94,19 @@
...
-  
-        expect(ApplianceConsole::LogicalVolumeManagement).to receive(:new).and_return(double(@spec_name, :setup => true))
+
+        expect(ApplianceConsole::LogicalVolumeManagement)
+          .to receive(:new).and_return(double(@spec_name, :setup => true))
         expect(File).to receive(:executable?).with("/sbin/restorecon").and_return(true)
         expect(AwesomeSpawn).to receive(:run!).with('/sbin/restorecon -R -v /var/www/miq/vmdb/log')
-        expect(FileUtils).to receive(:mkdir_p).with("#{ApplianceConsole::LogfileConfiguration::LOGFILE_DIRECTORY}/apache")
+        expect(FileUtils)
+          .to receive(:mkdir_p).with("#{ApplianceConsole::LogfileConfiguration::LOGFILE_DIRECTORY}/apache")
```
carbonin: I would prefer this split to happen later so the first line isn't so short.
Maybe:
```ruby
expect(FileUtils).to receive(:mkdir_p)
  .with("#{ApplianceConsole::LogfileConfiguration::LOGFILE_DIRECTORY}/apache")
```

spec/appliance_console/logfile_configuration_spec.rb
```diff
@@ -115,15 +117,17 @@
...
```
carbonin: Same as above

lib/gems/pending/appliance_console/logfile_configuration.rb
jvlcek: @carbonin Thank you for the review. I had thought of doing it the way you suggested but then thought that would require the caller to have knowledge of what the methods `start_evm` and `stop_evm` are doing and how. So I think it would be better to leave it as is. Let's discuss if you feel strongly about it.

spec/appliance_console/logfile_configuration_spec.rb
jvlcek: @carbonin I'm OK with doing this but the way I have it all of the lines that are too long, for consistency are split at the `.to`. So again let's discuss if you feel strongly about changing this.

spec/appliance_console/logfile_configuration_spec.rb
jvlcek: Again I was originally going to code this the way you are requesting. Before posting the PR I consulted http://betterspecs.org/#contexts My interpretation of which suggests the way I have it.

jvlcek:
@carbonin I am totally open to changing the code as you suggest if you feel strongly about it. However, as I noted in my comments, I `think` the code is OK as is. If you feel strongly please ping me to discuss. Thank you. JoeV

jvlcek:
@carbonin On second thought, after reading `https://www.destroyallsoftware.com/talks/boundaries` and `https://www.destroyallsoftware.com/screencasts/catalog/functional-core-imperative-shell` and considering there is only a single spec test impacted, I have decided to implement the code the way you suggest.

I will go ahead and update the PR. No need to discuss.

Thank you. JoeV

jvlcek:
@carbonin I've address the comments you've made. Thank you again for the help. Sorry for any complications with my initial perspective on leaving the code as it was... JoeV

simaishi:
@jvlcek Would you mind creating an Euwe PR for this? Cherry-pick isn't working for this PR...

simaishi:
Backpoted to Euwe via https://github.com/ManageIQ/manageiq/pull/13479

PR #38, opened 2017-01-16
Apply Selection Spec to `hostSystemsStorageDevice`

agrare:
Use the provided selection spec to filter the properties returned by the VC.  The selection spec was being applied to the result but not to the propPath before the call.

This was already done in MiqVimHost in https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/VMwareWebService/MiqVimHost.rb#L267 but didn't get moved over to `MiqVimInventory`

This should reduce the size of the payload returned by excluding `config.storageDevice.multipathInfo`, `config.storageDevice.plugStoreTopology`, and `config.storageDevice.softwareInternetScsiEnabled`

### DIFF
--- a/lib/gems/pending/VMwareWebService/MiqVimInventory.rb
+++ b/lib/gems/pending/VMwareWebService/MiqVimInventory.rb
@@ -1352,8 +1352,13 @@ class MiqVimInventory < MiqVimClientBase
   end
 
   def hostSystemsStorageDevice(hostMors, selSpec = nil)
-    sd = getMoPropMulti(hostMors, "config.storageDevice")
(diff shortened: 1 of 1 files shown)


roliveri:
@agrare should this be back ported to any branches?

agrare:
@roliveri just `euwe`

simaishi:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1432544

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/14253

PR #39, opened 2017-01-17
Add RewriteCond to disable rewrite to index of css and js files 

bpereto:
Add RewriteCond to disable rewrite to index of css and js files in thee self service ui and make the ui work again.

At the moment the self service ui in the nightly appliance does not work and shows a blank site. there is an issue with the rewriterules in the apache config, which rewrites the css and js assets to index.html.

Also do not rewrite the `data` dir, which is used by the css.


carbonin:
Is this still needed. I can see the SUI fine in the most recent nightly builds.

@chriskacerguis ping

chriskacerguis:
Nope. Don't need as we use webpack now.

carbonin:
Okay, thanks. Closing this.

PR #40, opened 2017-01-17
Miq password decrypt with an alternate key

bdunne:
Allow MiqPassword#decrypt to also use an alternate key.
https://bugzilla.redhat.com/show_bug.cgi?id=1400995

### DIFF
--- a/lib/gems/pending/util/miq-password.rb
+++ b/lib/gems/pending/util/miq-password.rb
@@ -24,20 +24,18 @@ class MiqPassword
   end
 
   def decrypt(str, legacy = false)
-    if str.nil? || str.empty?
-      str
-    else
-      ver, enc = self.class.split(str)
-      return "" if enc.empty?
+    return str if str.nil? || str.empty?
 
-      ver ||= "0" # if we don't know what it is, just assume legacy
-      key_name = (ver == "2" && legacy) ? "alt" : "v#{ver}"
+    ver, enc = self.class.split(str)
+    return "" if enc.empty?
 
-      begin
-        self.class.keys[key_name].decrypt64(enc).force_encoding('UTF-8')
-      rescue
-        raise MiqPasswordError, "can not decrypt v#{ver}_key encrypted string"
-      end
+    ver ||= "0" # if we don't know what it is, just assume legacy
+    key_name = (ver == "2" && legacy) ? "alt" : "v#{ver}"
+
+    begin
+      self.class.keys[key_name].decrypt64(enc).force_encoding('UTF-8')
+    rescue
+      raise MiqPasswordError, "can not decrypt v#{ver}_key encrypted string"
     end
   end
 
--- a/lib/gems/pending/util/miq-password.rb
+++ b/lib/gems/pending/util/miq-password.rb
@@ -23,17 +23,21 @@ class MiqPassword
     "#{ver}:{#{value}}"
   end
 
-  def decrypt(str, legacy = false)
+  def decrypt(str, legacy = false, key = nil)
     return str if str.nil? || str.empty?
 
     ver, enc = self.class.split(str)
     return "" if enc.empty?
 
     ver ||= "0" # if we don't know what it is, just assume legacy
-    key_name = (ver == "2" && legacy) ? "alt" : "v#{ver}"
+
+    key ||= begin
+      key_name = (ver == "2" && legacy) ? "alt" : "v#{ver}"
+      self.class.keys[key_name]
+    end
 
     begin
-      self.class.keys[key_name].decrypt64(enc).force_encoding('UTF-8')
+      key.decrypt64(enc).force_encoding('UTF-8')
     rescue
       raise MiqPasswordError, "can not decrypt v#{ver}_key encrypted string"
     end
--- a/spec/util/miq-password_spec.rb
+++ b/spec/util/miq-password_spec.rb
@@ -321,6 +321,17 @@ describe MiqPassword do
     end
   end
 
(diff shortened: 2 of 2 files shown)


bdunne:
@carbonin @kbrock @Fryguy Please review

lib/gems/pending/util/miq-password.rb
```diff
@@ -23,21 +23,23 @@ def encrypt(str, ver = "v2", key = self.class.keys[ver])
     "#{ver}:{#{value}}"
   end
 
-  def decrypt(str, legacy = false)
-    if str.nil? || str.empty?
-      str
-    else
-      ver, enc = self.class.split(str)
-      return "" if enc.empty?
+  def decrypt(str, legacy = false, key = nil)
```
carbonin: I'm not sure if it's worth doing now/in this PR, but I have no idea what `legacy` means in this context. Can we rename it to something better?

lib/gems/pending/util/miq-password.rb
```diff
@@ -23,21 +23,23 @@ def encrypt(str, ver = "v2", key = self.class.keys[ver])
...
+    return str if str.nil? || str.empty?
+
+    ver, enc = self.class.split(str)
+    return "" if enc.empty?
+
+    ver ||= "0" # if we don't know what it is, just assume legacy
 
-      ver ||= "0" # if we don't know what it is, just assume legacy
+    key ||= begin
       key_name = (ver == "2" && legacy) ? "alt" : "v#{ver}"
```
carbonin: These few lines make me even more confused about `legacy` (especially the comment).

`legacy` seems to only be used when `ver` is 2. So why are we referring to `ver == 0` as legacy?

spec/util/miq-password_spec.rb
```diff
@@ -365,4 +376,22 @@ def with_key
   def erberize(password, passmethod = "MiqPassword")
     "<%= #{passmethod}.decrypt(\"#{password}\") %>"
   end
+
+  def with_random_key
```
carbonin: I was confused with the term "random" here. Maybe something like `with_external_key` or `with_alternate_key`?

It's not really random because if it were you wouldn't be able to assert on the encrypted value.

bdunne: yeah, naming is hard

lib/gems/pending/util/miq-password.rb
bdunne: I'm not really sure, so I'd rather leave it alone for now.

kbrock: posted thoughts below

lib/gems/pending/util/miq-password.rb
kbrock: ugh. who wrote that?

those are different "legacy"

line 33 means legacy aka pre 5.0 keys.

line 34 means legacy aka the v2 keys that is no longer valid (because we are actively migrating away from it) - it is stored as `:alt`

kbrock:
disclaimer:
It took 2 years to get a process that would allow users to migrate from one v2 key to another.
I later discovered (due to sporadic build failures) that you could use the wrong v2 key to decrypt some text and it would false positive that it worked.

This was introduced.
Fast forward to today, and I think some things were over engineered and others (being able to pass in a key) is missing.

Let me describe how the current code works, so we can change it to something better. (note: I'm not defending the current code. I agree it needs to change)

----

When we are migrating from one v2 key to another v2 key, the old v2 key (the one that is no longer valid) is called legacy and it is stored in the hash as `:alt`. ( `#include "snark"` )

When we are in regular mode, if it finds text that is encrypted with a v2 key (it has the form `v2:{TEXT}` it will use the key stored as `:v2` to decrypt. Otherwise, it will use the key encrypted with the `:v0` or `:v1` key. If it does not start with `V*`, then possibly it is older stored string with `<%= %>` or other. this is the second `legacy` reference.

When we want to migrate from one key to another, then decrypt is called in in legacy mode (should probably be called "migration" mode or "change key" mode). In this case, `legacy=true`. If it finds text of the form `v2:{TEXT}`, it will use the key stored as `:alt` to decrypt it. (v0 and v1 work as well)

When it re-encrypts the text, it will use the key stored under `:v2`.

If you are upgrading from a 5.0 database, you will need to upgrade to 5.4 (or 5.3?) to get everything up to date first before proceeding to a more recent version. We truncated migrations before then among other things.

So we can probably drop legacy, v0, and v1 keys.

Either way, we are not going to upgrade text encrypted with legacy, v0, or v1 at the same time as changing from one v2 to another v2 key.

So, maybe this signature can just be changed to `decrypt(str, key = nil)`. And if `key == true`, then set key to `:alt`.

We can discuss more tomorrow

kbrock:
followup, `recrypt` is an interesting read: https://github.com/bdunne/manageiq-gems-pending/blob/f6a8cca72fc22600c2d6b6202873c7d8bb9b6601/lib/gems/pending/util/miq-password.rb#L46

lib/gems/pending/util/miq-password.rb
bdunne: I think it might be used when you're re-encrypting things, but not sure

Fryguy:
I don't get what you are trying to solve...can we chat?

carbonin:
@Fryguy We need to be able to decrypt strings using a specified key in a thread safe way.

`#encrypt` takes a key as a parameter, `#decrypt` did not. This is adding that ability to `#decrypt`

bdunne:
Closing due to change in strategy.

PR #41, opened 2017-01-18
Do not configure ext auth if http is not configured.

jvlcek:
This PR simply ensures attempts to configure external authentication are
abandoned if the manageiq http configuration files have not yet been
created.

### To test this upstream:

disable evmserverd
attempt to use the appliance_console to configure external authentication.
ensure not traceback is produced

### To disable the evmserverd service upstream do:

```
  systemctl stop evmserverd && systemctl disable evmserverd
  systemctl stop $APPLIANCE_PG_SERVICE && systemctl disable $APPLIANCE_PG_SERVICE
  echo "rm PG data"
  rm -rf $APPLIANCE_PG_DATA/*
  vmdb
  rm -f REGION config/database.yml GUID
  rm /etc/httpd/conf.d/manageiq*

```

After executing the above commands attempt to use the appliance_console to
configure external authentication, ensuring no traceback is produced.

https://bugzilla.redhat.com/show_bug.cgi?id=1396184

### DIFF
--- a/lib/gems/pending/appliance_console/external_httpd_authentication.rb
+++ b/lib/gems/pending/appliance_console/external_httpd_authentication.rb
@@ -79,11 +79,12 @@ module ApplianceConsole
     end
 
     def post_activation
-      say("\nRestarting sssd and httpd ...")
-      %w(sssd httpd).each { |service| LinuxAdmin::Service.new(service).restart }
+      say("\nRestarting httpd, if running ...")
+      httpd_service = LinuxAdmin::Service.new("httpd")
+      httpd_service.restart if httpd_service.running?
 
-      say("Configuring sssd to start upon reboots ...")
-      LinuxAdmin::Service.new("sssd").enable
+      say("Restarting sssd and configure it to start on reboots ...")
+      LinuxAdmin::Service.new("sssd").restart.enable
     end
 
     private
--- a/spec/appliance_console/external_httpd_authentication_spec.rb
+++ b/spec/appliance_console/external_httpd_authentication_spec.rb
@@ -1,6 +1,7 @@
 require "active_support/all"
 require "appliance_console/external_httpd_authentication"
 require "appliance_console/prompts"
+require "linux_admin"
 
 describe ApplianceConsole::ExternalHttpdAuthentication do
   let(:host) { "this.server.com" }
@@ -143,6 +144,22 @@ describe ApplianceConsole::ExternalHttpdAuthentication do
       subject.enable_kerberos_dns_lookups
       expect(File.read(@test_kerberos_config)).to eq(expected_kerberos_config)
       expect(File.read("#{@test_kerberos_config.path}.miqbkp")).to eq(expected_kerberos_config)
+    end
+  end
+
+  context "#post_activation" do
+    before do
+      @spec_name = File.basename(__FILE__).split(".rb").first.freeze
+    end
+
+    it "when http is not running it is not restarted" do
+      allow(subject).to receive(:say)
+      httpd_service = double(@spec_name, :running? => false)
+      expect(httpd_service).not_to receive(:restart)
+      allow(LinuxAdmin::Service).to receive(:new).with("httpd").and_return(httpd_service)
+      sssd_service = double(@spec_name, :restart => double(@spec_name, :enable => true))
+      allow(LinuxAdmin::Service).to receive(:new).with("sssd").and_return(sssd_service)
+      subject.post_activation
     end
   end
 end


jvlcek:
@miq-bot add_label appliance/console, bug

jvlcek:
@carbonin Please review

lib/gems/pending/appliance_console/external_httpd_configuration.rb
```diff
@@ -154,6 +165,7 @@ def installation_valid?
 
       def valid_environment?
         return false unless installation_valid?
+        return false unless http_configuration_valid?
```
carbonin: I think this can be combined into one condition.

```ruby
return false unless installation_valid? && http_configuration_valid?
```

lib/gems/pending/appliance_console/external_httpd_configuration.rb
```diff
@@ -15,6 +15,7 @@ module ExternalHttpdConfiguration
       SSSD_CONFIG          = "/etc/sssd/sssd.conf".freeze
       PAM_CONFIG           = "/etc/pam.d/httpd-auth".freeze
       HTTP_KEYTAB          = "/etc/http.keytab".freeze
+      HTTP_CONFIG_FILE     = "/etc/httpd/conf.d/manageiq-redirects-ws".freeze
```
carbonin: Is the absence of this file the reason this is happening or are we looking to this file as a marker of some other condition?

I guess I'm curious if we could check something less fragile like if `httpd` is running in case we change the name of the file or something. It just seems like this file might not be the best marker as this code didn't even know about it before this change.

jvlcek: @carbonin If this is not yet created restarting httpd by the appliance_console/`Configure External Auth` fails. This file is created when manageIQ is configured. It is possible that the file was created but someone could have stopped the httpd service.

I had discussed this very scenario with @abellotti and we felt the safest approach would be to check for the file.

jvlcek: @abellotti Can you please comment?

lib/gems/pending/appliance_console/external_httpd_configuration.rb
jvlcek: @carbonin 👍  Willdo!

lib/gems/pending/appliance_console/external_httpd_configuration.rb
abellotti: right, I don't think we document that evmserverd must be running for configuring external auth, thus the above check which essentially says appliance was previously started and all these httpd files were installed earlier.  Otherwise a simple check that both evmserverd & httpd are running would suffice.

@jvlcek one other possibility is restart httpd only if it is currently running (i.e. evmserverd is up and running and we want the changes made in effect immediately).

carbonin: I'm asking more along the lines of is this checking for the condition that is actually causing the failure or some other incidental side-effect of the state where we would succeed?

 If the presence or absence of this file was the direct cause of the console failure that this is fixing then I think this makes sense.

jvlcek: @abellotti 

I had originally coded the solution to check if evmserverd is running. However doing that turns out to be problematic because it can expose a timing issue, where evmserverd has been started but the http config file has not yet been created. Which results in the same traceback this PR attempting to avoid.

jvlcek: @abellotti and @carbonin I could check if the http service is running but that could be an overly restrictive use case. It would be possible for a user to configure external auth if the appliance had already been configured but httpd could have been stopped.

But perhaps requiring httpd to be running would not be unreasonable and it would allow us to avoid the concern @carbonin raised.

@abellotti and @carbonin Thoughts?

jvlcek: @carbonin Yes, the failure is a direct effect of that file not yet being created.

carbonin: >the failure is a direct effect of that file not yet being created.

Okay, then this is fine :+1:

jvlcek:
@carbonin and @abellotti Thank you for all the input. After some discussions with @abellotti and some more experimenting  I've provided, what I think is, a cleaner solution.

HTTPD will only be restarted if it is running after configuring external authentication.

This will allow extern auth to be configured before evmserverd.

Thank you for all the input!

JoeV

lib/gems/pending/appliance_console/external_httpd_authentication.rb
```diff
@@ -79,11 +79,12 @@ def activate
...
-      say("\nRestarting sssd and httpd ...")
-      %w(sssd httpd).each { |service| LinuxAdmin::Service.new(service).restart }
+      say("\nRestarting httpd, if running ...")
+      httpd_service = LinuxAdmin::Service.new("httpd")
+      httpd_service.restart if httpd_service.running?
 
-      say("Configuring sssd to start upon reboots ...")
-      LinuxAdmin::Service.new("sssd").enable
+      say("Restarting sssd and configure it to start on reboots ...")
+      LinuxAdmin::Service.new("sssd").enable.restart
```
abellotti: Can we do .restart.enable instead ?
i.e. we may not want to enable it on reboots if it can't restart

lib/gems/pending/appliance_console/external_httpd_authentication.rb
```diff
@@ -79,11 +79,12 @@ def activate
     end
 
     def post_activation
-      say("\nRestarting sssd and httpd ...")
-      %w(sssd httpd).each { |service| LinuxAdmin::Service.new(service).restart }
+      say("\nRestarting httpd, if running ...")
+      httpd_service = LinuxAdmin::Service.new("httpd")
+      httpd_service.restart if httpd_service.running?
```
abellotti: 👍

lib/gems/pending/appliance_console/external_httpd_authentication.rb
jvlcek: @abellotti Good idea! WIlldo

jvlcek:
@abellotti and @carbonin Let me know if this is good to merge and I'll squash the commit.

carbonin:
Looks good @jvlcek

jvlcek:
@carbonin and @abellotti Commits squashed. Ready for merge. Thank you! JoeV

PR #42, opened 2017-01-20
Prerequisites for running ansible inside the appliance

carbonin:
This PR creates the postgres database that will be used for ansible on the appliance when we create our database.

Additionally this PR adds the host and password to `database.yml` even when we are running a database locally. This was done so that the password is accessible at configuration time to provide to the ansible process so that it can contact the database.

https://www.pivotaltracker.com/story/show/137048481

### DIFF
--- a/lib/gems/pending/appliance_console/database_configuration.rb
+++ b/lib/gems/pending/appliance_console/database_configuration.rb
@@ -150,10 +150,10 @@ FRIENDLY
     def settings_hash
       {
         'adapter'  => 'postgresql',
-        'host'     => local? ? nil : host,
-        'port'     => local? ? nil : port,
+        'host'     => local? ? "localhost" : host,
+        'port'     => port,
         'username' => username,
-        'password' => local? ? nil : password.presence,
+        'password' => password.presence,
         'database' => database
       }
     end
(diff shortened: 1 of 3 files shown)


carbonin:
An alternative approach here could be to execute this stuff when we do the configuration and generate and save the db password like we do for the other keys and passwords (https://github.com/ManageIQ/manageiq/pull/13584). This way we have a separate place to deal with any password rules or special character restrictions that the ansible side may have.

Thoughts @gtanzillo @jrafanie ?

jrafanie:
> An alternative approach here could be to execute this stuff when we do the configuration and generate and save the db password like we do for the other keys and passwords (ManageIQ/manageiq#13584). This way we have a separate place to deal with any password rules or special character restrictions that the ansible side may have.
> 
> Thoughts @gtanzillo @jrafanie ?

I don't know, maybe I'm missing something but it feels like doing this as configure database time is exactly when it needs to happen, when creating the internal database.  I'm open to other ideas but I don't see why we'd do it differently.

gtanzillo:
I'm in agreement with @jrafanie. Unless there's some reason (other being uncomfortable with adding host and password to the database.yaml file) why we shouldn't or can't create the DB for ansible, I think it should be done at the same time as the vmdb DB. So I'm good with this. anyway, we already save the password when the DB is not local.

PR #45, opened 2017-01-25
Remove the awx database creation

carbonin:
This will be done when the ansible role is started from within the app itself.

This approach allows us to generate a new password rather than using the same one as our superuser. We want to do this because ansible stores its database password in plain-text on the filesystem.

Moving this functionality into https://github.com/ManageIQ/manageiq/pull/13584


jrafanie:
> This will be done when the ansible role is started from within the app itself.

When the ansible worker starts.... it will do it 😉

jrafanie:
:shipit:

carbonin:
>You did this just for the deletions PR didn't you?

You say that like I didn't just add this @jrafanie

jrafanie:
> You say that like I didn't just add this @jrafanie

A deletion is a deletion.  :1st_place_medal:

PR #46, opened 2017-01-26
Fix ruby 2.2 test failure on Tempfile.new without args

cben:
```
Failure/Error: @test_crontab_1 = Tempfile.new

    ArgumentError:
      wrong number of arguments (0 for 1..2)
```

The basename arg was required in 2.2:
http://ruby-doc.org/stdlib-2.2.5/libdoc/tempfile/rdoc/Tempfile.html#method-c-new
and became optional later:
http://ruby-doc.org/stdlib-2.4.0/libdoc/tempfile/rdoc/Tempfile.html#method-c-new
2.3 still documents it as required but de-facto works if omitted, which is why it failed for me locally, but passed on Travis.

### DIFF
--- a/spec/appliance_console/database_maintenance_periodic_spec.rb
+++ b/spec/appliance_console/database_maintenance_periodic_spec.rb
@@ -2,15 +2,15 @@ require "appliance_console/database_maintenance_periodic"
 
 describe ApplianceConsole::DatabaseMaintenancePeriodic do
   before do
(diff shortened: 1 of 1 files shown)


cben:
@miq-bot add-label bug

carbonin:
@Fryguy @bdunne if we care about ruby 2.2 should we add a travis run using that version?

PR #52, opened 2017-02-06
Set configuration for log FS by appliance_console_cli

aljesusg:
Like tmp FS I added the configuration in appliance_console_cli to set by argument logdisk

### DIFF
--- a/lib/gems/pending/appliance_console/logfile_configuration.rb
+++ b/lib/gems/pending/appliance_console/logfile_configuration.rb
@@ -15,8 +15,8 @@ module ApplianceConsole
 
     include ApplianceConsole::Logging
 
-    def initialize
-      self.disk                = nil
+    def initialize(config = {})
+      self.disk                = config[:disk]
       self.new_logrotate_count = nil
 
       self.size = MiqSystem.disk_usage(LOGFILE_DIRECTORY)[0][:total_bytes]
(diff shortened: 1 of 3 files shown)


aljesusg:
Maybe we can merge tmp and log disk configuration in the same function and toogle with a flag

lib/gems/pending/appliance_console/logfile_configuration.rb
```diff
@@ -15,8 +15,8 @@ class LogfileConfiguration
 
     include ApplianceConsole::Logging
 
-    def initialize
-      self.disk                = nil
+    def initialize(config = {})
```
carbonin: This feels unnecessary. I would prefer setting disk using the `attr_accessor` rather than passing in an options hash.

carbonin:
This PR makes some progress on this BZ as well https://bugzilla.redhat.com/show_bug.cgi?id=1385697

aljesusg:
Yeah I know @carbonin appliance_console_cli should have all operations of appliance_console but will be complicated with all options and params. Maybe we can add an argument for each tool (database, log, server ...)

carbonin:
@aljesusg Right, I'm not suggesting we attempt to implement all of the missing features in this PR, just tracking that this PR gets us some of the way there.

carbonin:
@aljesusg And yes, it might be better to refactor some of the options as they are getting a bit numerous, but I'm not sure how that will play with backwards compatibility of existing usage of the script.

I'll open an issue to discuss the changes around adding the missing functionality, then maybe we can start checking things off there.

aljesusg:
I did the changes @carbonin

aljesusg:
@carbonin please review if we need something else.

aljesusg:
Thanks a lot for the support @carbonin

PR #53, opened 2017-02-06
Start/Stop/Status server  option

aljesusg:
I added an option to start,stop or get status from evmserverd. 
issue https://github.com/ManageIQ/manageiq-gems-pending/issues/54
What do you think about?? 
I wanna have more options in CLI. Maybe can divide all operations in topics if we wanna start service or make FS or configure something, like server, database,...

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -83,6 +83,10 @@ module ApplianceConsole
       options[:extauth_opts]
     end
 
+    def set_server_state?
+      options[:server]
+    end
+
     def initialize(options = {})
       self.options = options
     end
@@ -134,6 +138,7 @@ module ApplianceConsole
         opt :postgres_server_cert, "install certs for postgres server", :type => :boolean
         opt :http_cert,            "install certs for http server",     :type => :boolean
         opt :extauth_opts,         "External Authentication Options",   :type => :string
+        opt :server,               "Server status",                     :type => :string
       end
       Trollop.die :region, "needed when setting up a local database" if options[:region].nil? && local_database?
       self
@@ -141,7 +146,8 @@ module ApplianceConsole
 
     def run
       Trollop.educate unless set_host? || key? || database? || tmp_disk? || log_disk? ||
-                             uninstall_ipa? || install_ipa? || certs? || extauth_opts? || time_zone?
+                             uninstall_ipa? || install_ipa? || certs? || extauth_opts? ||
+                             time_zone? || set_server_state?
       if set_host?
         system_hosts = LinuxAdmin::Hosts.new
         system_hosts.hostname = options[:host]
@@ -158,6 +164,7 @@ module ApplianceConsole
       install_ipa if install_ipa?
       install_certs if certs?
       extauth_opts if extauth_opts?
+      set_server_state if set_server_state?
     rescue AwesomeSpawn::CommandResultError => e
       say e.result.output
       say e.result.error
@@ -320,6 +327,21 @@ module ApplianceConsole
       extauthopts_hash = extauthopts.parse(options[:extauth_opts])
       raise "Must specify at least one external authentication option to set" unless extauthopts_hash.present?
       extauthopts.update_configuration(extauthopts_hash)
+    end
+
+    def set_server_state
+      service = LinuxAdmin::Service.new("evmserverd")
+      service_running = service.running?
+      case options[:server]
+      when "start"
+        service.start unless service_running
+      when "stop"
+        service.stop if service_running
+      when "restart"
+        service.restart
+      else
+        raise "Invalid server action"
+      end
     end
 
     def self.parse(args)
--- a/spec/appliance_console/cli_spec.rb
+++ b/spec/appliance_console/cli_spec.rb
@@ -227,6 +227,38 @@ describe ApplianceConsole::Cli do
     end
   end
 
+  context "#set_server_state" do
+    let(:evm_service) { double("evmserved") }
+    before do
+      allow(LinuxAdmin::Service).to receive(:new).with("evmserverd").and_return(evm_service)
+    end
+
+    it "state is start" do
+      expect(evm_service).to receive(:running?).and_return(false)
+      expect(evm_service).to receive(:start)
+      subject.parse(%w(--server start)).run
+    end
+
+    it "state is stop" do
+      expect(evm_service).to receive(:running?).and_return(true)
+      expect(evm_service).to receive(:stop)
+      subject.parse(%w(--server stop)).run
+    end
+
+    it "state is restart" do
+      expect(evm_service).to receive(:running?).and_return(true)
+      expect(evm_service).to receive(:restart)
+      subject.parse(%w(--server restart)).run
+    end
+
+    it "state is wrong" do
+      expect(evm_service).to receive(:running?).and_return(true)
+      expect do
+        subject.parse(%w(--server aa)).run
+      end.to raise_error(/Invalid server action/)
+    end
+  end
+
   context "#config_tmp_disk" do
     it "configures disk" do
       expect(subject).to receive(:disk_from_string).with('x').and_return('/dev/x')


lib/gems/pending/appliance_console/cli.rb
```diff
@@ -286,6 +293,15 @@ def extauth_opts
       extauthopts.update_configuration(extauthopts_hash)
     end
 
+    def server_mode
+      service = LinuxAdmin::Service.new("evmserverd")
+      case server_state
+      when :status then service.running? ? "running" : "not running"
```
carbonin: I think this method could be simplified if you checked `service.running?` outside of the case statement.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -145,6 +151,7 @@ def run
       install_ipa if install_ipa?
       install_certs if certs?
       extauth_opts if extauth_opts?
+      server_mode if server_state?
```
carbonin: I think these methods would be better if they were named as verbs.

```ruby
set_server_state if set_server_state?
```

Reads a lot better to me.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -286,6 +293,15 @@ def extauth_opts
       extauthopts.update_configuration(extauthopts_hash)
     end
 
+    def server_mode
+      service = LinuxAdmin::Service.new("evmserverd")
+      case server_state
```
carbonin: Shouldn't this be `options[:server]`?

I'm not sure where the `server_state` variable is coming from here.

lib/gems/pending/appliance_console/cli.rb
carbonin: Generally, I think this type of change throughout this file would make it a lot easier to read, but let's stick to this change for this patch.

carbonin:
This PR also makes some progress on this BZ https://bugzilla.redhat.com/show_bug.cgi?id=1385697

aljesusg:
Give me some time is in WIP @carbonin ^^ Thanks a lot for the support

spec/appliance_console/cli_spec.rb
```diff
@@ -214,6 +214,46 @@
...
+      expect(subject).to receive(:say)
+      expect(LinuxAdmin::Service).to receive(:new)
+
+      subject.parse(%w(--server status)).run
+    end
+
+    it "state is start" do
+      expect(subject).to receive(:say)
+      expect(LinuxAdmin::Service).to receive(:new)
+      expect_any_instance_of(LinuxAdmin::Service.new("evmserverd").class).to receive(:start).and_return(true)
```
chrisarcand: As noted [here](https://github.com/ManageIQ/manageiq-gems-pending/pull/53#issuecomment-277832314), let's try and avoid using `_any_instance_of`. Fortunately this case looks to be easy to fix as you're stubbing `new` directly before it; it doesn't matter much service type class is *actually* initialized, as you can just make whatever assertions you want on your test double:

```
context "#set_server_state" do
  let(:service) { double }

  it "state is start" do
    # ....
    expect(LinuxAdmin::Service).to receive(:new).with("evmserverd").and_return(service)
    expect(service).to receive(:start).and_return(true)
    # whatever other assertions you want
```

I know there are other specs in this file that use `any_instance_of` but we're trying to avoid using it entirely and not adding any more would make their removal a little easier 😄

aljesusg:
@carbonin & @chrisarcand I did the changes,Do we need anything else here ?

jrafanie:
> /cc @jrafanie @chrisarcand I think this was the issue we ran into when we merged a build with failed spec tests in it.

@carbonin wow.  Calling exit in a test is fun!

aljesusg:
I added the condition to educate @carbonin I hadn't rebase my branch  with master. How do you see now? Thanks

carbonin:
@aljesusg I'll go through and bump the tests on all the PRs when https://github.com/ManageIQ/manageiq-gems-pending/pull/62 is merged. Then we'll be able to catch these things more easily.

chessbyte:
@carbonin bump

carbonin:
@chessbyte I'm not sure what we want to do with this.

The specs still need some work in this PR as they are currently failing.
@aljesusg Do you want to finish these changes?

aljesusg:
Yes @carbonin, If you need it I'll finish them. I was working in showback with sergio ocon and I stopped this.

aljesusg:
@carbonin Ok This PR is ready for merge. Review please

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,25 @@ def extauth_opts
       extauthopts.update_configuration(extauthopts_hash)
     end
 
+    def set_server_state
+      service = LinuxAdmin::Service.new("evmserverd")
+      service_running = service.running?
+      case options[:server]
+      when :status
+        service_running ? "running" : "not running"
+        say "EvmServed is " + service.running?
```
carbonin: The service name is evmserverd - I think this is misspelled and has some odd capitalization.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,25 @@ def extauth_opts
...
+      when :start
+        service.start unless service_running
+      when :stop
+        service.stop if service_running
+      when :restart
+        say "EvmServed is restarting"
```
carbonin: I don't think we need to print anything here if we don't for the other state changes (start/stop)

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,25 @@ def extauth_opts
...
+        service.restart
+      else
+        say "wrong server state"
```
carbonin: I think `Invalid server action` would be better as the user is writing up actions rather than states.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,25 @@ def extauth_opts
       extauthopts.update_configuration(extauthopts_hash)
     end
 
+    def set_server_state
+      service = LinuxAdmin::Service.new("evmserverd")
+      service_running = service.running?
+      case options[:server]
```
carbonin: I think this will be a string at this point, so the `when` statements. Should also be strings.

spec/appliance_console/cli_spec.rb
```diff
@@ -227,6 +227,48 @@
...
+      expect(subject).to receive(:say)
+      expect(evm_service).to receive(:running?).and_return(true)
+      subject.parse(%w(--server status)).run
+    end
+
+    it "state is start" do
+      expect(subject).to receive(:say)
+      expect(evm_service).to receive(:running?).and_return(true)
+      expect(evm_service).to receive(:start)
+      evm_service.start
```
carbonin: Why are you calling methods on a double in all of these tests?

Don't do that.

lib/gems/pending/appliance_console/cli.rb
carbonin: Also, I'm not sure that we do it anywhere else, but you might want to `exit 1` here so that we get an error return code on the command line.

carbonin: Or a `raise` would work as well.

spec/appliance_console/cli_spec.rb
aljesusg: I can't test this without double with
``ruby
 context "#set_server_state" do
    before do
      allow(LinuxAdmin::Service).to receive(:new).with("evmserverd")
    end

    it "state is status" do
      expect(LinuxAdmin::Service).to receive(:new).with("evmserverd")
      expect_any_instance_of(LinuxAdmin::Service).to receive(:status).and_return(true)
      subject.parse(%w(--server status)).run
    end
```

throw the error
NoMethodError:
       undefined method `running?' for nil:NilClass

I don't know how test this kind of service without double

Comment (carbonin) [inline spec/appliance_console/cli_spec.rb]: You still need the double, but just remove the line I commented on.

You still need to `expect(evm_service).to receive(:running?).and_return(true)`, but calling `evm_service.start` is defeating the purpose of `expect(evm_service).to receive(:start)`.

Comment (aljesusg) [inline spec/appliance_console/cli_spec.rb]: I did the changfes, butin the state is start if I not call to evm_service it throw this error....I removed all call methods of double in the tests
```ruby
    (Double "evmserved").start(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments
```

spec/appliance_console/cli_spec.rb
```diff
@@ -227,6 +227,43 @@
...
+    end
+    it "state is status" do
+      expect(evm_service).to receive(:running?).and_return(true)
+      subject.parse(%w(--server status)).run
+    end
+
+    it "state is start" do
+      expect(evm_service).to receive(:running?).and_return(true)
+      expect(evm_service).to receive(:start)
+      evm_service.start
```
carbonin: This line also should be removed.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,23 @@ def extauth_opts
...
+      when "status"
+        service_running ? "running" : "not running"
+      when "start"
+        service.start unless service_running
+      when "stop"
+        service.stop if service_running
+      when "restart"
+        service.restart
+      else
+        raise "Wrong server state"
```
carbonin: I think "Invalid server action" would be a better message here.

lib/gems/pending/appliance_console/cli.rb
carbonin: 👍 Thanks

spec/appliance_console/cli_spec.rb
aljesusg: I told you in old messages, if I remove this

```ruby
1) ApplianceConsole::Cli#set_server_state state is start
     Failure/Error: expect(evm_service).to receive(:start)

       (Double "evmserved").start(*(any args))
           expected: 1 time with any arguments
           received: 0 times with any arguments
     # ./plugins/manageiq-gems-pending/spec/appliance_console/cli_spec.rb:242:in `block (3 levels) in <top (required)>'
```

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -322,6 +329,23 @@ def extauth_opts
       extauthopts.update_configuration(extauthopts_hash)
     end
 
+    def set_server_state
+      service = LinuxAdmin::Service.new("evmserverd")
+      service_running = service.running?
+      case options[:server]
+      when "status"
+        service_running ? "running" : "not running"
```
carbonin: It looks like this doesn't really do anything. I think we can probably just remove the status case.

What do you think @aljesusg ?

aljesusg: Yeah...This before says the sate of running but we removed all the says methods so...If you want I can remove it  @carbonin

aljesusg:
@carbonin I did the changes

aljesusg:
Thanks @carbonin

Issue #54, opened 2017-02-06
Add missing features to appliance_console_cli


carbonin:
https://bugzilla.redhat.com/show_bug.cgi?id=1385697 was opened to request adding features to the `appliance_console_cli` that are currently only available in the interactive console.

It will be easier to add things here in a check list and track each missing feature separately.
Please add to this list if I missed any.

- [ ] evmserverd operations (https://github.com/ManageIQ/manageiq-gems-pending/pull/53)
- [x] Log file disk configuration (https://github.com/ManageIQ/manageiq-gems-pending/pull/52)
- [ ] Log rotation configuration
- [ ] Database HA
  - [ ] Create standby
  - [ ] Create primary
  - [ ] Start failover monitor
- [ ] Database maintenance
- [ ] SCAP
- [ ] Network config
  - [ ] DHCP
  - [ ] Static
  - [ ] Test network config
- [x] Set timezone
- [ ] Set date and time
- [ ] Database restore

Fryguy:
@miq-bot move_issue manageiq-appliance_console

PR #56, opened 2017-02-07
[WIP] Set date & time

aljesusg:
Added option to set date and time by CLI . Issue #54

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -79,6 +79,14 @@ module ApplianceConsole
       options[:timezone]
     end
 
+    def date_time?
+      options[:date] && options[:time]
+    end
+
+    def time_sync?
+      options[:timesync]
+    end
+
     def extauth_opts?
       options[:extauth_opts]
     end
@@ -128,6 +136,9 @@ module ApplianceConsole
         opt :ipapassword,   "IPA Server password",  :type => :string
         opt :ipadomain,     "IPA Server domain (optional)", :type => :string
         opt :iparealm,      "IPA Server realm (optional)", :type => :string
+        opt :date,                 "Date in YYYY-MM-DD",                     :type => :string
+        opt :time,                 "Time in HH:MM:SS",                       :type => :string
+        opt :timesync,             "Automatic time synchronization",         :type => :boolean, :default => false
         opt :ca,                   "CA name used for certmonger",       :type => :string,  :default => "ipa"
         opt :timezone,             "Time zone",                         :type => :string
         opt :postgres_client_cert, "install certs for postgres client", :type => :boolean
@@ -141,7 +152,8 @@ module ApplianceConsole
 
     def run
       Trollop.educate unless set_host? || key? || database? || tmp_disk? || log_disk? ||
-                             uninstall_ipa? || install_ipa? || certs? || extauth_opts? || time_zone?
+                             uninstall_ipa? || install_ipa? || certs? || extauth_opts? ||
+                             time_zone? || date_time?
       if set_host?
         system_hosts = LinuxAdmin::Hosts.new
         system_hosts.hostname = options[:host]
@@ -157,6 +169,7 @@ module ApplianceConsole
       uninstall_ipa if uninstall_ipa?
       install_ipa if install_ipa?
       install_certs if certs?
+      set_date_time if date_time? || time_sync?
       extauth_opts if extauth_opts?
     rescue AwesomeSpawn::CommandResultError => e
       say e.result.output
@@ -226,6 +239,20 @@ module ApplianceConsole
         say("Timezone configured")
       else
         say("Timezone not configured")
+      end
+    end
+
+    def set_date_time
+      @new_date = options[:date]
+      @new_time = options[:time]
+      if @new_date =~ ApplianceConsole::DateTimeConfiguration::DATE_REGEXP &&
+         @new_time =~ ApplianceConsole::DateTimeConfiguration::TIME_REGEXP
+        say "Configuring date and time"
+        @manual_time_sync = options[:timesync]
+        config_datetime = ApplianceConsole::DateTimeConfiguration.new
+        raise "Failed setting date time." unless config_datetime.activate
+      else
+        raise "Failed to configure date time. Wrong format."
       end
     end
 
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -83,10 +83,6 @@ module ApplianceConsole
       options[:date] && options[:time]
     end
 
-    def time_sync?
-      options[:timesync]
-    end
-
     def extauth_opts?
       options[:extauth_opts]
     end
@@ -169,7 +165,7 @@ module ApplianceConsole
       uninstall_ipa if uninstall_ipa?
       install_ipa if install_ipa?
       install_certs if certs?
-      set_date_time if date_time? || time_sync?
+      set_date_time if date_time?
       extauth_opts if extauth_opts?
     rescue AwesomeSpawn::CommandResultError => e
       say e.result.output
@@ -243,14 +239,22 @@ module ApplianceConsole
     end
 
     def set_date_time
-      @new_date = options[:date]
-      @new_time = options[:time]
-      if @new_date =~ ApplianceConsole::DateTimeConfiguration::DATE_REGEXP &&
-         @new_time =~ ApplianceConsole::DateTimeConfiguration::TIME_REGEXP
+      if options[:date] =~ ApplianceConsole::DateTimeConfiguration::DATE_REGEXP &&
+          options[:time] =~ ApplianceConsole::DateTimeConfiguration::TIME_REGEXP
         say "Configuring date and time"
-        @manual_time_sync = options[:timesync]
         config_datetime = ApplianceConsole::DateTimeConfiguration.new
-        raise "Failed setting date time." unless config_datetime.activate
+        begin
+          options[:timesync] ? config_datetime.enable_auto_sync : config_datetime.disable_auto_sync
+        rescue
+          raise "Failed to configure auto sync"
+        end
+        if options[:timesync]
+          begin
+            LinuxAdmin::TimeDate.system_time = Time.parse("#{options[:date]} #{options[:time]}").getlocal
+          rescue
+            raise "Failed to configure date time"
+          end
+        end
       else
         raise "Failed to configure date time. Wrong format."
       end
--- a/spec/appliance_console/cli_spec.rb
+++ b/spec/appliance_console/cli_spec.rb
@@ -144,13 +144,13 @@ describe ApplianceConsole::Cli do
 
     it "should set date and time with auto sync" do
       expect(subject).to receive(:say)
-      expect(datetime_configuration).to receive(:activate).and_return(true)
+      expect(datetime_configuration).to receive(:enable_auto_sync).and_return(true)
       subject.parse(%w(--date 2017-02-07 --time 12:00:00 --timesync true")).run
     end
 
     it "should set date and time without auto sync" do
       expect(subject).to receive(:say)
-      expect(datetime_configuration).to receive(:activate).and_return(true)
+      expect(datetime_configuration).to receive(:disable_auto_sync).and_return(true)
       subject.parse(%w(--date 2017-02-07 --time 12:00:00)).run
     end
 
--- a/spec/appliance_console/cli_spec.rb
+++ b/spec/appliance_console/cli_spec.rb
@@ -1,5 +1,6 @@
 require "appliance_console/cli"
 require "appliance_console/timezone_configuration"
+require "appliance_console/date_time_configuration"
 
 describe ApplianceConsole::Cli do
   subject { described_class.new }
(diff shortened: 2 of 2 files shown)


aljesusg:
@miq-bot add_label bug, enhancement, appliance/console
@miq-bot  assign @carbonin

aljesusg:
@miq-bot remove_label wip

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -70,6 +70,14 @@ def tmp_disk?
       options[:tmpdisk]
     end
 
+    def date_time?
```
carbonin: Can we make these more action based names? I feel like `set_date_time?` and `set_time_auto_sync?` would read better.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
       config.post_activation
     end
 
+    DATE_REGEXP   = /^(2[0-9]{3})-(0?[1-9]|1[0-2])-(0?[1-9]|[12][0-9]|3[01])/
+    TIME_REGEXP   = /^(0?[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])/
```
carbonin: You should be able to reference these as `ApplianceConsole::DateTimeConfiguration::DATE_REGEXP`, and `ApplianceConsole::DateTimeConfiguration::TIME_REGEXP`. That way if we ever need to change them, we only have to do it in one place.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
...
+
+    def set_date_time
+      date, time = options[:datetime].split("T")
+      if date.validate(DATE_REGEXP) && time.validate(TIME_REGEXP)
+        say "Configuring date and time"
+        autosync = auto_sync? ? options[:autosync] : false
```
carbonin: This line is not necessary, `options[:autosync]` will always be either `true` or `false` so you can pass it directly.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -144,6 +154,7 @@ def run
       uninstall_ipa if uninstall_ipa?
       install_ipa if install_ipa?
       install_certs if certs?
+      set_date_time if date_time?
```
carbonin: This only allows you to configure time sync if you are also setting the date and time. I feel like this should be:

```ruby
set_date_time if date_time? || auto_sync?
```

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -118,11 +126,13 @@ def parse(args)
         opt :ipapassword,   "IPA Server password",  :type => :string
         opt :ipadomain,     "IPA Server domain (optional)", :type => :string
         opt :iparealm,      "IPA Server realm (optional)", :type => :string
-        opt :ca,                   "CA name used for certmonger",       :type => :string,  :default => "ipa"
-        opt :postgres_client_cert, "install certs for postgres client", :type => :boolean
-        opt :postgres_server_cert, "install certs for postgres server", :type => :boolean
-        opt :http_cert,            "install certs for http server",     :type => :boolean
-        opt :extauth_opts,         "External Authentication Options",   :type => :string
+        opt :datetime,             "DateTime in YYYY-MM-DDTHH:MM:SS format", :type => :string
+        opt :autosync,             "Automatic time synchronization",         :type => :boolean
```
carbonin: The way this is now, I'm not sure how I would be able to turn auto time sync off using the console. Maybe the `--autosync` option should be a string and take a value `on` or `off`.

Also I think autosync isn't really descriptive enough. Maybe `timesync` instead?

lib/gems/pending/appliance_console/cli.rb
carbonin: Of course this will change if we go the route of using a string for the sync option.

lib/gems/pending/appliance_console/date_time_configuration.rb
```diff
@@ -12,10 +12,10 @@ class DateTimeConfiguration
 
     include ApplianceConsole::Logging
 
-    def initialize
-      @new_date         = nil
-      @new_time         = nil
-      @manual_time_sync = false
+    def initialize(config = {})
```
carbonin: I would prefer using the existing `attr_accessor`s than adding a hash here.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
       config.post_activation
     end
 
+    DATE_REGEXP   = /^(2[0-9]{3})-(0?[1-9]|1[0-2])-(0?[1-9]|[12][0-9]|3[01])/
+    TIME_REGEXP   = /^(0?[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])/
+
+    def set_date_time
+      date, time = options[:datetime].split("T")
+      if date.validate(DATE_REGEXP) && time.validate(TIME_REGEXP)
```
carbonin: I think this would be better using the `=~` operator.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
...
+        say "Configuring date and time"
+        autosync = auto_sync? ? options[:autosync] : false
+        config_datetime = ApplianceConsole::DateTimeConfiguration.new(:date      => date,
```
carbonin: Where do you call `#activate`? This looks like it will create an object, but won't actually set the date or time.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
       config.post_activation
     end
 
+    DATE_REGEXP   = /^(2[0-9]{3})-(0?[1-9]|1[0-2])-(0?[1-9]|[12][0-9]|3[01])/
+    TIME_REGEXP   = /^(0?[0-9]|1[0-9]|2[0-3]):([0-5][0-9]):([0-5][0-9])/
+
+    def set_date_time
+      date, time = options[:datetime].split("T")
```
carbonin: This is a weird interface. Maybe a better choice would be to have separate `--date` `--time` and `--timesync` options.

Plus this feels like it makes it impossible to set the date without also setting the time and vice versa.

carbonin:
@aljesusg Can you make sure you test this on an appliance? It looks like this one would not have actually done what we need it to do.

lib/gems/pending/appliance_console/cli.rb
Fryguy: Why not just use real Date/Time objects? instead of hand parsing?

lib/gems/pending/appliance_console/date_time_configuration.rb
Fryguy: ¯\\\_(ツ)_/¯ I am actually ok with this.

lib/gems/pending/appliance_console/date_time_configuration.rb
```diff
@@ -12,10 +12,10 @@ class DateTimeConfiguration
...
+      @new_date         = config.has_key(:date) ? config[:date] : nil
```
Fryguy: `has_key` should be `key?` ... not sure what class this is, but Hashes have the `key?` method which is preferable.

Fryguy: Also if a Hash, config[:date] is nil if the key doesn't exist, so there's no need for the key check at all given your default...

lib/gems/pending/appliance_console/date_time_configuration.rb
```diff
@@ -12,10 +12,10 @@ class DateTimeConfiguration
...
+      @new_time         = config.has_key(:time) ? config[:time] : nil
+      @manual_time_sync = config.has_key(:auto_sync) ? config[:auto_sync] : false
```
Fryguy: For this one, if config is a Hash, then it's preferable to do `config.fetch(:auto_sync, false)`

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -207,6 +218,27 @@ def set_external_db
...
+                                                                      :auto_sync => autosync)
+        raise "Failed setting date time." unless config_datetime.activate
+      else
+        raise "Failed to configure date time. Wrong format."
+      end
+    end
+
+    def validate(regex)
+      !self[regex].nil?
+    end
```
Fryguy: This feels unnecessary...for conditional purposes you don't need to explicitly check nil, so the callers are currently this:

`date.validate(DATE_REGEXP) && time.validate(TIME_REGEXP)`

but can just as easily be this without introducing a helper method:

`date =~ DATE_REGEXP && time =~ TIME_REGEXP`

lib/gems/pending/appliance_console/cli.rb
carbonin: I think this was mostly lifted from the `DateTimeConfiguration` class

lib/gems/pending/appliance_console/date_time_configuration.rb
carbonin: I know it's being done in other places... I just don't really like options hashes. It's to easy for it it get out of hand. Maybe kwargs?

lib/gems/pending/appliance_console/cli.rb
aljesusg: ok, time sync is False by fedault in the appliance, I'll make the change for on or off

lib/gems/pending/appliance_console/cli.rb
aljesusg: why set?  this function is to check if there is the option datetime no for set it

lib/gems/pending/appliance_console/cli.rb
aljesusg: Ok... I though that I needed date and time for autosync but not...I'll make the change

lib/gems/pending/appliance_console/cli.rb
aljesusg: Yeah....in configure_date_time function

LinuxAdmin::TimeDate.system_time = Time.parse("#{new_date} #{new_time}").getlocal

I should pass the values in string type @Fryguy

lib/gems/pending/appliance_console/cli.rb
aljesusg: in unless 
raise "Failed setting date time." unless config_datetime.activate

that function return true or false if success

lib/gems/pending/appliance_console/cli.rb
aljesusg: Yeah I didn't reme,mber this operator. Thanks @Fryguy

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -214,6 +226,22 @@ def set_external_db
...
 
+    def set_date_time
+      date = options[:date]
+      time = options[:time]
+      if date =~ ApplianceConsole::DateTimeConfiguration::DATE_REGEXP &&
+         time =~ ApplianceConsole::DateTimeConfiguration::TIME_REGEXP
+        say "Configuring date and time"
+        @new_date = date
+        @new_time = time
+        @manual_time_sync = options[:timesync]
```
carbonin: Why are these instance variables?

I think these should probably be setting the instance variables of `config_datetime` below.
As of now it looks like this would unconditionally set the time sync to manual and then fail to set the date and time as those fields in the `DateTimeConfiguration` object would be `nil`.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -157,6 +169,7 @@ def run
       uninstall_ipa if uninstall_ipa?
       install_ipa if install_ipa?
       install_certs if certs?
+      set_date_time if date_time? || time_sync?
```
carbonin: I think this should only check `date_time?`.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -79,6 +79,14 @@ def time_zone?
       options[:timezone]
     end
 
+    def date_time?
+      options[:date] && options[:time]
+    end
+
+    def time_sync?
+      options[:timesync]
```
carbonin: This doesn't really make sense.

I would interpret this method's use to be "Do I need to set the time sync status". When really it's saying "Should I enable automatic time sync".

I think the issue is that this feels like there is no way to express "I want to set the automatic time sync to off"

Really, because you always need a value for this when setting the date and time (and because the option defaults to `false`) this should probably just be removed.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -229,6 +242,20 @@ def set_time_zone
       end
     end
 
+    def set_date_time
+      @new_date = options[:date]
+      @new_time = options[:time]
```
carbonin: These still don't need to be instance variables here and the values still need to be set on the `config_datetime` object before calling `config_datetime.activate` otherwise this will do nothing.

aljesusg: I did the changes and change the function @carbonin

chessbyte:
@aljesusg The appliance console has been extracted to a new [repo](https://github.com/ManageIQ/manageiq-appliance_console).  Please move this PR there.

PR #58, opened 2017-02-09
Make CLI specs green again

carbonin:
Fix the build failures introduced in #52

These got past travis because of an issue where the build goes green even though tests have failed.


jrafanie:
Thanks @carbonin

PR #61, opened 2017-02-14
Add a RewriteCond for '/cws'

petervo:


bdunne:
@petervo Can you add a description?  What is `cws`?

petervo:
cockpit web service. This is related to #12506 that's not getting merged right now anyways.

petervo:
Now that ManageIQ/manageiq#12506 is in, this needs to get in as well.

carbonin:
@petervo Sorry for the delay. I think this PR needs to move to an edit in [this file](https://github.com/ManageIQ/manageiq-appliance/blob/master/COPY/etc/httpd/conf.d/manageiq-redirects-ui) in the manageiq-appliance repo.

We made all of these static in https://github.com/ManageIQ/manageiq/pull/15147 and https://github.com/ManageIQ/manageiq-appliance/pull/124

PR #62, opened 2017-02-14
Prevent Trollop from exiting the rspec process

carbonin:
`Trollop.die` and `Trollop.educate` exit the calling process with a 0 exit status.

This is undesirable as it will cause the spec suite to go green without having run all the specs. In addition, the writer of a spec that would cause the process to exit likely did not want that behavior and would expect the spec to fail.

This change achieves that behavior by stubbing the calls to `Trollop.die` and `Trollop.educate` to raise errors.

Now we can also test that these methods are called by expecting that these particular error classes are raised.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -19,21 +19,9 @@ Dir.mkdir(RAILS_ROOT.join("log"))
 # in ./support/ and its subdirectories.
 Dir[File.expand_path(File.join(__dir__, 'support/**/*.rb'))].each { |f| require f }
 
-class TrollopEducateSpecError < StandardError; end
-class TrollopDieSpecError < StandardError; end
-
 RSpec.configure do |config|
   config.after(:each) do
     Module.clear_all_cache_with_timeout if Module.respond_to?(:clear_all_cache_with_timeout)
-  end
-
-  config.before(:each) do
-    err_string = <<-EOF.strip_heredoc
-      Don't allow methods that exit the calling process to be executed in specs.
-      If you were testing that we call Trollop.educate or Trollop.die, expect that a TrollopEducateSpecError or TrollopDieSpecError be raised instead
-    EOF
-    allow(Trollop).to receive(:educate).and_raise(TrollopEducateSpecError.new(err_string))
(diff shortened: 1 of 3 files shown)


spec/spec_helper.rb
```diff
@@ -19,11 +19,23 @@
...
 
+  config.before(:each) do
+    err_string = <<-EOF.strip_heredoc
+      Don't allow methods that exit the calling process to be executed in specs.
+      If you were testing that we call Trollop.educate or Trollop.die, expect that a TrollopEducateSpecError or TrollopDieSpecError be raised instead
+    EOF
+    allow(Trollop).to receive(:educate).and_raise(TrollopEducateSpecError.new(err_string))
+    allow(Trollop).to receive(:die).and_raise(TrollopDieSpecError.new(err_string))
+  end
+
```
jrafanie: Maybe put this in spec/support/trollop.rb?

I think you can do `RSpec.configure` in that file... I think.  Maybe?  😕

bdunne: Sounds reasonable, it should get loaded by https://github.com/ManageIQ/manageiq-gems-pending/pull/62/files#diff-93830fa29d616f7c87903d08b5b1b29aL20

spec/spec_helper.rb
```diff
@@ -19,11 +19,23 @@
...
 
+class TrollopEducateSpecError < StandardError; end
+class TrollopDieSpecError < StandardError; end
+
 RSpec.configure do |config|
   config.after(:each) do
     Module.clear_all_cache_with_timeout if Module.respond_to?(:clear_all_cache_with_timeout)
   end
 
+  config.before(:each) do
```
bdunne: Can this be a `before(:suite)` for better performance?

carbonin: No, this has to be per-test. The error I get from rspec if I do `before(:suite)` or `before(:all)` is:

>The use of doubles or partial doubles from rspec-mocks outside of the per-test lifecycle is not supported.

jrafanie: That makes sense.  I'd be afraid if mocks/doubles lasted through multiple tests.

PR #63, opened 2017-02-15
Add MiqPassword.encrypt_hash & decrypt_hash.

lfu:
These methods are used for the password fields in options of ServiceTemplate and Service.
 
https://www.pivotaltracker.com/n/projects/1937537/stories/139479571

Part of #https://github.com/ManageIQ/manageiq/pull/13934

cc @gmcculloug @bzwei

### DIFF
--- a/lib/gems/pending/util/miq-password.rb
+++ b/lib/gems/pending/util/miq-password.rb
@@ -2,6 +2,7 @@ require 'ezcrypto'
 require 'encryption/CryptString'
 require 'base64'
 require 'yaml'
+require 'util/extensions/miq-deep'
 
 class MiqPassword
   class MiqPasswordError < StandardError; end
@@ -70,6 +71,32 @@ class MiqPassword
 
   def self.encrypted?(str)
     !!split(str).first
+  end
+
+  def self.encrypt_hash(values, &block)
+    hash_values = values.deep_clone
+
+    hash_values.each { |k, val| hash_values[k] = encrypt_hash(val, &block) if val.kind_of?(Hash) }
+
+    hash_values.select { |f, _| yield f }.each do |key, val|
+      next if val.strip.empty?
+      hash_values[key] = try_encrypt(val)
+    end
+
+    hash_values
+  end
+
+  def self.decrypt_hash(values, &block)
+    hash_values = values.deep_clone
+
+    hash_values.each { |k, val| hash_values[k] = decrypt_hash(val, &block) if val.kind_of?(Hash) }
+
+    hash_values.select { |f, _| yield f }.each do |key, val|
+      next if val.strip.empty?
+      hash_values[key] = try_decrypt(val)
+    end
+
+    hash_values
   end
 
   def self.md5crypt(str)
--- a/spec/util/miq-password_spec.rb
+++ b/spec/util/miq-password_spec.rb
@@ -353,6 +353,37 @@ describe MiqPassword do
     end
   end
 
+  describe '.encrypt_hash' do
+    let(:reg_password_field) { /[Pp]assword/ }
+
+    it 'should not encrypt blank' do
+      options = { 'my_password' => '' }
+      results = MiqPassword.encrypt_hash(options) { |key| key =~ reg_password_field }
+      expect(results['my_password']).to be_empty
+    end
+
+    it "should encrypt all keys containing 'password'" do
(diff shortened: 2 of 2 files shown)


spec/util/miq-password_spec.rb
```diff
@@ -2,6 +2,7 @@
 
 require 'util/miq-password'
 require 'tempfile'
+require 'util/extensions/miq-deep'
```
carbonin: I feel like this should be included in `miq-password.rb` as that's where `deep_clone` is being used.

spec/util/miq-password_spec.rb
```diff
@@ -353,6 +354,54 @@
...
+      expect(new["my_password"]).to be_empty
+      expect(new.keys.include?("password::my_password")).to be false
+    end
+
+    it "should add 'password::' prefix" do
+      options = { "my_password" => "secret" }
+      new = MiqPassword.encrypt_hash(options)
+      expect(new.keys.first).to eq("password::my_password")
+      expect(new["password::my_password"]).to be_encrypted("secret")
+      expect(new.keys.include?("my_password")).to be false
```
carbonin: Please don't use keywords as variable names, it makes things like this ambiguous.

lfu: Thought there was something unusual 👍

lib/gems/pending/util/miq-password.rb
```diff
@@ -72,6 +73,40 @@ def self.encrypted?(str)
...
   end
 
+  def self.encrypt_hash(values)
+    hash_values = values.deep_clone
+
+    hash_values.select { |f, _| f =~ /[Pp]assword/ }.each do |key, val|
+      next if val.strip.empty?
+
+      key_in_symbol = key.kind_of?(Symbol)
+      hash_values.delete(key)
```
bzwei: can you double check if it is safe to delete/insert an hash entry within an iteration?

lib/gems/pending/util/miq-password.rb
```diff
@@ -72,6 +73,40 @@ def self.encrypted?(str)
...
+
+      key = "password::#{key}" unless key =~ /password::/
```
bzwei: Need to rethink the scenario where key already starts with `password::` - here the key name is not changed, but the value is encrypted. It is no longer symmetric. Then the `#decrpyt_hash` method will not get back exactly the same

carbonin: I'm not actually a huge fan of this in general. `try_decrypt` already checks to see if the thing you gave it is encrypted so it doesn't feel like there should be a reason to use this to indicate that something is a password field. It feels like keeping the key regex's the same would be enough.

lib/gems/pending/util/miq-password.rb
```diff
@@ -72,6 +73,40 @@ def self.encrypted?(str)
     !!split(str).first
   end
 
+  def self.encrypt_hash(values)
+    hash_values = values.deep_clone
+
+    hash_values.select { |f, _| f =~ /[Pp]assword/ }.each do |key, val|
```
carbonin: Regarding https://github.com/ManageIQ/manageiq-gems-pending/pull/63/files#r101402312 I think this regex should be put into a constant or method and used by both `encrypt_hash` and `decrypt_hash`.

lib/gems/pending/util/miq-password.rb
kbrock: This is a slippery slope, knowing what is a password.

Wonder if it makes sense to look into `Vmdb::Settings::Walker` - possibly moving that walker into here

lib/gems/pending/util/miq-password.rb
kbrock: Please don't hardcode the name of passwords in here.

lib/gems/pending/util/miq-password.rb
```diff
@@ -72,6 +73,28 @@ def self.encrypted?(str)
     !!split(str).first
   end
 
+  def self.encrypt_hash(values, &block)
```
bzwei: Please add comments on what do you expect from the block - the purpose and the return type.

kbrock: no - please don't enhance `encrypt_hash` - please use our existing mechanisms.

Fryguy:
I second the sentiments of @kbrock and @carbonin .  This feels like the wrong solution.  Hash walking is not the domain of MiqPassword.

PR #66, opened 2017-02-20
Look for productized translation files under productization directory of the application

gtanzillo:
Now that gems/pending is a real gem, the productized translation files are no longer existing in a path relative to the ruby script that loads it in the gem. The files will now created under the productization directory during the build. We need to load files from that location as well.

https://bugzilla.redhat.com/show_bug.cgi?id=1422433

/cc @carbonin

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -27,8 +27,11 @@ include HighLine::SystemExtensions
 
 require 'i18n'
 locales_dir = ENV['CONTAINER'] ? "container" : "appliance"
-LOCALES = File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__)
-I18n.load_path = Dir[LOCALES].sort
+locales_paths = [
+  File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__),
+  File.expand_path(File.join("productization/appliance_console/locales", locales_dir, "*.yml"), RAILS_ROOT)
+]
+locales_paths.each { |p| I18n.load_path += Dir[p].sort }
 I18n.enforce_available_locales = true
 I18n.backend.load_translations
 
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -35,6 +35,8 @@ locales_paths.each { |p| I18n.load_path += Dir[p].sort }
 I18n.enforce_available_locales = true
 I18n.backend.load_translations
 
+SCAP_RULES_DIR = File.expand_path("productization/appliance_console/config", RAILS_ROOT)
+
 $terminal.wrap_at = 80
 $terminal.page_at = 35
 
@@ -595,7 +597,7 @@ Static Network Configuration
 
       when I18n.t("advanced_settings.scap")
         say("#{selection}\n\n")
-        ApplianceConsole::Scap.new.lockdown
+        ApplianceConsole::Scap.new(SCAP_RULES_DIR).lockdown
         press_any_key
 
       when I18n.t("advanced_settings.summary")
--- a/lib/gems/pending/appliance_console/scap.rb
+++ b/lib/gems/pending/appliance_console/scap.rb
@@ -2,6 +2,10 @@ require 'linux_admin'
 
 module ApplianceConsole
   class Scap
+    def initialize(rules_dir)
+      @rules_dir = rules_dir
+    end
+
     def lockdown
       if packages_installed? && config_exists?
         say("Locking down the appliance for SCAP...")
@@ -20,7 +24,7 @@ module ApplianceConsole
     private
 
     def yaml_filename
-      File.expand_path("config/scap_rules.yml", __dir__)
+      File.expand_path("scap_rules.yml", @rules_dir)
     end
 
     def packages_installed?


lib/gems/pending/appliance_console.rb
```diff
@@ -27,8 +27,9 @@
 
 require 'i18n'
 locales_dir = ENV['CONTAINER'] ? "container" : "appliance"
-LOCALES = File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__)
-I18n.load_path = Dir[LOCALES].sort
+locales_paths =  [File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__)]
+locales_paths += [File.expand_path(File.join("gems/pending/appliance_console/locales", locales_dir, "*.yml"), RAILS_ROOT)]
```
Fryguy: is this is because of productization and where it drops the file?

Fryguy:
It feels like this is productization's problem, and not a problem we should fix here.  That is, it should drop the file into the right location with respect to the gem, instead of into the rails app.

lib/gems/pending/appliance_console.rb
gtanzillo: Yes, that's right @Fryguy. That path that I added was the original location.

Fryguy:
@gtanzillo I'm thinking that maybe since this is part of productizing, that we should reuse one of the productization directories.

So, since `/var/www/miq/vmdb/productization` is a known place, perhaps `/var/www/miq/vmdb/productization/appliance_console/locales` ?

Alternately, perhaps it's nicer if this is provided via manageiq-appliance repo's productization (can't remember how that one works).

lib/gems/pending/appliance_console/scap.rb
```diff
@@ -20,7 +22,7 @@ def lockdown
     private
 
     def yaml_filename
-      File.expand_path("config/scap_rules.yml", __dir__)
+      File.expand_path("productization/appliance_console/config/scap_rules.yml", RAILS_ROOT)
```
carbonin: I think this would be better if we just passed the config file or directory in `initialize`; that way we could avoid spreading the knowledge of where `RAILS_ROOT` may be.

gtanzillo:
Created https://gitlab.cloudforms.lab.eng.rdu2.redhat.com/cloudforms/cfme_productization/merge_requests/421 for productization changes that this depends on.

lib/gems/pending/appliance_console/scap.rb
```diff
@@ -20,7 +24,7 @@ def lockdown
     private
 
     def yaml_filename
-      File.expand_path("config/scap_rules.yml", __dir__)
+      File.expand_path(File.join(@rules_dir, "scap_rules.yml"))
```
Fryguy: Much cleaner to do:"

```ruby
File.expand_path("scap_rules.yml", @rules_dir)
```

lib/gems/pending/appliance_console/scap.rb
```diff
@@ -2,6 +2,10 @@
 
 module ApplianceConsole
   class Scap
+    def initialize(options)
+      @rules_dir = options[:rules_dir]
+    end
```
Fryguy: If the rules_dir is not specified, then doesn't the below code fail?  Should this be a required parameter, and thus not options?

lib/gems/pending/appliance_console.rb
```diff
@@ -597,7 +600,9 @@ module ApplianceConsole
 
       when I18n.t("advanced_settings.scap")
         say("#{selection}\n\n")
-        ApplianceConsole::Scap.new.lockdown
+        ApplianceConsole::Scap.new(
+          :rules_dir => File.expand_path("productization/appliance_console/config", RAILS_ROOT)
+        ).lockdown
```
Fryguy: This feels weird that we enforce productization here, but I guess it will have to do.

lib/gems/pending/appliance_console.rb
```diff
@@ -27,8 +27,11 @@
 
 require 'i18n'
 locales_dir = ENV['CONTAINER'] ? "container" : "appliance"
-LOCALES = File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__)
-I18n.load_path = Dir[LOCALES].sort
+locales_paths = [
+  File.expand_path(File.join("appliance_console/locales", locales_dir, "*.yml"), __dir__),
+  File.expand_path(File.join("productization/appliance_console/locales", locales_dir, "*.yml"), RAILS_ROOT)
+]
+locales_paths.each { |p| I18n.load_path += Dir[p].sort }
```
Fryguy: This all looks good to me :+1:

gtanzillo:
@Fryguy Can you merge this one since @bdunne will be away for a couple of weeks.

Fryguy:
@gtanzillo I had a couple comments...did you want to clean them up?

lib/gems/pending/appliance_console.rb
gtanzillo: @Fryguy I agree. Not sure how much it will help but, I can put this path in a constant (like `SCAP_RULES_DIR`) at the top of the file and just reference that here. What do you think?

lib/gems/pending/appliance_console/scap.rb
gtanzillo: I'll make this change.

gtanzillo:
@Fryguy I just made some updates, please take another look. Thanks.

PR #67, opened 2017-02-21
Update `ovirt` gem to 0.15.0

jhernand:
This new version of the gem is needed because it introduces support for
checking the TLS certificates of the server.


jhernand:
This is a requirement of ManageIQ/manageiq#14004.

@Fryguy please review.

/cc @cben

jhernand:
Thanks @Fryguy .

PR #68, opened 2017-02-22
Extract Networking Configuration to a submenu

isimluk:
There were already 4 options for network configuration. With IPv6 this will be extended further.

Logically these 4 options are tight together and they are unrelated from other options. Once the user is done with networking, he or she may not want to return back.

I have been thinking about this and I wanted to avoid this move for a while. Then, I have come to a conclusion that this should pay off in the long run. There is growing ammount of options in the
appliance_console and separation will make it easier for mammals to comprehend.

This is how it looks now

```
Advanced Setting

1) Set-up Networking
2) Set Timezone
3) Set Date and Time
4) Restore Database From Backup
5) Configure Database
6) Configure Database Replication
7) Configure Database Maintenance
8) Logfile Configuration
9) Configure Application Database Failover Monitor
10) Configure External Authentication (httpd)
11) Update External Authentication Options
12) Generate Custom Encryption Key
13) Stop EVM Server Processes
14) Start EVM Server Processes
15) Restart Appliance
16) Shut Down Appliance
17) Summary Information
18) Quit
```
and
```
Network Configuration

1) Set DHCP Network Configuration
2) Set Static Network Configuration
3) Test Network Configuration
4) Set Hostname

Choose the network configuration: 
```

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -150,37 +150,44 @@ To modify the configuration, use a web browser to access the management page.
       clear_screen
       selection = ask_with_menu("Advanced Setting", AS_OPTIONS, nil, true)
       case selection
-      when I18n.t("advanced_settings.dhcp")
-        say("DHCP Network Configuration\n\n")
-        if agree("Apply DHCP network configuration? (Y/N): ")
-          say("\nApplying DHCP network configuration...")
+      when I18n.t('advanced_settings.networking')
+        options = {
+          'Set DHCP Network Configuration'   => 'dhcp',
+          'Set Static Network Configuration' => 'static',
+          'Test Network Configuration'       => 'testnet',
+          'Set Hostname'                     => 'hostname'
+        }
+        case ask_with_menu('Network Configuration', options)
+        when 'dhcp'
+          say("DHCP Network Configuration\n\n")
+          if agree("Apply DHCP network configuration? (Y/N): ")
+            say("\nApplying DHCP network configuration...")
 
-          resolv = LinuxAdmin::Dns.new
-          resolv.search_order = []
-          resolv.nameservers = []
-          resolv.save
+            resolv = LinuxAdmin::Dns.new
+            resolv.search_order = []
+            resolv.nameservers = []
+            resolv.save
 
-          eth0.enable_dhcp
-          eth0.save
+            eth0.enable_dhcp
+            eth0.save
 
-          say("\nAfter completing the appliance configuration, please restart #{I18n.t("product.name")} server processes.")
-          press_any_key
-        end
+            say("\nAfter completing the appliance configuration, please restart #{I18n.t("product.name")} server processes.")
+            press_any_key
+          end
+        when 'static'
+          say("Static Network Configuration\n\n")
+          say("Enter the new static network configuration settings.\n\n")
 
-      when I18n.t("advanced_settings.static")
-        say("Static Network Configuration\n\n")
-        say("Enter the new static network configuration settings.\n\n")
+          new_ip   = ask_for_ipv4("IP Address", ip)
+          new_mask = ask_for_ipv4("Netmask", mask)
+          new_gw   = ask_for_ipv4("Gateway", gw)
+          new_dns1 = ask_for_ipv4("Primary DNS", dns1)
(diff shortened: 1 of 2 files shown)


Fryguy:
UX team already has plans for the entire appliance console menu, so whatever we decide, I would like to make sure we are running the same path before proceeding.  cc @serenamarie125

isimluk:
Terrific news UX wants to invest in the black console. My $0.02 are that we should just make a small plug-in to cockpit and drop the black console.

Anyway, let me just highlight that based on my UX training this change is improvement no matter the paradigm.

chessbyte:
@isimluk I like this idea of extracting Network Configuration.  Maybe we should do something similar with Database Configuration.

Fryguy:
FYI, we will need downstream changes for this as well

serenamarie125:
This (submenus for related options) is right in line with what we were thinking!  Does Set Hostname really fit in here though @isimluk @Fryguy ?

isimluk:
It seems natural to me to include `Set Hostname` in the networking. But don't know what other thing?

lib/gems/pending/appliance_console.rb
```diff
@@ -150,37 +150,44 @@ module ApplianceConsole
       clear_screen
       selection = ask_with_menu("Advanced Setting", AS_OPTIONS, nil, true)
       case selection
-      when I18n.t("advanced_settings.dhcp")
```
Fryguy: I think we've lost the ability to i18n these lines.  I'm ok with this for now, but that should probably be put back.

lib/gems/pending/appliance_console/locales/appliance/en.yml
```diff
@@ -25,10 +22,7 @@ en:
     - shutdown
     - summary
     - quit
-    dhcp: Set DHCP Network Configuration
-    static: Set Static Network Configuration
-    testnet: Test Network Configuration
-    hostname: Set Hostname
+    networking: Set-up Networking
```
Fryguy: `Set-up` is not proper English.  This should be either `Setup` or `Set up` or just `Set`

Fryguy:
@isimluk Found a few minor things.

@gtanzillo Up to you if you want to merge now and cleanup later or just cleanup now.

carbonin:
I believe this will also require some changes to productization right @Fryguy @gtanzillo ?

lib/gems/pending/appliance_console/locales/appliance/en.yml
jrafanie: Maybe `Configure` ?

lib/gems/pending/appliance_console.rb
```diff
@@ -150,37 +150,44 @@ module ApplianceConsole
...
-          resolv = LinuxAdmin::Dns.new
-          resolv.search_order = []
-          resolv.nameservers = []
-          resolv.save
+      when I18n.t('advanced_settings.networking')
+        options = {
+          'Set DHCP Network Configuration'   => 'dhcp',
+          'Set Static Network Configuration' => 'static',
+          'Test Network Configuration'       => 'testnet',
+          'Set Hostname'                     => 'hostname'
```
jrafanie: These need to come from the locale file... I think that's why @Fryguy is saying

lib/gems/pending/appliance_console/locales/appliance/en.yml
Fryguy: Oh I like `Configure` much better.

Fryguy:
@carbonin Yes, it will require changes to productization.  @isimluk Be sure to make an MR to productization once this is merged.

jrafanie:
I like this change.  I agree with @carbonin, we should make sure productization is done.  My other comments are above.  Nice work @isimluk 👍

lib/gems/pending/appliance_console.rb
isimluk: When working on this, 13 days ago, I contacted our i18n guy and he was ok with dropping few sentences from the yaml given the overall state of the localization of the appliance_console.

Majority (!) of the stuff in the appliance_console is not localized. Even the stuff that is localized is not translated.

We can work on the localization as a separate issue. Perhaps we could start after the plans from the UX team materialize. That way, we will be sure to translate final strings.

PR #69, opened 2017-02-22
Drop this regexp and never look back

isimluk:


isimluk:
It seems the CI is green if I click on it. No idea why it seems to be yellow on this page.

carbonin:
That's odd. I'll close/reopen once to see if it registers.

isimluk:
Thanks!!

PR #70, opened 2017-02-22
Allow users to input ipv6 where it makes sense

isimluk:
## What
In many prompts we allow users to put IP address. In some prompts it makes sense to only accept IPv4 addresses (when setting up static IPv4). In other cases it makes sense to accept IPv4 or IPv6.

This PR makes `Test Network Configuration` to magically work with IPv6.

I'll continue to work with other dialogs to make sure, we are able to process IPv6. However, we can allow IPv6 inputs to be valid now.

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -171,11 +171,11 @@ To modify the configuration, use a web browser to access the management page.
         say("Static Network Configuration\n\n")
         say("Enter the new static network configuration settings.\n\n")
 
-        new_ip   = ask_for_ip("IP Address", ip)
-        new_mask = ask_for_ip("Netmask", mask)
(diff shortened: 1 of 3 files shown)


carbonin:
It feels like this will probably come close to solving https://bugzilla.redhat.com/show_bug.cgi?id=1212896

Do you want to take that @isimluk?

isimluk:
Sure, taking, working on it anyway. It will get fixed in one of the following prs.

isimluk:
We will need new prompt for static ipv6 as things are different and people may want to have both.

isimluk:
Getting this into the build early will make my job easier, when I will try to set-up multi region, multi-zone over ipv6. Thanks!

isimluk:
Thanks!!

Issue #33, opened 2016-12-21
Prompt for and configure a disk when setting up a standby database


carbonin:
We should allow users to configure an attached disk when configuring a standby database. Right now we only do this when configuring a primary.

The result is we always configure the database on the root partition which isn't always ideal.

PR #71, opened 2017-02-22
Extends standby database gem to use an external disk if available.

lmiccini:
Resolves https://github.com/ManageIQ/manageiq-gems-pending/issues/33

The nightly manageiq build isn't exactly cooperating when trying to set up a DB replica, so there might be issues as I wasn't able to test properly.

I have attached a similar patch but for 5.7.0.17 (where database_replication_standby.rb is quite different from upstream) to https://bugzilla.redhat.com/show_bug.cgi?id=1425327 .

Cheers
Luca

### DIFF
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -78,14 +78,14 @@ module ApplianceConsole
       data_dir_empty? &&
         initialize_postgresql_disk if disk
         PostgresAdmin.prep_data_directory &&
-        relabel_postgresql_dir &&
-        generate_cluster_name &&
-        create_config_file(standby_host) &&
-        clone_standby_server &&
-        start_postgres &&
-        register_standby_server &&
-        write_pgpass_file &&
-        (run_repmgrd_configuration ? start_repmgrd : true)
+          relabel_postgresql_dir &&
+          generate_cluster_name &&
+          create_config_file(standby_host) &&
+          clone_standby_server &&
+          start_postgres &&
+          register_standby_server &&
+          write_pgpass_file &&
+          (run_repmgrd_configuration ? start_repmgrd : true)
     end
 
     def data_dir_empty?
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -93,10 +93,6 @@ module ApplianceConsole
       false
     end
 
-    def relabel_postgresql_dir
-      AwesomeSpawn.run!("/sbin/restorecon -R -v #{mount_point}")
-    end
-
     def clone_standby_server
       params = { :h  => primary_host,
                  :U  => database_user,
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -38,7 +38,7 @@ module ApplianceConsole
     end
 
     def choose_disk
-       @disk = ask_for_disk("database disk")
+      @disk = ask_for_disk("database disk")
     end
 
     def initialize_postgresql_disk
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -71,15 +71,15 @@ module ApplianceConsole
     def activate
       say("Configuring Replication Standby Server...")
       initialize_postgresql_disk if disk
-      PostgresAdmin.prep_data_directory &&
-        data_dir_empty?
+      PostgresAdmin.prep_data_directory
+      data_dir_empty? &&
         generate_cluster_name &&
-          create_config_file(standby_host) &&
-          clone_standby_server &&
-          start_postgres &&
-          register_standby_server &&
-          write_pgpass_file &&
-          (run_repmgrd_configuration ? start_repmgrd : true)
+        create_config_file(standby_host) &&
+        clone_standby_server &&
+        start_postgres &&
+        register_standby_server &&
+        write_pgpass_file &&
+        (run_repmgrd_configuration ? start_repmgrd : true)
     end
 
     def data_dir_empty?
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -72,14 +72,14 @@ module ApplianceConsole
       say("Configuring Replication Standby Server...")
       initialize_postgresql_disk if disk
       PostgresAdmin.prep_data_directory &&
-        data_dir_empty? &&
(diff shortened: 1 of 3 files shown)


lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -35,6 +37,25 @@ def ask_questions
...
+                                    :mount_point         => mount_point,
+                                    :name                => "pg",
+                                    :volume_group_name   => PostgresAdmin.volume_group_name,
+                                    :filesystem_type     => PostgresAdmin.database_disk_filesystem,
+                                    :logical_volume_path => PostgresAdmin.logical_volume_path).setup
+      end
+    end
+
+    def mount_point
+      Pathname.new(ENV.fetch("APPLIANCE_PG_MOUNT_POINT"))
```
carbonin: I think this belongs in `PostgresAdmin`

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -35,6 +37,25 @@ def ask_questions
       confirm
     end
 
+    def choose_disk
+      @disk = ask_for_disk("database disk")
```
carbonin: I don't think this really needs a method. You can just do this in `#ask_questions`.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -75,6 +99,10 @@ def data_dir_empty?
       false
     end
 
+    def relabel_postgresql_dir
```
carbonin: Do we need this here? The data directory should be empty before we do the clone.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -55,13 +76,16 @@ def ask_for_repmgrd_configuration
...
       data_dir_empty? &&
-        generate_cluster_name &&
-        create_config_file(standby_host) &&
-        clone_standby_server &&
-        start_postgres &&
-        register_standby_server &&
-        write_pgpass_file &&
-        (run_repmgrd_configuration ? start_repmgrd : true)
+        initialize_postgresql_disk if disk
+        PostgresAdmin.prep_data_directory &&
```
carbonin: I think the check for the empty data directory should go after these because the clone will fail if it isn't empty.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -35,6 +37,25 @@ def ask_questions
       confirm
     end
 
+    def choose_disk
+      @disk = ask_for_disk("database disk")
+    end
+
+    def initialize_postgresql_disk
```
carbonin: I would like to move this somewhere else, but I don't know where it would fit ... Any ideas @jvlcek ?

jvlcek: Perhaps part of PostgresAdmin in postgres_admin.rb

carbonin: I was thinking that too, but `LogicalVolumeManagement` is internal to appliance console. It would be odd for `PostgresAdmin` to require into the appliance_console files.

jvlcek: Ah yes. Well then, since refactoring `LogicalVolumeManagement` seems to be beyond the scope of this PR, perhaps leaving `initialize_postgresql_disk` here but making it private?

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -35,9 +37,21 @@ def ask_questions
...
+                                    :volume_group_name   => PostgresAdmin.volume_group_name,
+                                    :filesystem_type     => PostgresAdmin.database_disk_filesystem,
+                                    :logical_volume_path => PostgresAdmin.logical_volume_path).setup
+      end
+    end
+
     def confirm
       super
       say(<<-EOS)
+        Database Disk:              #{disk}
```
carbonin: This should be conditional on if the disk was set.

lmiccini: right :)

lib/gems/pending/appliance_console/database_replication_standby.rb
carbonin: It looks like you removed the caller, but not this method definition. I think this can go too.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -54,7 +70,9 @@ def ask_for_repmgrd_configuration
 
     def activate
       say("Configuring Replication Standby Server...")
-      data_dir_empty? &&
+      initialize_postgresql_disk if disk
+      PostgresAdmin.prep_data_directory &&
```
carbonin: This doesn't return a value that makes sense with this `&&` chaining. Maybe this should be part of `initialize_postgresql_disk`?

lmiccini: agree. how about we just remove the "&&"? if the dir is empty the next action will kick in anyhow, if not it should break.

carbonin: That works too :+1:

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -54,14 +70,16 @@ def ask_for_repmgrd_configuration
...
+        data_dir_empty?
```
carbonin: No, this is wrong. If the directory is not empty, we want this to fail before doing any of this other work.

I was advocating removing the `&&` from after `PostgresAdmin.prep_data_directory` because it doesn't conform to the return value we would expect.

Then if `PostgresAdmin.prep_data_directory` fails we will just fail on `data_dir_empty?` and not execute the rest.

lmiccini: Not sure I understand fully how you are thinking about this. How about:
1. initialize_postgresql_disk if disk (initialize the disk if the user selected one)
2. PostgresAdmin.prep_data_directory (prepare data dir. here we remove '&&' )
3. check data_dir_empty?

or should it be the other way around, so check data_dir_empty first and then move to the next?
I am afraid in this case we might also fail if a user tried to configure the internal database before (so we have a valid pg dir, despite the content being disposable).

what do you think?

carbonin: Then we will raise if `initialize_postgresql_disk` fails so we will not attempt to execute the later tasks and `PostgresAdmin.prep_data_directory` is not expected to fail as most of the commands are `rm -rf`.

lmiccini:
@carbonin pong. will try to work them out today.

carbonin:
Closing this in favor of #96

PR #74, opened 2017-02-23
Force run umount if you forced run mount

isimluk:
If we run sudo for mount, we need to do the same for umount. The comment already mentioned that.

### DIFF
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -34,7 +34,7 @@ class MiqGenericMountSession
     # If sudo is required, ensure you have /etc/sudoers.d/miq
     # Cmnd_Alias MOUNTALL = /bin/mount, /bin/umount
     # %wheel ALL = NOPASSWD: MOUNTALL
-    if rv.include?("mount: only root can do that") || rv.include?('umount failed: Operation not permitted')
+    if $CHILD_STATUS.exitstatus == 1 && cmd_str =~ /^(mount|umount) /
       rv = `sudo #{cmd_str} 2>&1`
     end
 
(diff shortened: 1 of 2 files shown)


lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -34,7 +34,9 @@ def self.runcmd(cmd_str)
     # If sudo is required, ensure you have /etc/sudoers.d/miq
     # Cmnd_Alias MOUNTALL = /bin/mount, /bin/umount
     # %wheel ALL = NOPASSWD: MOUNTALL
-    rv = `sudo #{cmd_str} 2>&1` if rv.include?("mount: only root can do that")
+    if rv.include?("mount: only root can do that") || rv.include?('umount failed: Operation not permitted')
```
carbonin: Can we make this less dependent on the error message, that seems fragile to me.

Maybe this is more exact?

```
if cmd_str =~ /^(mount|umount)/ && $CHILD_STATUS.exit == 1
```

isimluk: Fixed.

Thanks for review!!

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -34,7 +34,9 @@ def self.runcmd(cmd_str)
     # If sudo is required, ensure you have /etc/sudoers.d/miq
     # Cmnd_Alias MOUNTALL = /bin/mount, /bin/umount
     # %wheel ALL = NOPASSWD: MOUNTALL
-    rv = `sudo #{cmd_str} 2>&1` if rv.include?("mount: only root can do that")
+    if $CHILD_STATUS.exitstatus && cmd_str =~ /^(mount|umount) /
```
carbonin: This is still not correct, in ruby all integers are truthy.

I would say this should use either `!$CHILD_STATUS.success?` (which returns a `true` or `false` value) or check explicitly for exit code 1 which is defined as `incorrect invocation or permissions` in the `mount (8)` man pages.

PR #75, opened 2017-02-23
Require ovirt 0.15.1 to fix os type issue

agrare:
Ovirt 0.15.1 fixed an issue where the template os type was overwritten.  Fixed https://github.com/ManageIQ/ovirt/pull/78


agrare:
cc @bdunne @Fryguy

PR #77, opened 2017-02-24
Restrict minor, but allow betas

chrisarcand:
On second thought, this is better

This still restricts the minor version as Rails doesn't actually follow
semver and no changes between minor versions is NOT guaranteed - yet
still allows for beta versions of the next minor to be used.

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -20,9 +20,9 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "> 2.2.2"
 
-  s.add_runtime_dependency "actionpack",              "~> 5.0"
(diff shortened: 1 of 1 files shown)


mfeifer:
@miq-bot set_milestone Sprint 55 Ending Feb 27, 2017

mfeifer:
@miq-bot set_milestone "Sprint 55 Ending Feb 27, 2017"

mfeifer:
@miq-bot set_milestone Sprint 55

chrisarcand:
@mfeifer You can't add the milestone because you're trying to add a closed one (only the current sprint is open). This could probably be considered a bug in the bot because doing what you're doing now is a valid use case :) I'll fix the bot accordingly.

PR #78, opened 2017-02-25
Remove ezcrypto dependency

Fryguy:
This PR removes EzCrypto gem and instead uses the existing CryptString class, but morphs in into a more general MiqPassword::Key class.

@chrisarcand @kbrock Please review.  **NOTE** There is a change needed on ManageIQ main repo before merging this, as fix_auth creates a CryptString directly, and will have to change to create an MiqPassword::Key instead.

cc @jrafanie (I pinged @chrisarcand because I thought this was for Rails 5.1, but now realize it's actually for Ruby 2.4)


lib/gems/pending/encryption/CryptString.rb
```diff
@@ -1,34 +0,0 @@
-require 'openssl'
-
-class CryptString
```
kbrock: Yay!

nice to get this out of here.

Can we drop this functionality all together instead of merging into `MiqPassword::Key`?

lib/gems/pending/util/miq-password.rb
```diff
@@ -185,16 +186,79 @@ def self.ez_load(filename, recent = true)
     if !File.exist?(filename)
       nil
     elsif recent
-      EzCrypto::Key.load(filename)
+      params = YAML.load_file(filename)
+      Key.new(*params.values_at(:algorithm, :key, :iv))
```
kbrock: may be nice to have `Key.load` and `Key#save` for use from rails console

lib/gems/pending/util/miq-password.rb
```diff
@@ -185,16 +186,79 @@ def self.ez_load(filename, recent = true)
...
+      @iv        = iv
+      @raw_iv    = iv && Base64.decode64(iv)
+    end
+
+    def encrypt(str)
+      apply(:encrypt, str)
+    end
+
+    def encrypt64(str)
+      Base64.encode64(encrypt(str)).chomp
```
kbrock: would using `strict_encode64` work for us here?

Think there may have been a reason I avoided in the past, but so much has changed since then...

lib/gems/pending/util/miq-password.rb
```diff
@@ -185,16 +186,79 @@ def self.ez_load(filename, recent = true)
...
+    def self.generate_key(password = nil, salt = nil)
+      password ||= OpenSSL::Random.random_bytes(GENERATED_KEY_SIZE)
+      Base64.encode64(Digest::SHA256.digest("#{password}#{salt}")[0, GENERATED_KEY_SIZE]).chomp
+    end
+
+    def initialize(algorithm = nil, key = nil, iv = nil)
+      @algorithm = algorithm || "aes-256-cbc"
+      @key       = key || generate_key
+      @raw_key   = Base64.decode64(@key)
+      @iv        = iv
```
kbrock: do we need both `iv` and `raw_iv`?

kbrock:
wonder if we can remove `v1` from fix auth and that way we remove reference to `CryptString`

lib/gems/pending/util/miq-password.rb
Fryguy: I was thinking about that...probably a followup PR.

lib/gems/pending/util/miq-password.rb
Fryguy: Although we don't have code that uses it at the moment, it felt better to keep it in the API for consistency.  We can always remove in a followup PR, I think.

lib/gems/pending/util/miq-password.rb
kbrock: well, this right here is that method ;)

lib/gems/pending/util/miq-password.rb
Fryguy: I didn't know about that method...I'll check on it.

Fryguy:
> wonder if we can remove v1 from fix auth and that way we remove reference to CryptString

v0 is the only problematic key because it is not Base64 encoded by default in the key file and also because the final output is not surrounded by the `v#{}` wrapper.  If we could get rid of v0, then the v1 and v2 code uses the same code paths, and a ton of code can go away.  In this PR, both v1 and v2 now use CryptString (renamed to MiqPassword::Key).  The only two lines we really used from ezcrypto were [here](https://github.com/ManageIQ/manageiq-gems-pending/pull/78/files#diff-284158ccdc0e5bd02f45f4f4298ea594R206) for key generation, and CryptString did the rest already [here](https://github.com/ManageIQ/manageiq-gems-pending/pull/78/files#diff-284158ccdc0e5bd02f45f4f4298ea594R254) for actually applying the cipher.

Fryguy:
LOL I wrote all that and then realized you were talking about *fix_auth*.  Yeah, I'm not sure how we can remove v1 from fix_auth unless we can be sure a user is *past* that point.

chrisarcand:
Nice work doing this, wow. I thought this would be a much more painful for 2.4.

kbrock:
@Fryguy What is v0 key?

In Feb 2008, @gtanzillo checked in miq-password.rb and we were already using v1 keys.

I have been called with many encryption questions and have yet to see any was official password support for the open source transition (5.2 -> 5.3) and no one in support was able to show me something encrypted with a v0 key.

----

command to look at old manageiq / miq-password history (target dc7ed35c)

```bash
cd manageiq
git log --follow -w -p 16d47119d7 -- lib/util/miq-password.rb
```

jrafanie:
Wow, does this drop ezcrypto by doing it directly with openssl?  Ie, without a change in formats?  If so, wow, nice!

Fryguy:
> Wow, does this drop ezcrypto by doing it directly with openssl? Ie, without a change in formats? If so, wow, nice!

Yup exactly.

Fryguy:
@kbrock https://github.com/ManageIQ/manageiq-gems-pending/blob/master/spec/util/miq-password_spec.rb and search "v0".  I was the key that was in place before the v1 key.  It was hardcoded, aes-128-cbc (as opposed to 256), and unadorned (that is, it didn't have the `v#{}`wrapper around it.

kbrock:
@Fryguy yea, I'm very familiar with the technology behind the v0 key. But business wise, I can't figure out if we have let users access it within the past 9 years

jrafanie:
I'm good with this change.  @Fryguy @kbrock any further concerns before merging?  This will make our ruby 2.4 PR's output so much more sane.

PR #82, opened 2017-03-03
Use localhost shortcut in dialogs to configure internal db

isimluk:
s/127.0.0.1/localhost/ for internal db config.

Let's not bother ipv6 only admins with archaic quaternion.

## Previously:
```
Each database region number must be unique.
Enter the database region number: 1
Enter the database password on 127.0.0.1: *******
```

## Now
```
Each database region number must be unique.
Enter the database region number: 1
Enter the database password on localhost: *******
```

### DIFF
--- a/lib/gems/pending/appliance_console/internal_database_configuration.rb
+++ b/lib/gems/pending/appliance_console/internal_database_configuration.rb
@@ -24,7 +24,7 @@ module ApplianceConsole
     end
 
     def set_defaults
-      self.host              = "127.0.0.1"
+      self.host              = 'localhost'
       self.username          = "root"
       self.database          = "vmdb_production"
       self.run_as_evm_server = true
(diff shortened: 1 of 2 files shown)


isimluk:
```
Activating the configuration using the following settings...
Host:     localhost
Username: root
Database: vmdb_production
Region:   1

Initialize postgresql disk starting
Initialize postgresql disk complete
Initialize postgresql starting
Initialize postgresql complete
Checking for connections to the database...

Create region starting
Create region complete

Configuration activated successfully.

Press any key to continue.
```

Note that first / summary screen already shows `localhost` anyway.
```
ManageIQ Server:         running
ManageIQ Database:       localhost
```

The appliance works.

isimluk:
After the change, connection to this db from other appliance works as well.

isimluk:
Thanks!

PR #83, opened 2017-03-03
Remove the WriteVm functionality since it's not used anywhere

### LINKED ISSUE #80: `WriteVm/build_linux/init.rb` is broken based on paths
When this code (manageiq-gems-pending) was refactored out of the ManageIQ core repo, the path referenced [here](https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/WriteVm/build_linux/init.rb#L13) was removed, so this code does not, and can not work.

I'm not familiar enough with it's original purpose to remove it myself, but it hasn't been touched since at least 2009 ...

Fryguy:
Fixes #80 

@chessbyte Please review


chessbyte:
@roliveri I don't think we will ever be writing to VMs.  And even if we do, not sure this crufty code will be of any use.

jrafanie:
Very nice 🏆 

I agree, we can always resurrect this code in the future using the magic of git.

roliveri:
@Fryguy @chessbyte - Do we still have the prototypes repo? This should be moved there. While we may not be writing to customer's VMs, there is probably code in there that can be applied in other areas.

Fryguy:
@roliveri We have the prototypes repo downstream.  I can move this there.

roliveri:
@Fryguy That would be good, thanks.

PR #90, opened 2017-03-10
Favor Apache over Rails to display HTTP error documents

skateman:
This bring us closer to unified error pages in both UIs https://github.com/ManageIQ/manageiq/issues/14244


carbonin:
This isn't really dynamic right? Would it be better to put this in one of the static files in the appliance repo (https://github.com/ManageIQ/manageiq-appliance/tree/master/COPY/etc/httpd/conf.d)?

/cc @abellotti

skateman:
Sure thing, but I think we still need the `ProxyErrorOverride` in this PR...

abellotti:
static is fine, we needed to do this for external auth errors for Kerberos authentication (web SSO) https://github.com/ManageIQ/manageiq-appliance/blob/master/TEMPLATE/etc/httpd/conf.d/manageiq-external-auth.conf.erb

carbonin:
We are no longer rendering these files from MiqApache at runtime. This change should be made in the manageiq-appliance repo [here](https://github.com/ManageIQ/manageiq-appliance/blob/master/COPY/etc/httpd/conf.d/manageiq-redirects-ui)

PR #91, opened 2017-03-13
Remove caching of apache config files :fire:

jrafanie:
It's not clear this caching was needed.  We can add caching if we
really need it.


jrafanie:
ok @carbonin all 💚

PR #95, opened 2017-03-14
Use apachectl configtest exit code to test syntax.

jrafanie:
It's brittle to check the standard output/error for "Syntax Ok".

The exit code should be non-zero when there's a problem:

With invalid syntax:
```
[root@localhost ~]# apachectl configtest
httpd: Syntax error on line 353 of /etc/httpd/conf/httpd.conf: Syntax
error on line 2 of /etc/httpd/conf.d/manageiq-balancer-ui.conf: Expected
</Proxy2> but saw </Proxy>
[root@localhost ~]# echo $?
1
```

With valid syntax:
```
[root@localhost ~]# apachectl configtest
[Tue Mar 14 11:39:56.933429 2017] [so:warn] [pid 16656] AH01574: module
ssl_module is already loaded, skipping
AH00558: httpd: Could not reliably determine the server's fully
qualified domain name, using localhost.localdomain. Set the 'ServerName'
directive globally to suppress this message
Syntax OK
[root@localhost ~]# echo $?
0
```

### DIFF
--- a/lib/gems/pending/util/miq_apache/miq_apache.rb
+++ b/lib/gems/pending/util/miq_apache/miq_apache.rb
@@ -130,12 +130,10 @@ module MiqApache
         res = MiqUtil.runcmd('apachectl configtest')
       rescue => err
         $log.warn("MIQ(MiqApache::Control.config_ok?) Configuration syntax failed with error: #{err} for result: #{res}") if $log
-        return false
+        false
+      else
+        true
       end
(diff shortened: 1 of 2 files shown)


jrafanie:
Awaiting testing results from @fbladilo 🙇

fbladilo:
@jrafanie :+1: we now know the real cause of issues, this seems like a better approach, to check for the return value instead output.

PR #100, opened 2017-03-24
Update the armrest version to 0.7.0.

djberg96:
This PR is to keep things in sync with the Azure provider repo.

https://github.com/ManageIQ/manageiq-providers-azure/pull/42

As with that PR, the major version changes shouldn't be scary. The change from 0.5.x to 0.6.x was caused by an incompatible change to Role classes that we don't use internally. The change from 0.6.x to 0.7.x is the backwards incompatible, but necessary, change for getting templates.

Update: I had to update the memoist dependency in order to avoid a conflict.

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "activesupport",           ">= 5.0", "< 5.1"
   s.add_runtime_dependency "addressable",             "~> 2.4"
   s.add_runtime_dependency "awesome_spawn",           "~> 1.4"
-  s.add_runtime_dependency "azure-armrest",           "~> 0.5.2"
+  s.add_runtime_dependency "azure-armrest",           "~> 0.7.0"
   s.add_runtime_dependency "bcrypt",                  "~> 3.1.10"
   s.add_runtime_dependency "binary_struct",           "~> 2.1"
   s.add_runtime_dependency "bundler",                 ">= 1.8.4" # rails-assets requires bundler >= 1.8.4, see: https://rails-assets.org/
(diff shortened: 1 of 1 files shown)


djberg96:
@miq-bot add_label euwe/yes

djberg96:
@miq-bot add_label darga/yes

djberg96:
@Fryguy @jerryk55 @roliveri Any objections?

roliveri:
@djberg96 No objections from me.

djberg96:
@miq-bot add_label fine/yes

simaishi:
Fine backport details:

```
$ git log -1
commit aaee67427bce5ae3023827afe0e966bf38c7b2f0
Date:   Fri Mar 31 13:40:44 2017 -0400

    Merge pull request #100 from djberg96/armrest_version
    
    Update the armrest version to 0.7.0.
    (cherry picked from commit eeb4a3e3ad4283968d3f1f0e70642c488d527a5c)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1438092
```

djberg96:
@simaishi We still need this in euwe, correct? Looks like it's still at 0.5.2:

https://github.com/ManageIQ/manageiq/blob/euwe/gems/pending/Gemfile#L11

Or am I looking in the wrong place?

simaishi:
We do, but this has to wait.. I can only backport fixes for 'blocker' only right now, and this isn't a blocker.

djberg96:
@simaishi Will this suffice?

https://bugzilla.redhat.com/show_bug.cgi?id=1438092

It's marked for 5.8 (and there's a clone for 5.6.5), just not one for 5.7, though there should be. We really must have this updated.

@blomquisg What process do I need to follow if that's not good enough?

simaishi:
@djberg96 I had a quick chat with @blomquisg about this BZ last week and he mentioned it's a blocker for 5.8, but not a blocker for 5.7/5.6.  So this will be included in 5.7.3 and 5.6.5.  5.7.2 is accepting blocker only at this point.

Please check with @blomquisg if this should be a 5.7.2 blocker.

simaishi:
Backported to Euwe via https://github.com/ManageIQ/manageiq/pull/14799

PR #101, opened 2017-03-24
Add ruby 2.4 support

jrafanie:
~~* Copied the webmock patch from https://github.com/ManageIQ/manageiq/pull/13104~~

~~* [This code is looking for Bignum and fails on ruby 2.4](https://github.com/ManageIQ/manageiq-gems-pending/blob/8dacc4146c526493bc8f83f661bcf84dbc2220ac/lib/gems/pending/fs/ntfs/directory_index_node.rb#L55).  Bignum and Fixnum were merged to Integer in ruby 2.4.  It doesn't look like can just replace Bignum with Integer.  I tried.~~

EDIT: see https://github.com/ManageIQ/manageiq-smartstate/pull/10

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,5 +4,6 @@ language: ruby
 rvm:
   - 2.2.6
   - 2.3.3
+  - 2.4.1
 before_install: gem install bundler -v 1.12.5
 after_script: bundle exec codeclimate-test-reporter


jrafanie:
@jerryk55 Please take a look at the second bullet in the description.  We don't need to fix it right away, I can't tell how hard it is to get 2.3 and 2.4 compatibility.

djberg96:
That's because `Integer == Fixnum` and `Integer == Bignum` both return true in 2.4.

I think to maintain compatibility you're going to have to do something like this:

    def too_big(num)
      bitness = ['a'].pack('P').length > 4 ? "64" : "32"
      if bitness == '32'
        raise something if num > 2**31
      else
        raise something if num > 2**63
      end
    end

Fryguy:
@jrafanie Where are we with Ruby 2.4 support.  I have heard a lot about GC improvements in 2.4 and I would like for this to be the default for gaprindashvili.  The sooner we can get it on appliances, the better, IMO.

jrafanie:
ok, this should be good to go when tests pass.  Note smartstate code was extracted and the problematic conditional was removed in https://github.com/ManageIQ/manageiq-smartstate/pull/10

djberg96:
👍

Fryguy:
@chessbyte Please be sure to mark future extractions with Ruby 2.4.1 in the .travis.yml

PR #102, opened 2017-03-28
Bump hawkular-client gem version to 2.9.0

israel-hdez:
Version 2.9.0 of hawkular-client gem was released on March 13th. Some
deprecations were made. Bumping version to be able to stop using the
deprecated names in ManageIQ code.

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "ffi",                     "~>1.9.3"
   s.add_runtime_dependency "ffi-vix_disk_lib",        "~>1.0.2"  # used by lib/VixDiskLib
   s.add_runtime_dependency "fog-openstack",           "=0.1.20"
-  s.add_runtime_dependency "hawkular-client",         "=2.8.0"
+  s.add_runtime_dependency "hawkular-client",         "=2.9.0"
   s.add_runtime_dependency "highline",                "~> 1.6.21" # Needed for the appliance_console
   s.add_runtime_dependency "httpclient",              "~>2.7.1"
   s.add_runtime_dependency "image-inspector-client",  "~>1.0.3"


abonas:
@miq-bot add_labels providers/hawkular

abonas:
cc @moolitayer @simon3z

abonas:
@chessbyte @blomquisg are there any labels in this repo or it was decided they are not needed here?
also, I might have missed a discussion on the topic, but why is hawkular in a pending gems repo?

abonas:
@miq-bot assign @Jiri-Kremser

abonas:
@chessbyte why can't I assign this to Jiri?

abonas:
cc @Jiri-Kremser please review. are we ok with ~> instead of =  in the gem version?

yaacov:
@moolitayer thanks! , the metrics collector is not affected from this change.

Jiri-Kremser:
> @moolitayer  (BTW the change log link here is broken)

It should be ok, now. thx

Jiri-Kremser:
I'll definitely prefer the `=2.9.0` instead of `~>`

israel-hdez:
@Jiri-Kremser Are you ok if I use ’>~2.9.0’ instead to locking to an specific version? Just in case of bugfixes.

Jiri-Kremser:
@israel-hdez I suppose you meant `~>` and not `>~`. Or is there any `>~` semver operator?

We had this before and it wasn't good. What it allows is to break the MiQ build if we release the hawkular client gem with a change that we consider as harmless. It would require us to run the MiQ test suite with the new gem version to find out if there really isn't any api breakage. Untill, we have this automation, I wouldnt do that. Sometimes it's hard to figure out, because the breakage can be also in the JSON payload, URLs, headers.. not only the method signatures in ruby.

israel-hdez:
@Jiri-Kremser Oh, Yes!... I meant `~>2.9.0`. Sorry, I wrote that from my cellphone and typed symbols as I found them and, additionally, just woke up and my brain was still asleep.

Yes, I understand your point. That's because I suggested using `~>2.9.0` to avoid a minor version update and allow only patch/bugfix releases. But if even with that there have been problems, I'll lock to 2.9.0 without discussion. You say.

israel-hdez:
Locked the version to 2.9.0 to stay secure.

abonas:
@miq-bot assign @bronaghs

abonas:
@miq-bot assign @chessbyte

simaishi:
Backported to Fine as this will be indirectly backported by https://github.com/ManageIQ/manageiq-gems-pending/pull/138 anyway and I'd rather have 2 clean cherry-picks...

```
$ git log -1
commit 7aa1d127790f82c58e3a61608e99bd1517e2f829
Author: Oleg Barenboim <chessbyte@gmail.com>
Date:   Thu Mar 30 11:29:17 2017 -0400

    Merge pull request #102 from israel-hdez/bump_hawkular_client_gem_version
    
    Bump hawkular-client gem version to 2.9.0
    (cherry picked from commit 5ade0070dd24b8af8730af026ff6418a015d269d)
```

PR #103, opened 2017-03-28
Add LogDumpObject.

lfu:
Move the common code out of MiqRequestTask::Dumping.
Part of [Refactor MiqRequestTask::Dumping. #14547](https://github.com/ManageIQ/manageiq/pull/14547).

@miq-bot assign @gmcculloug 
@miq-bot add_label fine/yes, euwe/no
cc @bzwei

### DIFF
--- a/lib/gems/pending/util/log_dump_object.rb
+++ b/lib/gems/pending/util/log_dump_object.rb
@@ -1,0 +1,35 @@
+require 'active_support/concern'
+
+module LogDumpObject
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def dump_obj(obj, prefix = nil, &block)
+      meth = "dump_#{obj.class.name.underscore}".to_sym
+
+      if self.respond_to?(meth)
+        return send(meth, obj, prefix, &block)
+      end
+
+      yield obj, prefix
+    end
+
+    def dump_hash(hd, prefix, &block)
+      hd.each { |k, v| dump_obj(v, "#{prefix}[#{k.inspect}]", &block) }
+    end
+
+    def dump_array(ad, prefix, &block)
+      ad.each_with_index { |d, i| dump_obj(d, "#{prefix}[#{i}]", &block) }
+    end
+  end
+
+  def dump_obj(obj, prefix = nil, options = {})
+    self.class.dump_obj(obj, prefix) do |val, prefix|
+      value = val
+      if options.try(:protected).try(:path).to_miq_a.any? { |filter| prefix =~ filter }
+        value = "<PROTECTED>"
+      end
+      $log.info("#{prefix}(#{val.class}) = #{value.inspect}")
+    end
+  end
+end
--- a/spec/util/log_dump_object_spec.rb
+++ b/spec/util/log_dump_object_spec.rb
@@ -1,0 +1,41 @@
+require 'util/log_dump_object'
+require 'util/extensions/miq-array'
+
+describe LogDumpObject do
+  before do
+    @test_object = Class.new do
+      include LogDumpObject
+    end.new
+  end
+
+  describe '#dump_obj' do
+    it "calls .dump_obj" do
+      expect(@test_object.class).to receive(:dump_obj)
+      @test_object.dump_obj({:param_1 => 1})
+    end
+
+    it 'hides passwords' do
+      @test_object.dump_obj(
+        {:my_password => "secret"},
+        "my choices: ",
+        :protected => {:path => /[Pp]assword/}
+      ) do |obj, prefix|
+        expect(obj).to eq("secret")
+        expect(prefix).to eq("my choices: [:my_password]")
+        expect($log).to receive(:info).with(/ = <PROTECTED>/)
+      end
+    end
+  end
+
+  describe '.dump_obj' do
+    it 'accepts a hash' do
+      expect(@test_object.class).to receive(:dump_hash)
+      @test_object.class.dump_obj({:param_1 => 1}) {}
+    end
+
+    it 'accepts an array' do
+      expect(@test_object.class).to receive(:dump_array)
+      @test_object.class.dump_obj(%w(1 2 3)) {}
+    end
+  end
+end


lib/gems/pending/util/log_dump_obj.rb
```diff
@@ -0,0 +1,35 @@
+module LogDumpObj
+  extend ActiveSupport::Concern
+
+  module ClassMethods
+    def dumpObj(obj, prefix = nil, prnt_obj = STDOUT, prnt_meth = :puts, options = {})
```
chessbyte: why is `print` abbreviated to `prnt`?  I can *maybe* understand abbreviating `object` to `obj` and `method` to `meth`, but even there I am not sure what the benefit is.

lib/gems/pending/util/log_dump_obj.rb
```diff
@@ -0,0 +1,35 @@
...
+          prnt_obj.send(prnt_meth, "#{prefix}(#{obj.class}) = #{obj.inspect}")
+        end
+      end
+    end
+
+    def dumpHash(hd, prefix, prnt_obj, prnt_meth, options)
+      hd.each { |k, v| dumpObj(v, "#{prefix}[#{k.inspect}]", prnt_obj, prnt_meth, options) }
+    end
+
+    def dumpArray(ad, prefix, prnt_obj, prnt_meth, options)
```
chessbyte: why the camelCase method names instead of snake_case?

bzwei: @chessbyte I think because this is a refactoring work (see also https://github.com/ManageIQ/manageiq/pull/14547), the method names are unmodified for backward compatibility purpose. 

Maybe we can rename the method to a snake_case and make an alias for it.

bdunne:
@lfu Can you add some tests?

gmcculloug:
@lfu Nice job getting this cleaned up.  @bdunne Please review.

bdunne:
While I don't see anything particularly wrong with the PR, there are a couple of reasons I'm hesitating to merge this:
- We're trying to shrink `manageiq-gems-pending`, not add to it.
- Would this module be used in other `manageiq-gems-pending` code?
- This code always felt strange to me.

bdunne:
I think the reason it feels strange to me is because there are methods for dumping arrays, hashes, etc. that get included into a class.  I think a more natural approach would be adding a method to `Array` that would know how to write an array to a logger or stdout for display purposes.  Thoughts @gmcculloug or @Fryguy ?

bdunne:
Also, I stumbled across this https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/MiqDumpObj.rb

bdunne:
So, here are my latest thoughts...
What does this code do?
- Walk a nested set of hashes or arrays
- Mask protected hash values
- Print the thing being walked line by line
Did I miss anything?

Questions:
Is it important that the logging method (STDOUT.puts / logger.info) is called for each item in the array / hash?
Can we split the logic of walking and masking first then logging second?
Can we get to a point where we're:
- walking the nested object and masking protected fields
- just calling `pretty_print` at the top level then logging that

gmcculloug:
> Questions:
Is it important that the logging method (STDOUT.puts / logger.info) is called for each item in the array / hash?

No, at the time it felt like a cleaner output to have all the values lineup.  Admittedly the log header creates a lot of useless text.

> Can we split the logic of walking and masking first then logging second?

Yes, the only possible issue I see is dealing with deeply nested structures.

> Can we get to a point where we're:
> - walking the nested object and masking protected fields
> - just calling pretty_print at the top level then logging that

I like the idea of using an existing formatter like pretty_print for most of the structures we are dumping.  

The one exception is the logging we do for VMware structures ([VimHash and VimArray](https://github.com/ManageIQ/manageiq/blob/069ce7f35ef7cb20d6094e434c7da4f03e633603/app/models/miq_request_task/dumping.rb#L31)) to expose the `xsiType` and `vimType` which is incredible important for the structures we pass during provisioning.  

The current logging showing **xsiType** for VMware config spec:
```ruby
Config spec:  (VimHash) xsiType: <VirtualMachineConfigSpec>  vimType: <>
Config spec: ["extraConfig"] (VimArray) xsiType: <ArrayOfOptionValue>  vimType: <>
Config spec: ["deviceChange"] (VimArray) xsiType: <ArrayOfVirtualDeviceConfigSpec>  vimType: <>
Config spec: ["deviceChange"][0]["device"] (VimHash) xsiType: <VirtualE1000e>  vimType: <>
Config spec: ["deviceChange"][0]["device"]["deviceInfo"] (VimHash) xsiType: <Description>  vimType: <>
Config spec: ["deviceChange"][0]["device"]["backing"] (VimHash) xsiType: <VirtualEthernetCardNetworkBackingInfo>  vimType: <>
Config spec: ["deviceChange"][0]["device"]["backing"]["deviceName"](String) = "VM NFS Network"
```

Pretty-print logging for similar data:
```ruby
Config spec: {"annotation"=>
  "Owner: g m\nEmail: gm@rh.com\nSource: Win2k12DC-template\n\nMIQ GUID=77e286e8-18d0-11e7-83a5-6003089fa294",
 "extraConfig"=>[{"key"=>"cpuid.coresPerSocket", "value"=>"1"}],
 "deviceChange"=>
  [{"operation"=>"edit",
    "device"=>
     {"key"=>"4000",
      "deviceInfo"=>
       {"label"=>"Network adapter 1", "summary"=>"VM NFS Network"},
      "backing"=>{"deviceName"=>"VM NFS Network"},
```

The loss of data like `VirtualMachineConfigSpec`, `VirtualE1000e`, `VirtualEthernetCardNetworkBackingInfo` would be detrimental in my opinion for VMware, but I do not think this would effect other providers.

gmcculloug:
Based on @bdunne comments should we do something like this to format an object for logging (or printing) that would handle our object logging for provisioning:

```ruby
def pp_with_password_filter(object)
  ''.tap do |output|
    PP.pp(object, output).gsub!(/[Pp]assword.*=>(.*)/) { |m| m.gsub!($1, "[FILTERED]") }
    output.gsub!(MiqPassword::REGEXP) { |m| m.gsub!($2, "[FILTERED]") }
  end
end
```

The the caller would use the result to log or print the results:
```ruby
# Example usage:
p = MiqProvision.find(734)
puts "Options:\n#{pp_with_password_filter(p.options)}"
$log.info "Options:\n#{pp_with_password_filter(p.options)}"
```

The logging looks like this:
```
[----] I, [2017-04-10T12:09:00.137616 #27820:3ffe6c43f954]  INFO -- : Options:
{:use_pre_dialog=>false,
 :request_type=>"template",
 :owner_first_name=>"Greg",
 :root_password=>[FILTERED]
 :customization_template_script=>nil,
 :schedule_type=>["immediately", "Immediately on Approval"],
 :dns_servers=>nil,
 :sysprep_organization=>nil,
 :sysprep_password=>[FILTERED]
 :sysprep_server_license_mode=>["perServer", "Per server"],
...
```

cc @Fryguy

lfu:
@gmcculloug The printout used to be looked like this:
```
[----] I, [2017-04-10T12:43:13.211398 #36093:3fe220c5e1fc]  INFO -- :  (VimArray) xsiType: <ArrayOfHostInternetScsiHbaStaticTarget>  vimType: <>
[----] I, [2017-04-10T12:43:13.211503 #36093:3fe220c5e1fc]  INFO -- : [0] (VimHash) xsiType: <HostInternetScsiHbaStaticTarget>  vimType: <>
[----] I, [2017-04-10T12:43:13.211582 #36093:3fe220c5e1fc]  INFO -- : [0]["address"](String) = "10.1.1.210"
[----] I, [2017-04-10T12:43:13.211649 #36093:3fe220c5e1fc]  INFO -- : [0]["iScsiName"](String) = "iqn.1992-08.com.netapp:sn.135107242"
[----] I, [2017-04-10T12:43:13.211704 #36093:3fe220c5e1fc]  INFO -- : [1] (VimHash) xsiType: <HostInternetScsiHbaStaticTarget>  vimType: <>
[----] I, [2017-04-10T12:43:13.211753 #36093:3fe220c5e1fc]  INFO -- : [1]["address"](String) = "10.1.1.100"
[----] I, [2017-04-10T12:43:13.211819 #36093:3fe220c5e1fc]  INFO -- : [1]["iScsiName"](String) = "iqn.2008-08.com.starwindsoftware:starwindm1-starm1-test1"
```
where the value's class was printed out along with the value besides hiding the sensitive data.

In your comment above, seems the intention is only to hide out the password fields. 
Don't we need to print out the value's class?

gmcculloug:
@lfu I would suggest using this newer format for most of the objects we want to dump to the log.  There would be exceptions, mainly when dumping the VMware objects where we want to retain the `xsiType`.

PR #107, opened 2017-03-31
Prefer String#iec_60027_2_to_i from more_core_extensions

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/14601


Fryguy:
ManageIQ/manageiq#14601 is merged.

PR #110, opened 2017-04-05
Use new postgres config scheme

carbonin:
This is a complementary PR to https://github.com/ManageIQ/manageiq-appliance/pull/116

This uses the newly named `postgresql.conf` file and makes dynamic changes to the configuration using `ALTER SYSTEM` SQL commands rather than relying on reference files and ERB.

### DIFF
--- a/lib/gems/pending/appliance_console/internal_database_configuration.rb
+++ b/lib/gems/pending/appliance_console/internal_database_configuration.rb
@@ -92,7 +92,7 @@ module ApplianceConsole
     def configure_postgres
       self.ssl = File.exist?(PostgresAdmin.certificate_location.join("postgres.key"))
 
-      copy_template "postgresql.conf.erb"
+      copy_template "postgresql.conf"
       copy_template "pg_hba.conf.erb"
       copy_template "pg_ident.conf"
     end
--- a/lib/gems/pending/appliance_console/internal_database_configuration.rb
+++ b/lib/gems/pending/appliance_console/internal_database_configuration.rb
@@ -130,6 +130,11 @@ module ApplianceConsole
       block_until_postgres_accepts_connections
     end
 
+    def restart_postgres
+      LinuxAdmin::Service.new(PostgresAdmin.service_name).restart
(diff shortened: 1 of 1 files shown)


lib/gems/pending/appliance_console/internal_database_configuration.rb
```diff
@@ -126,25 +130,49 @@ def start_postgres
...
+      yield conn
+    ensure
+      conn.close
+    end
+
+    def apply_initial_configuration
+      shared_buffers = run_as_evm_server ? SHARED_DB_SHARED_BUFFERS : DEDICATED_DB_SHARED_BUFFERS
+      with_pg_connection do |conn|
+        conn.exec("ALTER SYSTEM SET ssl TO on") if ssl
+        conn.exec("ALTER SYSTEM SET shared_buffers TO #{shared_buffers}")
```
gtanzillo: If we're setting shared_buffers this this way, with `ALTER SYSTEM`, is it still necessary to have it in `01_miq_overrides.conf`?

Also, are these the only ones we want to set this way or are you planning to add more later?

carbonin: As of now this is just a replacement for the old reference conf files (`EVM_V5_Dedicated_Appliance_postgresql.conf`/`EVM_V5_Shared_Appliance_postgresql.conf`).

The only difference between these and the `postgresql.conf` that was used by default was the shared_buffers setting. So now that we know if we are a dedicated database appliance or not we can just apply this directly.

carbonin: Also, yeah, we could remove it from the `conf` file.

carbonin: Done in https://github.com/ManageIQ/manageiq-appliance/pull/116/commits/77764f0b043205b4839732e2a3b0a05f137f3e56

carbonin:
Successfully tested this out on an appliance in the shared and dedicated cases.

Removing WIP.

PR #117, opened 2017-04-11
Upgrade nokogiri to 1.7.x

djberg96:
This PR upgrades the Nokogiri dependency from 1.6.x to 1.7.x. This addresses complaints from Hakiri which indicated that there have been security patches to nokogiri/libxml2.

See CVE-2016-4658 for details:

http://cve.mitre.org/cgi-bin/cvename.cgi?name=cve-2016-4658

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -48,7 +48,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "more_core_extensions",    "~>3.2"
   s.add_runtime_dependency "net-scp",                 "~>1.2.1"
   s.add_runtime_dependency "net-sftp",                "~>2.1.2"
-  s.add_runtime_dependency "nokogiri",                "~>1.6.8"
+  s.add_runtime_dependency "nokogiri",                "~>1.7.1"
   s.add_runtime_dependency "openscap",                "~>0.4.3"
   s.add_runtime_dependency "ovirt",                   "~>0.15.1"
   s.add_runtime_dependency "parallel",                "~>1.9" # For OvirtInventory


djberg96:
The failures are caused by azure-armrest and this repo not being in sync. I'm not sure if rubywbem matters.

djberg96:
@Fryguy or @jrafanie Please merge.

jrafanie:
@isimluk @carbonin Please also review...

manageiq-gems-pending.gemspec
```diff
@@ -48,7 +48,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "more_core_extensions",    "~>3.2"
   s.add_runtime_dependency "net-scp",                 "~>1.2.1"
   s.add_runtime_dependency "net-sftp",                "~>2.1.2"
-  s.add_runtime_dependency "nokogiri",                "~>1.6.8"
+  s.add_runtime_dependency "nokogiri",                "~>1.7.1"
```
carbonin: Can we make this less restrictive?

Maybe `"~>1.7", ">=1.7.1"`?

carbonin: Then when we have this issue with version 1.8 we won't have to do the same dance between all of these gems.

djberg96: @carbonin I'm just following the existing pattern. If @jrafanie is ok with changing it, I'm ok with it.

carbonin:
Also, if we merge this as is, ManageIQ will fail to bundle because of the restriction in the azure provider [here](https://github.com/ManageIQ/manageiq-providers-azure/blob/master/manageiq-providers-azure.gemspec#L16)

carbonin:
This looks fine as it is the same restriction we have in [azure-armrest](https://github.com/ManageIQ/azure-armrest/pull/272/files), making it any less restrictive here won't make a difference in the end.

isimluk:
Updating gem dependencies in all the subprojects is like

gems

perhaps we need a cool new framework to do the job?

simaishi:
Backported to Fine to avoid conflict backporting https://github.com/ManageIQ/manageiq-gems-pending/pull/204.

Fine backport details:

```
$ git log -1
commit e266fdbe1af1eabf7f01fb49173d98aed1177c35
Date:   Wed Apr 12 16:10:09 2017 -0400

    Merge pull request #117 from djberg96/gemspec
    
    Upgrade nokogiri to 1.7.x
    (cherry picked from commit 5ead8e64e2a8d88cd1e267a00375c5eeb0ef570b)
```

PR #118, opened 2017-04-11
Remove oVirt provider code

jhernand:
Currently the part of the oVirt provider that performs event monitoring
and inventory collection (using version 3 of the oVirt API and the
'ovirt' gem) lives in this repository, in the
'lib/gems/pending/ovirt_provider' directory. As the oVirt provider has
been extracted to a separate repository, it is convenient to move that
code there. The relevant copy has been added to that repository in this
pull request:

  Move code from manageiq-gems-pending
  https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9

This patch removes the code from this repository.

Note that part of the code can't be moved, or rather, isn't convinient
to move it now. The reason is that code, the 'rhevmTest.rb' file, for
example, doesn't conceptually belong into the oVirt provider. To keep
this code in this repository it is necessary to add a dependency, so
that 'manageiq-gems-pending' will now depend on
'manageiq-providers-ovirt'.

https://github.com/ManageIQ/manageiq-providers-ovirt/issues/8


jhernand:
@durandom @Fryguy please review.

Note that I am not certain about what to do with the `rhevmTest.rb` file, and in general with the oVirt-related files in `lib/gems/pending/MiqVm/test`, as I don't understand what is the purpose of those files. Are they to be considered part of the oVirt provider? Should they be moved to the new repository? If so, how can I test them to make sure they work after moving?

Adding the oVirt provider as a dependency of `manageiq-gems-pending` is really ugly, but I think it is the correct thing if these files are to be kept in this repository.

jhernand:
Note that this shouldn't be merged before ManageIQ/manageiq-gems-pending#118.

durandom:
> Note that I am not certain about what to do with the rhevmTest.rb file, and in general with the oVirt-related files in lib/gems/pending/MiqVm/test, as I don't understand what is the purpose of those files

@bdunne Are those tests that are actually run? I'm also not sure what to do with them.
I guess the cleanest way would be to move all rhv related stuff over to the ovirt repo

bdunne:
@durandom I wouldn't move that.  It's related to fleecing code and needs to adapt to pluggable providers.  Although, updating the `requires` and any references to constants would be nice.

Gemfile
```diff
@@ -7,6 +7,10 @@ gemspec
 gem "handsoap", "~>0.2.5", :require => false, :git => "https://github.com/ManageIQ/handsoap.git", :tag => "v0.2.5-5"
 gem "rubywbem",            :require => false, :git => "https://github.com/ManageIQ/rubywbem.git", :branch => "rubywbem_0_1_0"
 
+# This is needed because some of the smart analysis tests scripts, insde the 'lib/gems/pending/MiqVm/test' directory,
+# use the oVirt provider inventory classes.
+gem "manageiq-providers-ovirt", :require => false, :git => "https://github.com/ManageIQ/manageiq-providers-ovirt.git", :branch => "master"
```
Fryguy: Once this is merged, we'll need to open GH issue in manageiq-releasen repo to update the branch name when we cut gaprindashvili.

manageiq-gems-pending.gemspec
```diff
@@ -51,7 +51,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "nokogiri",                "~>1.6.8"
   s.add_runtime_dependency "openscap",                "~>0.4.3"
   s.add_runtime_dependency "ovirt",                   "~>0.15.1"
-  s.add_runtime_dependency "parallel",                "~>1.9" # For OvirtInventory
```
Fryguy: @bdunne Even though this is being removed and the comment says "For OvirtInventory", can you please verify this is needed by anyone else that was getting it automatically?  azure comes to mind, and also the core repo, but I'm not sure what else.

bdunne:
@jhernand With https://github.com/ManageIQ/manageiq-providers-ovirt/pull/9 merged, are you ready to remove the `[WIP]`?

From what I see, the only other references to `parallel` is marked as TODO.  It isn't actually required anywhere.

bdunne:
:) I marked as WIP, unmarking and merging.

PR #123, opened 2017-04-17
Add a webtreemap view for require_with_logging

Fryguy:
Additionally, add support for metrics based on allocations and memory.

This tool uses webtreemap, which can be obtained from https://github.com/evmar/webtreemap.  See the [demo](http://evmar.github.io/webtreemap/) (NOTE: This PR is old and is using v1, so would have to be upgraded to use v2)

After using require_with_logging to obtain a log file, then requires_log_to_webtreemap.rb can be used to parse that file into the webtreemap.  Here is an example of what webtreemap looks like against current ManageIQ, at boot, profiling allocations:

manageiq - master allocations by require 2017-04-17 18-11-25

@jrafanie Please review.

### DIFF
--- a/lib/gems/pending/util/require_with_logging.rb
+++ b/lib/gems/pending/util/require_with_logging.rb
@@ -3,6 +3,7 @@ puts "** Loading Require With Logging"
 $req_log_path ||= File.expand_path('.')
 $req_log_file ||= ENV["REQUIRE_LOG"] unless ENV["REQUIRE_LOG"].nil? || ENV["REQUIRE_LOG"] == "true"
 $req_log_file ||= "requires_#{Time.now.utc.strftime("%Y%m%d%H%M%S")}.log"
+$req_log_meth ||= :log_require_time
 
 $req_log ||= File.open(File.join($req_log_path, $req_log_file), "w")
 # $req_log = $stdout
@@ -30,16 +31,28 @@ module Kernel
     $req_depth += 1 if mode == :enter || mode == :reenter
   end
 
+  def log_require_time
+    Time.now
+  end
+
+  def log_require_allocations
+    GC.stat[:total_allocated_objects]
+  end
+
+  def log_require_mb
+    Sys::ProcTable.ps(Process.pid).smaps.pss / 1_000_000.0
+  end
+
   def with_require_logging(path, mode = :enter)
     log_require(path, mode)
-    t = Time.now
+    before = send($req_log_meth)
     begin
       ret = yield
     rescue Exception
-      log_require(path, :fail, Time.now - t) rescue nil
+      log_require(path, :fail, send($req_log_meth) - before) rescue nil
       raise
     end
-    log_require(path, ret, Time.now - t)
+    log_require(path, ret, send($req_log_meth) - before)
     ret
   end
 
--- a/tools/requires_log_to_webtreemap.rb
+++ b/tools/requires_log_to_webtreemap.rb
@@ -1,0 +1,49 @@
+if ARGV.length != 2
+  puts "Usage: #{$0} /path/to/requires.log /path/to/webtreemap"
+  puts
+  puts "This tool uses webtreemap, which can be obtained from https://github.com/evmar/webtreemap"
+  exit 1
+end
+
+require 'active_support/all'
+
+requires_file, webtreemap_dir = ARGV
+
+def parse_line(line)
+  match = line.match(/^(?<indent>[0-9 ]{3})  ..  (?:\| )*(?<location>[^(]+)(?: \((?<metric>[-0-9.]+)\))?$/)
+  return {} unless match
+  {
+    :indent   => match[:indent].to_i,
+    :location => match[:location].strip,
+  }.tap do |h|
+    h[:metric] = match[:metric].to_f if match[:metric]
+  end
+end
+
+nestings = []
+current_indent = -1
+
+File.open(requires_file) do |f|
+  f.each_line do |line|
+    line = parse_line(line)
+
+    if line[:indent] > current_indent
+      current_indent += 1
+      nestings << []
+    elsif line[:indent] < current_indent
+      current_indent -= 1
+      children = nestings.pop
+      dot_metrics = line[:metric] - children.map { |c| c[:data]["$area"] }.sum
+      children << {:name => ". (#{dot_metrics})", :data => {"$area" => dot_metrics}}
+      nestings.last << {:name => "#{line[:location]} (#{line[:metric]})", :data => {"$area" => line[:metric]}, :children => children}
+    elsif line[:metric]
+      nestings.last << {:name => "#{line[:location]} (#{line[:metric]})", :data => {"$area" => line[:metric]}}
+    end
+  end
+end
+
+children = nestings.pop
+result = {:name => "Total", :data => {"$area" => children.map { |c| c[:data]["$area"] }.sum}, :children => children}
+
+webtreemap_file = File.join(webtreemap_dir, "demo/demo.json")
+File.write(webtreemap_file, "var kTree=#{result.to_json}")


Fryguy:
cc @dmetzger57 @NickLaMuro @kbrock

lib/gems/pending/util/require_with_logging.rb
```diff
@@ -30,16 +31,28 @@ def log_require(path, mode, timing = nil)
...
+  def log_require_time
+    Time.now
+  end
+
+  def log_require_allocations
+    GC.stat[:total_allocated_objects]
+  end
+
+  def log_require_mb
+    Sys::ProcTable.ps(Process.pid).smaps.pss / 1_000_000.0
```
jrafanie: Note, this is linux only.  😢

tools/requires_log_to_webtreemap.rb
```diff
@@ -0,0 +1,49 @@
...
+      nestings.last << {:name => "#{line[:location]} (#{line[:metric]})", :data => {"$area" => line[:metric]}}
+    end
+  end
+end
+
+children = nestings.pop
+result = {:name => "Total", :data => {"$area" => children.map { |c| c[:data]["$area"] }.sum}, :children => children}
+
+webtreemap_file = File.join(webtreemap_dir, "demo/demo.json")
+File.write(webtreemap_file, "var kTree=#{result.to_json}")
```
jrafanie: Do we want to include a demo html that reads in this json file and presents it, [similar to how the demo does](https://github.com/evmar/webtreemap/tree/gh-pages/demo)?

tools/requires_log_to_webtreemap.rb
```diff
@@ -0,0 +1,49 @@
+if ARGV.length != 2
+  puts "Usage: #{$0} /path/to/requires.log /path/to/webtreemap"
+  puts
+  puts "This tool uses webtreemap, which can be obtained from https://github.com/evmar/webtreemap"
+  exit 1
```
jrafanie: ugh, this exit is causing the coverage helper to fail because you can't require this file.

NickLaMuro: @jrafanie There are no tests around the "tools" dir , so does it make sense to remove that from the coverage report?  Looks like this is the first file added to that.

jrafanie:
yes, we can add an exception or just make it require'able.  Not sure if the
latter is worth it.

lib/gems/pending/util/require_with_logging.rb
Fryguy: Should I wrap that in a darwin check?

tools/requires_log_to_webtreemap.rb
Fryguy: Yeah, I think tools dir should just not be included.  What do we do in manageiq proper?

tools/requires_log_to_webtreemap.rb
Fryguy: This currently assumes you have cloned the webtreemap repo and it will just write into it.  Not the greatest approach but I figure we can iterate on it later if we want.  Your approach might be the best one.  I mostly made this PR to get the code off my machine and out into the world. :)

NickLaMuro: > I mostly made this PR to get the code off my machine and out into the world. :)

Yeah... it has been at least 3 weeks since you showed this to me and Dennis...

lib/gems/pending/util/require_with_logging.rb
jrafanie: yeah, sys proctable doesn't guarantee each platform will have the same attributes/methods on the object returned from `ps`, so the OSX and other non-linux OS's will not have smaps.

Maybe, we should raise on non-Linux platforms.

tools/requires_log_to_webtreemap.rb
jrafanie: https://github.com/ManageIQ/manageiq/blob/master/spec/coverage_helper.rb#L6

Maybe, add the tools to this list:  https://github.com/ManageIQ/manageiq-gems-pending/blob/master/spec/coverage_helper.rb#L4

Fryguy: Consider require_with_logging is already in that list, yeah, that seems right.

kbrock: `if $0 == __FILE__` is your friend?

Fryguy:
I still would like this to be merged, so reopening.  We may want to discuss alternate ways to do diagnostic type stuff since the gems/pending gem is getting smaller and smaller.  Perhaps the https://github.com/ManageIQ/manageiq-performance repo is a better fit?  @NickLaMuro Thoughts?

Fryguy:
Updated OP with "(NOTE: This PR is old and is using v1, so would have to be upgraded to use v2)"

lib/gems/pending/util/require_with_logging.rb
NickLaMuro: The `Sys::ProcTable.ps` does a [directory scan](https://github.com/djberg96/sys-proctable/blob/master/lib/linux/sys/proctable.rb#L114) that I really don't like when getting the info for a single process, so it is pretty memory hungry for what we are trying to do (just getting mem value of the process).  I have simplified slimed that down as much as I could using the existing functions of `sys-proctable` in a extra data initializer I was using for a `MiqServer` process:

https://gist.github.com/NickLaMuro/2d494d9fc0050f4e705ebab58437d108#file-miq_server_mem_log_patch-local-rb-L20

But it probably would be better to consider pushing a common interface into `sys-proctable` for getting simple things like memory usage and memory info.  Something that has been on my `WANTTODO` list, but haven't had the cycles to go and do it.  #firstWorldMemoryLeakProblems

NickLaMuro: I also have a complex version of this that uses whats available in `manageiq-performance` (in a overly long WIP PR):

https://github.com/ManageIQ/manageiq-performance/pull/27/commits/0d9a3ee

Lifts some parts of that from the `get_process_mem` gem, and uses sys-proctable as a fall back (did not know about the faster version I came up with above when I wrote that, FYI).  Since `get_process_mem` portion is reading from the `/proc/pid/status` file is most undoubtedly faster, since it is much shorter.

The other thing I am curious about is if for something like this, where we are reading from the same file A BUNCH of times (well, file-like-interface), if maintaining a single keeping a File object memoize, and opening, closing, and rewinding as needed would be more performant.  Would something interesting to try.

NickLaMuro:
**NOTE:**  I started typing this up prior to doing my review from above, left it, then came back to add a few things before pressing enter... there might be some continuity issues between the two... but too lazy to edit them...

> Perhaps the https://github.com/ManageIQ/manageiq-performance repo is a better fit?

So I only skimmed this this morning, but I ironically was kind of thinking the same thing before I read this question myself.

I have an open PR with some similar functionality in it that I lifted from [`derailed_benchmarks`](https://github.com/schneems/derailed_benchmarks) (includes [a patch](https://github.com/schneems/derailed_benchmarks/pull/104) that fixes `require_relative` not being factored in) that should be close to this:

https://github.com/ManageIQ/manageiq-performance/pull/27
https://github.com/ManageIQ/manageiq-performance/pull/27/commits/037ec57a

But currently doesn't have any tree view support, and I still have some cleanup that I want to do, like adding modularization with the generated script views and such, but definitely something we could tweak to make with this effort.

I would have to compare the code for what is done in the derailed `require_tree` stuff and what you have here and see what is more accurate and such.  Haven't looked recently, but I think there is something that is done better with the require logging stuff that you do that works better with some of our custom code, but I forget.

* * *

#### Sidenote 1

I also had an effort for adding this to travis a while back here:

https://github.com/ManageIQ/manageiq/pull/14261

Which I ran into some troubles and such with it running properly on travis, but there is a sample output of it here:

https://s3.amazonaws.com/archive.travis-ci.org/jobs/211965745/log.txt

Not necessarily anything that needs to be merged, but just something that gives you an idea of what this output of the `derailed_benchmark` version of this looks like.


* * *

#### Sidenote 2

Also, I am reminded of this flamegraph rewrite PR for `stackprof`:

https://github.com/tmm1/stackprof/pull/74

Which a comment that sticks out is the one by Sam regarding moving the flamegraph code for this into the flamegraph gem.  Seems like this is something that we could easily make use of and convert this into for those purposes, and just provide it a flamegraph file to use.

The javascript/html for that just ingests a url param that points to a datafile for the graph data, so we should be able to just convert the data we get here into a format that would accept.  I have a 8 hour or so bus ride on Thursday this week when heading home, so I might look into doing that then 😅 .

PR #124, opened 2017-04-19
Allow standby server initialization to resync the data

carbonin:
This will allow failed primary database servers to be more easily added back to the cluster as standby servers.

When there is existing data in the data directory the console will now prompt the user about if they would like to remove the existing data before continuing with the setup (rather than just failing outright).

https://bugzilla.redhat.com/show_bug.cgi?id=1426718
https://bugzilla.redhat.com/show_bug.cgi?id=1426769

### DIFF
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -13,7 +13,7 @@ module ApplianceConsole
     REGISTER_CMD    = 'repmgr standby register'.freeze
     REPMGRD_SERVICE = 'rh-postgresql95-repmgr'.freeze
 
-    attr_accessor :disk, :standby_host, :run_repmgrd_configuration
+    attr_accessor :disk, :standby_host, :run_repmgrd_configuration, :resync_data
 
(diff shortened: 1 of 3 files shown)


carbonin:
Screenshots on the way.

carbonin:
screenshot from 2017-04-20 15-58-13

After this PR, this is what the whole flow looks like.
Removing WIP.

carbonin:
While testing this change I came across a condition where I tried to bring a failed primary back up as a standby with the same node id that it had as a primary. This works fine until we try to register the node with `repmgr standby register`, which fails.

This means that the replication slot gets created on the primary with the requested node id and the server starts replicating, but repmgr is not set up correctly. I would rather not undo that work if things fail so in a follow-up PR I'll check the primary for the node id we want before doing the sync. If we find it and it isn't active we can `--force` the register if the user confirms the action which should put us in a good state.

This will involve connecting to the primary server to check for the existence of a node with the requested id, but I think that's better than cleaning up the server after the register fails.

carbonin:
Closing in favor of combined PR #126

PR #126, opened 2017-04-21
Check for existing nodes before attempting registration

carbonin:
This will allow failed primary database servers to be more easily added back to the cluster as standby servers.

When there is existing data in the data directory the console will now prompt the user about if they would like to remove the existing data before continuing with the setup (rather than just failing outright).

Additionally, this implements a check during the "ask questions" phase of setting up a standby server which will inform the user if they are trying to use a node number which is already assigned to a node record.

Before this change the initial clone and replication setup would succeed, but the registration would fail if a user tried to re-use a node number (from a previously failed node, for example). This would leave you in a state where the new standby server is started and is replicating from a new replication slot on the primary, but cannot join the repmgr cluster.

This can be avoided by registering with the `--force` flag, but we don't really want to do that unconditionally so we have to connect to the primary to determine if the requested node number is in use *before* doing the clone then prompt the user for confirmation.

This fixes a number of BZs:

https://bugzilla.redhat.com/show_bug.cgi?id=1426718 - Make the process of reintroducing a failed HA node more user-friendly
https://bugzilla.redhat.com/show_bug.cgi?id=1426769 - After reintroducing a failed primary node, there are old replication slots left on the "new" node
https://bugzilla.redhat.com/show_bug.cgi?id=1442911 - In new db master node, pg_xlog directory got fulled

### DIFF
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -13,7 +13,7 @@ module ApplianceConsole
     REGISTER_CMD    = 'repmgr standby register'.freeze
     REPMGRD_SERVICE = 'rh-postgresql95-repmgr'.freeze
 
-    attr_accessor :disk, :standby_host, :run_repmgrd_configuration, :resync_data
+    attr_accessor :disk, :standby_host, :run_repmgrd_configuration, :resync_data, :force_register
 
     def initialize
       self.cluster_name      = nil
@@ -35,6 +35,7 @@ module ApplianceConsole
       ask_for_database_credentials
       ask_for_standby_host
       ask_for_repmgrd_configuration
+      return false unless node_number_valid?
       return false if repmgr_configured? && !confirm_reconfiguration
       confirm
     end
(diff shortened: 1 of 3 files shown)


carbonin:
Screenshot:

screenshot from 2017-04-24 16-20-04

simaishi:
Fine backport details:

```
$ git log -1
commit 12fcafd08821a888bcd20ca3cf001653046c641e
Date:   Mon Apr 24 16:33:03 2017 -0400

    Merge pull request #126 from carbonin/failfast_for_duplicate_node_ids
    
    Check for existing nodes before attempting registration
    (cherry picked from commit 34299c6baa225d0492f42e4caf61d189dc468c48)
```

simaishi:

```
$ git log -1
commit 03855d22c4adaeb7828be390b500162b9239804a
Date:   Mon Apr 24 16:33:03 2017 -0400

    Merge pull request #126 from carbonin/failfast_for_duplicate_node_ids
    
    Check for existing nodes before attempting registration
    (cherry picked from commit 34299c6baa225d0492f42e4caf61d189dc468c48)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1450511
    https://bugzilla.redhat.com/show_bug.cgi?id=1445383
    https://bugzilla.redhat.com/show_bug.cgi?id=1450512
```

PR #127, opened 2017-04-22
Use version 0.16.0 of the 'ovirt' gem

jhernand:
The new version of the gem fixes the following problems:

  Add VM FQDN to guest_info
  https://bugzilla.redhat.com/1417965

  Fix handling of version 4.y in Vm.attach_payload
  https://bugzilla.redhat.com/1441746

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -50,7 +50,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "net-sftp",                "~>2.1.2"
   s.add_runtime_dependency "nokogiri",                "~>1.7.1"
   s.add_runtime_dependency "openscap",                "~>0.4.3"
-  s.add_runtime_dependency "ovirt",                   "~>0.15.1"
+  s.add_runtime_dependency "ovirt",                   "~>0.16.0"
   s.add_runtime_dependency "pg",                      "~>0.18.2"
   s.add_runtime_dependency "pg-dsn_parser",           "~>0.1.0"
   s.add_runtime_dependency "psych",                   "~>2.0.12"


jhernand:
@bdunne @pkliczewski @masayag @borod108 please review.

Note that I believe that this dependency should be moved, or at least replicated, to the oVirt provider.

masayag:
@jhernand Could you mention also https://github.com/ManageIQ/ovirt/pull/79 which fixed https://bugzilla.redhat.com/show_bug.cgi?id=1417965 "Add VM FQDN to guest_info" to the commit message ?

manageiq-gems-pending.gemspec
```diff
@@ -50,7 +50,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "net-sftp",                "~>2.1.2"
   s.add_runtime_dependency "nokogiri",                "~>1.7.1"
   s.add_runtime_dependency "openscap",                "~>0.4.3"
-  s.add_runtime_dependency "ovirt",                   "~>0.15.1"
+  s.add_runtime_dependency "ovirt",                   "~>0.16.0"
```
bdunne: This is becoming a problem with several providers.  I think we should loosen the dependency here and this should become:
```ruby
s.add_runtime_dependency "ovirt" # Version specified by manageiq-providers-ovirt
```
@Fryguy Any thoughts?

isimluk: The same reasoning could lead us to droping the dependency altogether, when we expect other project to take care.

jhernand:
@masayag updated the description of the pull request and the commit messages to reference both bugs.

manageiq-gems-pending.gemspec
bdunne: I would like to drop the dependency, but there is code here that still depends on it.  Until then, we we have a dependency conflict where we have to merge 2 repos at the same time to keep Travis green.

bdunne:
@jhernand Based on #133 please move this change to manageiq-providers-ovirt.

jhernand:
Here is the pull request for the oVirt provider: ManageIQ/manageiq-providers-ovirt#20.

I am therefore closing this one.

PR #130, opened 2017-04-24
Fix undefined vmConfig method for MiqOpenstackeImage and shorter OS device signature.

hsong-rh:
Fix the following corner cases:
1. OS device name is shorted than expect;
2. Both MiqOpenstackImage and MiqOpenstaceInstance don't have vmConfig defined;

https://bugzilla.redhat.com/show_bug.cgi?id=1372672

### DIFF
--- a/lib/gems/pending/metadata/MIQExtract/MIQExtract.rb
+++ b/lib/gems/pending/metadata/MIQExtract/MIQExtract.rb
@@ -158,10 +158,10 @@ class MIQExtract
   end
 
   def getVMConfig(_c)
-    # Get VM config in XML format
-    config_xml = @target.vmConfig.toXML(true, @target)
-
     begin
+      # Get VM config in XML format
+      config_xml = @target.vmConfig.toXML(true, @target)
+
       # Log snapshot data for diagnostic purposes
       config_xml.find_each("//vm/snapshots") do |s|
         formattedXml = ""
--- a/lib/gems/pending/metadata/util/win32/system_path_win.rb
+++ b/lib/gems/pending/metadata/util/win32/system_path_win.rb
@@ -95,7 +95,7 @@ module Win32
       # Decode disk signature
       unless result[:os_device].nil?
         d = result[:os_device].split(',')
-        result[:disk_sig] = (d[59] + d[58] + d[57] + d[56]).to_i(16)
+        result[:disk_sig] = (d[59] + d[58] + d[57] + d[56]).to_i(16) if d.size > 60
       end
       result
     end


hsong-rh:
@roliveri Please review.

lib/gems/pending/metadata/MIQExtract/MIQExtract.rb
```diff
@@ -158,10 +158,10 @@ def extract(category, &blk)
   end
 
   def getVMConfig(_c)
-    # Get VM config in XML format
-    config_xml = @target.vmConfig.toXML(true, @target)
-
     begin
+      # Get VM config in XML format
+      config_xml = @target.vmConfig.toXML(true, @target)
```
roliveri: I'm not sure this is the right solution to the problem.

Moving this call into the begin block will cause any failures to be silent.
This call shouldn't fail. The root cause of the problem is in the `MiqOpenStackImage.rb` and `MiqOpenStackInstance.rb` files. The corresponding objects delegate calls to a `MiqVm` object. The problem is `vmConfig` is not in the `SUPPORTED_METHODS` array, so the code doesn't know to delegate that method.

The proper solution may be just to add `:vmConfig` to the `SUPPORTED_METHODS` arrays.

lib/gems/pending/metadata/util/win32/system_path_win.rb
```diff
@@ -95,7 +95,7 @@ def self.parse_bcd(fs)
       # Decode disk signature
       unless result[:os_device].nil?
         d = result[:os_device].split(',')
-        result[:disk_sig] = (d[59] + d[58] + d[57] + d[56]).to_i(16)
+        result[:disk_sig] = (d[59] + d[58] + d[57] + d[56]).to_i(16) if d.size > 60
```
roliveri: This change seems to be unrelated to the previous. It probably should be in a separate PR.

What happens when `d.size <= 60`? Is that a valid disk?

lib/gems/pending/metadata/MIQExtract/MIQExtract.rb
hsong-rh: I tried to add :vmConfig into SUPPORTED_METHODS array, it didn't solve the problem. All the following called methods were complaint not defined.

I agree the failures should be silent here. I'll change to raise them to caller.

lib/gems/pending/metadata/util/win32/system_path_win.rb
hsong-rh: I can send another PR for this change. It's a valid disk even d.size <= 60. In our case here, d.size = 22. Except this signature, all other information are meaningful and valid. 

Actually I didn't find this os_device signature is used anywhere later.

lib/gems/pending/metadata/MIQExtract/MIQExtract.rb
roliveri: If you added :vmConfig to the  SUPPORTED_METHODS array and it still didn't work, then something else is wrong, because that SHOULD work. You need to figure out why it's not working.

roliveri: You can't just raise or log an error here. Errors in the original block SHOULD be silent, it's only the new call added to the block that shouldn't be silent.

Here again, this should all be moot because the call to vmConfig should work.

hsong-rh:
Will split into two PRs and fix in a different way. Closed this request.

PR #133, opened 2017-04-25
Move 'ovirt' version dependency to provider plugin

bdunne:
Specifying a "~>x.y.z" on 'ovirt' in both manageiq-gems-pending and manageiq-providers-ovirt makes it really hard to not break Travis for at least a short on one of many repos.  It's also painful for developers who try to `bundle update` while there are conflicting dependencies.  This happens frequently and it is often merged in one repo, but then tests fail in the other repo leaving us in a state where you can not `bundle` for an unreasonable amount of time.
Since I see manageiq-gems-pending as technical debt, I'm loosening the requirement here and giving control to the provider plugin.  Eventually all dependencies on ovirt in this repo should be removed anyway.

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -50,7 +50,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "net-sftp",                "~>2.1.2"
   s.add_runtime_dependency "nokogiri",                "~>1.7.1"
   s.add_runtime_dependency "openscap",                "~>0.4.3"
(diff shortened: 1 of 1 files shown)


carbonin:
Can we do the same thing with `azure-armrest`?

We had this exact issue with that gem recently.

bdunne:
Yes, and others too (openstack things).  I was going to take care of them in followup PRs unless you want them all here.

I was trying to resolve https://github.com/ManageIQ/manageiq-gems-pending/pull/127 with this PR.  That PR can't even `bundle` in Travis...

carbonin:
A followup sounds good.

PR #135, opened 2017-04-25
Let VMDBLogger.log_hashes filter out the sensitive value.

lfu:
VMDBLogger.log_hashes would log the sensitive value as FILTERED.

Blocks [#manageiq/pull/14878](https://github.com/ManageIQ/manageiq/pull/14878).

@miq-bot assign @gmcculloug 
@miq-bot add_label fine/yes, enhancement

### DIFF
--- a/lib/gems/pending/util/vmdb-logger.rb
+++ b/lib/gems/pending/util/vmdb-logger.rb
@@ -108,9 +108,13 @@ class VMDBLogger < Logger
     filter = Array(options[:filter]).flatten.compact.map(&:to_s) << "password"
     filter.uniq!
 
-    YAML.dump(h).split("\n").each do |l|
+    values = YAML.dump(h).gsub(MiqPassword::REGEXP, "[FILTERED]")
+    values.split("\n").each do |l|
       next if l[0...3] == '---'
-      logger.send(level, "  #{l}") unless filter.any? { |f| l.include?(f) }
+      if (key = filter.detect { |f| l.include?(f) })
+        l.gsub!(/#{key}.*: (.+)/) { |m| m.gsub!($1, "[FILTERED]") }
+      end
+      logger.send(level, "  #{l}")
     end
   end
 
--- a/spec/util/vmdb-logger_spec.rb
+++ b/spec/util/vmdb-logger_spec.rb
@@ -1,4 +1,5 @@
 require 'util/vmdb-logger'
+require 'util/miq-password'
 
 describe VMDBLogger do
   describe "#log_hashes" do
@@ -10,7 +11,7 @@ describe VMDBLogger do
       logger.log_hashes(hash)
 
       buffer.rewind
-      expect(buffer.read).to_not include("pa$$w0rd")
+      expect(buffer.read).to include(":password: [FILTERED]")
     end
 
     it "filters out passwords when keys are strings" do
@@ -18,7 +19,7 @@ describe VMDBLogger do
       logger.log_hashes(hash)
 
       buffer.rewind
-      expect(buffer.read).to_not include("pa$$w0rd")
+      expect(buffer.read).to include("password: [FILTERED]")
     end
 
     it "with :filter option, filters out given keys and passwords" do
@@ -26,7 +27,9 @@ describe VMDBLogger do
       logger.log_hashes(hash, :filter => :extra_key)
 
       buffer.rewind
-      expect(buffer.read).to_not include("pa$$w0rd")
+      message = buffer.read
+      expect(message).to include(':extra_key: [FILTERED]')
+      expect(message).to include(':password: [FILTERED]')
     end
 
     it "when :filter option is a Set object, filters out the given Set elements" do
@@ -34,7 +37,34 @@ describe VMDBLogger do
       logger.log_hashes(hash, :filter => %i(bind_pwd password amazon_secret).to_set)
 
       buffer.rewind
-      expect(buffer.read).to_not include("pa$$w0rd")
+      message = buffer.read
+      expect(message).to include(':bind_pwd: [FILTERED]')
+      expect(message).to include(':amazon_secret: [FILTERED]')
+      expect(message).to include(':password: [FILTERED]')
+    end
+
+    it "filters out encrypted value" do
+      hash = {:a => {:b => 1, :extra_key => "v2:{c5qTeiuz6JgbBOiDqp3eiQ==}"}}
+      logger.log_hashes(hash)
+
+      buffer.rewind
+      expect(buffer.read).to include(':extra_key: [FILTERED]')
+    end
+
+    it "filters out root_password" do
+      hash = {"a" => {"b" => 1, "root_password" => "pa$$w0rd"}}
+      logger.log_hashes(hash)
+
+      buffer.rewind
+      expect(buffer.read).to include("root_password: [FILTERED]")
+    end
+
+    it "filters out password_for_important_thing" do
+      hash = {:a => {:b => 1, :password_for_important_thing => "pa$$w0rd"}}
+      logger.log_hashes(hash)
+
+      buffer.rewind
+      expect(buffer.read).to include(":password_for_important_thing: [FILTERED]")
     end
   end
 


gmcculloug:
@Fryguy @jrafanie @bdunne 
Looking for feedback on the enhancements to this logging method as we are looking to use it in other areas.  Initially in PR https://github.com/ManageIQ/manageiq/pull/14878 but thinking this could also replace the `dump_obj` calls for most of the providers during provisioning.  (The VMware provision has some specific logging requirements, but others could switch to use this common method.)

I think the biggest change here is that the password lines would still be logged, so you could see the key exists, but the value would be marked as "[FILTERED]".  Additionally any encrypted values would also be filtered out.

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -108,9 +108,13 @@ def self.log_hashes(logger, h, options = {})
...
     filter.uniq!
 
-    YAML.dump(h).split("\n").each do |l|
+    values = YAML.dump(h).gsub(MiqPassword::REGEXP) { |m| m.gsub($2, "FILTERED") }
+    values.split("\n").each do |l|
       next if l[0...3] == '---'
-      logger.send(level, "  #{l}") unless filter.any? { |f| l.include?(f) }
+      if key = filter.detect { |f| l.include?(f) }
+        l.gsub!(/#{key}.*: (.+)/) { |m| m.gsub!($1, "[FILTERED]") }
+      end
```
carbonin: This looks like it is changing behavior in an interesting way.

Before this, the entire line which contained a key in `options[:filter]` was not logged, but now the line has to match a regex and the key name is logged. Why make that change in addition to the filtering for `MiqPassword::REGEXP`?

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -108,9 +108,13 @@ def self.log_hashes(logger, h, options = {})
     filter = Array(options[:filter]).flatten.compact.map(&:to_s) << "password"
     filter.uniq!
 
-    YAML.dump(h).split("\n").each do |l|
+    values = YAML.dump(h).gsub(MiqPassword::REGEXP) { |m| m.gsub($2, "FILTERED") }
```
carbonin: Is there a reason this is "FILTERED" and the other is "[FILTERED]" (with square brackets)?

lib/gems/pending/util/vmdb-logger.rb
carbonin: Ah I see @gmcculloug mentioned this:

> I think the biggest change here is that the password lines would still be logged, so you could see the key exists,

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -108,9 +108,13 @@ def self.log_hashes(logger, h, options = {})
...
+    values.split("\n").each do |l|
       next if l[0...3] == '---'
-      logger.send(level, "  #{l}") unless filter.any? { |f| l.include?(f) }
+      if key = filter.detect { |f| l.include?(f) }
+        l.gsub!(/#{key}.*: (.+)/) { |m| m.gsub!($1, "[FILTERED]") }
```
carbonin: Instead of doing string manipulation here, why not just dup the hash passed in and alter the value before doing the `YAML.dump`? That seems like it would be much more reliable than this regex dance.

carbonin: Unless of course you're intentionally trying to match keys like "password-but-not-really-a-password"?

lib/gems/pending/util/vmdb-logger.rb
gmcculloug: The substitution here ends up looking like `v2:{FILTERED}` where the braces come from the encrypted string.  Adding `[...]` into the mix seemed useless.

lib/gems/pending/util/vmdb-logger.rb
gmcculloug: The method was doing a YAML.dump before which means we do not have to be concerned with the depth of embedded objects.

If we stick with the YAML string I think we could do a `gsub!` for any filters over the whole string outside of the  `.each` loop.  Also, would `.each_line` work here instead of `.split("\n")`?

lib/gems/pending/util/vmdb-logger.rb
carbonin: Oh weird, so if we remove the capture groups from that regex then this will probably break. Couldn't we just not even print the v2 part? Is that really that important?

gmcculloug: This is based on an existing regex constant from `MiqPassword`.  Would just need to create a new constant there that captures the entire string.  Wasn't sure if it was worth it initially.

carbonin: I think this would still work with the existing regex:

```ruby
irb(main):002:0> "stuff...v2:{asbvawerfab}..morestuff".gsub(MiqPassword::REGEXP, "[FILTERED]")
=> "stuff...[FILTERED]..morestuff"
```

lib/gems/pending/util/vmdb-logger.rb
carbonin: So this would match `password_for_important_things: s3cr3t`, but not `root_password: s3cr3t`

If we want it to match both we should test for that.

spec/util/vmdb-logger_spec.rb
```diff
@@ -10,31 +11,47 @@
       logger.log_hashes(hash)
 
       buffer.rewind
-      expect(buffer.read).to_not include("pa$$w0rd")
+      message = buffer.read
+      expect(message).to_not include("pa$$w0rd")
+      expect(message).to include("FILTERED")
```
carbonin: Can we change these specs to all compare against the expected message value directly?

That would make what `log_hashes` is doing a whole lot more clear.

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -108,9 +108,13 @@ def self.log_hashes(logger, h, options = {})
...
+      end
+      logger.send(level, "  #{l}")
```
bdunne: I know this isn't changed in this PR, but what is the value of writing a bunch of individual log lines rather than a single multi-line message?

lib/gems/pending/util/vmdb-logger.rb
lfu: @gmcculloug ```.each_line``` would preserve the ```\n``` at the end of each line which would print out an extra blank line for each message in the log.

Fryguy:
My concern is that previously, this was intentionally a pretty blunt object.  We filtered line by line and did a `line.include?` with the filter key, so although you might get false positives, it would kill the line.  Now, the regex introduces a potential for something to slip through.  I'm wondering if it's cleaner to still just do a `line.include`?

lfu:
@Fryguy Updated.

Fryguy:
I'm good with this, but I don't know why the bot is not kicking a rubocop check here.

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -108,9 +108,13 @@ def self.log_hashes(logger, h, options = {})
     filter = Array(options[:filter]).flatten.compact.map(&:to_s) << "password"
     filter.uniq!
 
-    YAML.dump(h).split("\n").each do |l|
+    values = YAML.dump(h).gsub(MiqPassword::REGEXP, "[FILTERED]")
+    values.split("\n").each do |l|
       next if l[0...3] == '---'
-      logger.send(level, "  #{l}") unless filter.any? { |f| l.include?(f) }
+      if key = filter.detect { |f| l.include?(f) }
```
Fryguy: @lfu Can you fix the bot warning...to solve it, just use parens around your if conditional...

```ruby
if (key = filter.detect { |f| l.include?(f) })

**Comment by simaishi** [comment]
Fine backport details:

```
$ git log -1
commit 8f6cd84d1290af555639ac279d408d9ca605b524
Date:   Tue May 9 11:08:13 2017 -0400

    Merge pull request #135 from lfu/log_hashes
    
    Let VMDBLogger.log_hashes filter out the sensitive value.
    (cherry picked from commit 799ba5629dc9159a35e61f4d7730af704f874b03)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1458434
```

PR #136, opened 2017-04-25
Remove unnecessary dependency

bdunne:
Nothing in manageiq-gems-pending uses it and I don't see any reason to lock it down here.

As far as I know it is used and depended on by:
azure-armrest
foreman_api_client
hawkular-client
image-inspector-client
kubeclient
ovirt


bdunne:
@jrafanie Please review.  Last I remember you were involved maybe for IPv6 compatability?

jrafanie:
LGTM, no mention of RestClient constant anywhere in this repo.  The extracted repos will manage the dependency.

PR #141, opened 2017-04-26
Support versionless Keystone endpoints

tzumainn:
https://bugzilla.redhat.com/show_bug.cgi?id=1444650

### DIFF
--- a/lib/gems/pending/openstack/openstack_handle/handle.rb
+++ b/lib/gems/pending/openstack/openstack_handle/handle.rb
@@ -197,6 +197,17 @@ module OpenstackHandle
 
         raw_service = self.class.raw_connect_try_ssl(username, password, address, port, service, opts, api_version,
                                                      security_protocol)
+
+        # need to check if this is versionless keystone endpoint
+        if service == "Identity"
+          identity_prefix = (api_version == "v2" ? "v2.0" : "v3")
+          unless raw_service.credentials[:openstack_management_url].include?(identity_prefix)
+            opts[:openstack_identity_prefix] = identity_prefix
+            raw_service = self.class.raw_connect_try_ssl(username, password, address, port, service, opts, api_version,
+                                                         security_protocol)
+          end
+        end
+
         service_wrapper_name = "#{service}Delegate"
         # Allow openstack to define new services without explicitly requiring a
         # service wrapper.


tzumainn:
@miq-bot add_label blocker,bug,fine/yes

tzumainn:
@Ladas Could you tell me whether you think this solution looks all right (and suggest an alternative if you think there's a better way to do this)?

I did successfully test this fix against both versioned and a versionless Keystone endpoints.

lib/gems/pending/openstack/openstack_handle/handle.rb
```diff
@@ -197,6 +197,17 @@ def connect(options = {})
 
         raw_service = self.class.raw_connect_try_ssl(username, password, address, port, service, opts, api_version,
                                                      security_protocol)
+
+        # need to check if this is versionless keystone endpoint
+        if service == "Identity"
+          identity_prefix = (api_version == "v2" ? "v2.0" : "v3")
+          unless raw_service.credentials[:openstack_management_url].include?(identity_prefix)
+            opts[:openstack_identity_prefix] = identity_prefix
+            raw_service = self.class.raw_connect_try_ssl(username, password, address, port, service, opts, api_version,
```
Ladas: seems like we call raw_connect_try_ssl also on the line 198, can we add condition for this, so it doesn't run twice for identity?

durandom:
@tzumainn @aufi is moving the code away from gems-pending in https://github.com/ManageIQ/manageiq-providers-openstack/pull/14

Ladas:
@tzumainn btw. I think we should move the OpenStack code from here to the manageiq-providers-penstack gem, cc @durandom , right?

durandom:
@Ladas just mentioned  it's a blocker, so this should be merged before the code move

Ladas:
👍 to first merging this, the moving the code

aufi:
Agreed, added WIP label to https://github.com/ManageIQ/manageiq-providers-openstack/pull/14

lib/gems/pending/openstack/openstack_handle/handle.rb
tzumainn: @Ladas that's the part that I couldn't figure out; the reason we run it twice is because the only I could figure out to tell if keystone is versionless, is to connect to the service and look at openstack_management_url.  Note that we only run it twice if it's versionless.

Ladas: right, that makes sense

Ladas:
@miq-bot assign @blomquisg 

looks good to merge 👍

lib/gems/pending/openstack/openstack_handle/handle.rb
blomquisg: I wonder if (in a completely separate PR) the `raw_connect_try_ssl` could be refactored away entirely since I think we do a decent job of collecting whether the user is using SSL or not.  Whereas before we sort of had to guess...

Again, not for this PR.

blomquisg:
Something for another PR, maybe we should move these library functions over to the [`manageiq-providers-openstack`](https://github.com/manageiq/manageiq-providers-openstack) repo now that it's separated.  I don't see this being a useful gem outside of MIQ, and if there's anything that really truly universally useful, it probably need to be pushed back up to [`fog-openstack`](https://github.com/fog/fog-openstack).

Just a thought.  Definitely not a high priority and definitely not something to hold up this PR, but it would provide a bit more independence for updating these library functions in the future.

simaishi:
Fine backport details:

```
$ git log -1
commit acd3f1ed50b7fd8f3f74fcdc85500a3a2f06d156
Date:   Wed Apr 26 11:10:20 2017 -0400

    Merge pull request #141 from tzumainn/openstack-identity-specify-version
    
    Support versionless Keystone endpoints
    (cherry picked from commit a73a80be5418984155fc3eea443605232c718972)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1445893
```

PR #145, opened 2017-04-27
Fix Openstack Metrics resource_id field

aufi:
Openstack Gnocchi metrics store in some cases resource_id under
metadata, which changes its name to metadata.resource_id.
Updating code to handle it in same way.

https://bugzilla.redhat.com/show_bug.cgi?id=1443903

### DIFF
--- a/lib/gems/pending/openstack/openstack_handle/metric_delegate.rb
+++ b/lib/gems/pending/openstack/openstack_handle/metric_delegate.rb
@@ -17,6 +17,7 @@ module OpenstackHandle
     def list_meters(filters)
       data = []
       filters.each do |filter|
+        filter['field'] = 'resource_id' if filter['field'] == 'metadata.resource_id'
         next if filter['field'] == 'metadata.instance_id'
         raise "Unexpected Metric filter \"#{filter['field']}\"" unless filter['field'] == 'resource_id'
         res = resources.find_by_id(filter['value'])


aufi:
@miq-bot add_label bug,blocker,fine/yes

tzumainn:
:+1: looks good to me!  I've successfully tested this PR on an OSP10 and an OSP11 undercloud.

lib/gems/pending/openstack/openstack_handle/metric_delegate.rb
```diff
@@ -17,6 +17,7 @@ def initialize(dobj, os_handle, name)
     def list_meters(filters)
       data = []
       filters.each do |filter|
+        filter['field'] = 'resource_id' if filter['field'] == 'metadata.resource_id'
         next if filter['field'] == 'metadata.instance_id'
         raise "Unexpected Metric filter \"#{filter['field']}\"" unless filter['field'] == 'resource_id'
```
tzumainn: Out of curiosity, is there a reason why the change couldn't be to update the 'unless' check to also accept 'metadata.resource_id'?

tzumainn:
@miq-bot assign @blomquisg

lib/gems/pending/openstack/openstack_handle/metric_delegate.rb
aufi: I wish I had a better answer, but yes, putting it into _unless_ check should have same result. I understand the extra line as an another operation than in _unless_ check, so it looked more readable to me. I can change it if needed.

aufi:
A fix with updating filter field attribute _works_. Extending _unless_ condition instead of update filter caused failure. There is some _magic_ which looks like filter change is propagated to other methods.

Metrics capture code for Openstack should be refactored. Ideally, together with dropping Ceilometer metrics support.

simaishi:
Fine backport details:

```
$ git log -1
commit 4f7327d842701af709dda047f3b01a4f06a2f26c
Date:   Fri Apr 28 11:42:39 2017 -0400

    Merge pull request #145 from aufi/fix_osp_metrics_resource_id
    
    Fix Openstack Metrics resource_id field
    (cherry picked from commit ef214dfabe1b8e9faff3c08d4608754d83f2b684)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1446739
```

PR #146, opened 2017-04-27
hawkular-client isn't referenced anywhere in manageiq-gems-pending

bdunne:
Depends on https://github.com/ManageIQ/manageiq-providers-hawkular/pull/15 and https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/13

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -34,7 +34,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "ffi",                     "~>1.9.3"
   s.add_runtime_dependency "ffi-vix_disk_lib",        "~>1.0.3"  # used by lib/VixDiskLib
   s.add_runtime_dependency "fog-openstack",           "=0.1.20"
-  s.add_runtime_dependency "hawkular-client",         "~>3.0.2"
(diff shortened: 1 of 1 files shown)


bdunne:
Removing [WIP] since the dependency is already merged to manageiq-providers-kubernetes, so the gem will be included from that plugin even if it is not yet merged to manageiq-providers-hawkular.

Fryguy:
@bdunne What about https://github.com/ManageIQ/manageiq-providers-openshift ?

bdunne:
@Fryguy The only reference I see is:
```
app/models/manageiq/providers/openshift/container_manager/refresher.rb:15:        hawk = ManageIQ::Providers::Kubernetes::ContainerManager::MetricsCapture::HawkularClient.new(ems, '_ops')
```
So, I think this gem should depend on `manageiq-providers-kubernetes`

Fryguy:
That gets tricky though because of the git references, right?

bdunne:
Yep.  Maybe we need an issue?

Fryguy:
yeah, I think an issue is good.

bdunne:
https://github.com/ManageIQ/manageiq-providers-openshift/issues/14

PR #150, opened 2017-05-04
Update hawkular client version to ~>3.0.2

israel-hdez:
There is a bug in 3.0.1 that was fixed in 3.0.2. Also, using "~>"
instead of "=" to avoid updating the dependencies for each
micro-release.

https://bugzilla.redhat.com/show_bug.cgi?id=1445702


israel-hdez:
@miq-bot add_label gem changes
@miq-bot assign @abonas

durandom:
@miq-bot assign @bdunne

PR #151, opened 2017-05-04
Simplify MiqUUID

bdunne:
Depends on:
https://github.com/ManageIQ/manageiq/pull/14997
https://github.com/ManageIQ/manageiq-automation_engine/pull/12
https://github.com/ManageIQ/manageiq-providers-vmware/pull/56

We don't need `method_missing` to delegate to UUIDTools, update callers and remove `method_missing`.
`SecureRandom.uuid` generates a new guid, updated callers accordingly.


Fryguy:
cc @roliveri just a heads up on this change with respect to the fleecing code. @bdunne is killing MiqUUID, since the functionality isn't necessary given SecureRandom does the same thing and is part of ruby stdlib, and where the code is using UUIDTools directly, it's much more obvious to just call it directly instead of via a method_missing.

PR #153, opened 2017-05-08
Move VcenterThumbPrint and ESXThumbPrint under vdlVcConnection

agrare:
Both logic paths in `MiqVm` (vc/esx) work on `@ost.miqVimVm.invObj` and this can be easily handled internally to `vdlVcConnection`.  This cleans up the code in `MiqVm` a bit and also allows us to get rid of `gems/pending/vmware` completely in preparation for `gems/pending/VMwareWebService` moving to [manageiq-providers-vmware](https://github.com/ManageIQ/manageiq-providers-vmware)

I tested this with [`lib/gems/pending/MiqVm/test/remoteVm.rb`](https://github.com/agrare/manageiq-gems-pending/blob/e60931888d6eb4b44b9fcb0229ee12df0659e400/lib/gems/pending/MiqVm/test/remoteVm.rb) and confirmed the SHA1 fingerprint is the same for VCs and ESXi hosts.

See: https://github.com/ManageIQ/manageiq-providers-vmware/issues/39


agrare:
@roliveri can you review?
cc @Fryguy

agrare:
Thanks @roliveri !  Yeah a lot more will have to go into moving the broker over than just copying the directory, it'd be great to split this up into more usable pieces as we go.

lib/gems/pending/MiqVm/MiqVm.rb
```diff
@@ -83,16 +83,9 @@ def openDisks(diskFiles)
       if @ost.miqVim
         dInfo.vixDiskInfo            = {}
         dInfo.vixDiskInfo[:fileName] = @ost.miqVim.datastorePath(df)
-        if @ost.miqVimVm && @ost.miqVim.isVirtualCenter?
-          thumb_print    = VcenterThumbPrint.new(@ost.miqVimVm.invObj.server)
-          @vdlConnection = @ost.miqVimVm.vdlVcConnection(thumb_print) unless @vdlConnection
+        if @ost.miqVimVm
+          @vdlConnection = @ost.miqVimVm.vdlVcConnection unless @vdlConnection
```
chessbyte: should this be 
```ruby
@vdlConnection ||= @ost.miqVimVm.vdlVcConnection
```
?

agrare: Yes, the same pattern was used https://github.com/agrare/manageiq-gems-pending/blob/2edcf6dfbcf4767e6faa2cdb56bf3d32464125de/lib/gems/pending/MiqVm/MiqVm.rb#L90 so changed them both

lib/gems/pending/MiqVm/MiqVm.rb
```diff
@@ -83,18 +83,11 @@ def openDisks(diskFiles)
...
-        elsif @ost.miqVimVm
-          esx_host       = @ost.miqVimVm.invObj.server
-          esx_username   = @ost.miqVimVm.invObj.username
-          esx_password   = @ost.miqVimVm.invObj.password
-          thumb_print    = ESXThumbPrint.new(esx_host, esx_username, esx_password)
-          @vdlConnection = @ost.miqVimVm.vdlVcConnection(thumb_print) unless @vdlConnection
         else
-          @vdlConnection = @ost.miqVim.vdlConnection unless @vdlConnection
+          @vdlConnection ||= @ost.miqVim.vdlConnection
           $log.debug "openDisks (ESX): using disk file path: #{dInfo.vixDiskInfo[:fileName]}"
```
agrare: @roliveri was this log line in the wrong place?  I would have thought [lines 91-95](https://github.com/ManageIQ/manageiq-gems-pending/pull/153/files#diff-e30538d47896fa341c01c677c03bae57L91) were the ESX case?

roliveri: Yes, and it seems there should be a log message for each of the 3 cases.

agrare: How about we just print `"openDisks: using disk file path: #{dInfo.vixDiskInfo[:fileName]}"` in all cases?  It looks like it isn't dependent on the method of connection

agrare: Done

Fryguy:
@agrare I merged this, so you will need to rebuild your extracted VMwareWebService repo.

agrare:
> @agrare I merged this, so you will need to rebuild your extracted VMwareWebService repo.

@Fryguy :+1: thanks just rebuilt it, good to have another gems-pending dependency down

PR #154, opened 2017-05-08
MiqApache::Control.status doesn't appear to be used anywhere

bdunne:
(and it doesn't do anything)


jrafanie:
Wow, I don't want to know if that was me.  Nice find!

PR #155, opened 2017-05-09
[FINE] Fix host targeted refresh

masayag:
Partial backport of https://github.com/ManageIQ/manageiq-providers-ovirt/pull/31

In V3 the host resources retrieved by
/ovirt-engine/api/hosts/{host:id} contains the statistics element. The
statistics element contains a reference to the host element. The xpath
used to collect similar host entries considered the host sub-element
as a resource to collect, which lead to collecting the same host
multiple times. By specifying the exact top level host path, this will
be avoided.

https://bugzilla.redhat.com/show_bug.cgi?id=1448690


simaishi:
@jhernand @bdunne please review

PR #158, opened 2017-05-15
Remove obsolete code for Openstack Storage

aufi:
Setting service_type manually is not needed since https://github.com/fog/fog/pull/3113
was merged and Fog updated in MiQ, which have happened already.


aufi:
@miq-bot add_label fine/no

PR #159, opened 2017-05-16
Drop MiqApache tech debt

bdunne:
Depends on https://github.com/ManageIQ/manageiq/pull/15110

- Remove conditional that eventually runs the same command
- Remove unused methods
:scissors: :fire:


jrafanie:
I'm good with this if this code is not called 👍 

Technically, we could deprecate the methods not already deprecated but I don't think MiqApache was ever really an externally visible constant.

PR #163, opened 2017-05-17
Apache start stop for containers

bdunne:
Adding support for starting and stopping Apache in containers.


chessbyte:
@carbonin So this PR resolves the Privileged Container gap?

PR #168, opened 2017-05-19
Remove old and unused RcuWebService code

chessbyte:


Fryguy:
This depends on ManageIQ/manageiq#15153

```
$ git grep RcuWebService
app/models/netapp_rcu.rb:1:require 'RcuWebService/RcuClientBase'
```

chessbyte:
Marking as WIP as the caller is being deleted here: https://github.com/ManageIQ/manageiq/pull/15153

Fryguy:
ManageIQ/manageiq#15153 is merged, so this is good to delete.

PR #169, opened 2017-05-19
Extract VMwareWebServices

agrare:
Extracting to new repo https://github.com/ManageIQ/vmware_web_service

Also need: https://github.com/ManageIQ/manageiq/pull/15221


manageiq-gems-pending.gemspec
```diff
@@ -32,7 +32,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "bunny",                   "~>2.1.0"
   s.add_runtime_dependency "excon",                   "~>0.40"
   s.add_runtime_dependency "ffi",                     "~>1.9.3"
```
agrare: @jerryk55 can I remove `ffi` as well?

jerryk55: @agrare I am not aware of any other use of FFI other than by ffi-vix_disk_lib.

agrare: :+1: thanks!

PR #171, opened 2017-05-22
Remove dependency on MIQEncode in MiqVimDatastore

agrare:
Remove dependency on `manageiq-gems-pending` for base64encode

### DIFF
--- a/lib/gems/pending/VMwareWebService/MiqVimDataStore.rb
+++ b/lib/gems/pending/VMwareWebService/MiqVimDataStore.rb
@@ -3,7 +3,6 @@ require 'sync'
 require 'fs/MiqFS/MiqFS'
 require 'fs/VimDatastoreFS/VimDatastoreFS'
 require 'util/miq-extensions'  # Required patch to open-uri for get_file_content
-require 'util/miq-encode'
 
 class MiqVimDataStore
   attr_reader :accessible, :multipleHostAccess, :name, :dsType, :url, :freeBytes, :capacityBytes, :uncommitted, :invObj
@@ -287,7 +286,7 @@ class MiqVimDataStore
     /(\d*)\.(\d*)/ =~ @invObj.about['version']
     raise "get_file_content not supported on [#{@invObj.about['fullName']}]" if $1.to_i < 3 || ($1.to_i == 3 && $2.to_i < 5)
 
-    fileUrl = "https://#{@invObj.server}/folder/#{MIQEncode.base64Encode(filepath)}?dsName=#{MIQEncode.base64Encode(@name)}"
+    fileUrl = encode_datastore_url(filepath, @name)
     options = {:http_basic_authentication => [@invObj.username, @invObj.password]}
     if block_given?
       open(fileUrl, options) { |ret| yield(ret) }
@@ -300,5 +299,22 @@ class MiqVimDataStore
 
   def dupObj(obj)
     obj
+  end
+
+  private
+
+  def encode_datastore_url(file_path, ds_name)
+    uri_parts = {
+      :scheme => "https",
+      :host   => @invObj.server,
+      :path   => "/folder/#{encode_path_url(file_path)}",
+      :query  => "dsName=#{encode_path_url(ds_name)}"
+    }
+
+    URI::Generic.build(uri_parts).to_s
+  end
+
+  def encode_path_url(path)
+    URI.encode(path, /[^#{URI::PATTERN::UNRESERVED}]/)
   end
 end # module MiqVimDataStore
--- a/spec/VMwareWebService/MiqVimDataStore_spec.rb
+++ b/spec/VMwareWebService/MiqVimDataStore_spec.rb
@@ -1,0 +1,41 @@
+require 'VMwareWebService/MiqVim'
+
+describe MiqVimDataStore do
+  context "#encode_datastore_url" do
+    before do
+      @invObj = double
+      allow(@invObj).to receive(:server).and_return('192.168.1.2')
+      allow(@invObj).to receive(:sic).and_return(nil)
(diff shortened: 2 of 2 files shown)


bdunne:
I'm :+1: for dropping the dependency, but...
```ruby
MIQEncode.base64Decode(Base64.encode64("abc"))
=> "YWJj\n"
Base64.decode64(MIQEncode.base64Encode("abc"))
=> "i\xB7"
```

lib/gems/pending/VMwareWebService/MiqVimDataStore.rb
```diff
@@ -287,7 +286,7 @@ def get_file_content(filepath)
     /(\d*)\.(\d*)/ =~ @invObj.about['version']
     raise "get_file_content not supported on [#{@invObj.about['fullName']}]" if $1.to_i < 3 || ($1.to_i == 3 && $2.to_i < 5)
 
-    fileUrl = "https://#{@invObj.server}/folder/#{MIQEncode.base64Encode(filepath)}?dsName=#{MIQEncode.base64Encode(@name)}"
+    fileUrl = "https://#{@invObj.server}/folder/#{Base64.encode64(filepath)}?dsName=#{Base64.encode64(@name)}"
```
bdunne: I'm :+1: for dropping the dependency, but...
```ruby
MIQEncode.base64Decode(Base64.encode64("abc"))
=> "YWJj\n"
Base64.decode64(MIQEncode.base64Encode("abc"))
=> "i\xB7"
```

Apparently this is using URI.encode and decode.  Thanks for finding this, I probably would have hit it in the future.

I think you want
```ruby
URI.encode(data, /[^#{URI::PATTERN::UNRESERVED}]/)
```

agrare: Gah you're right, badly named method :(
What do you think about removing `base64Encode` from `MIQEncode` then?  I don't see any other callers of it [ref](https://github.com/search?utf8=%E2%9C%93&q=org%3AManageIQ+MIQEncode.base64Encode&type=Code) [ref](https://github.com/search?utf8=%E2%9C%93&q=org%3AManageIQ+MIQEncode.base64Decode&type=Code)

lib/gems/pending/VMwareWebService/MiqVimDataStore.rb
```diff
@@ -287,7 +286,10 @@ def get_file_content(filepath)
     /(\d*)\.(\d*)/ =~ @invObj.about['version']
     raise "get_file_content not supported on [#{@invObj.about['fullName']}]" if $1.to_i < 3 || ($1.to_i == 3 && $2.to_i < 5)
 
-    fileUrl = "https://#{@invObj.server}/folder/#{MIQEncode.base64Encode(filepath)}?dsName=#{MIQEncode.base64Encode(@name)}"
+    file_path_url = URI.encode(filepath, /[^#{URI::PATTERN::UNRESERVED}]/)
+    ds_name_url = URI.encode(@name, /[^#{URI::PATTERN::UNRESERVED}]/)
+
+    fileUrl = "https://#{@invObj.server}/folder/#{file_path_url}?dsName=#{ds_name_url}"
```
chessbyte: Would doing something like the following be cleaner?
```ruby
fileUrl = Uri.new("https", nil, @invObj.server, ...).to_s
```

lib/gems/pending/VMwareWebService/MiqVimDataStore.rb
```diff
@@ -287,7 +286,14 @@ def get_file_content(filepath)
...
 
-    fileUrl = "https://#{@invObj.server}/folder/#{MIQEncode.base64Encode(filepath)}?dsName=#{MIQEncode.base64Encode(@name)}"
+    uri_parts = {
+      :scheme => "https",
+      :host   => @invObj.server,
+      :path   => "/folder/#{filepath}",
+      :query  => "dsName=#{@name}"
+    }
+
+    fileUrl = URI::Generic.build(uri_parts).to_s
```
bdunne: Building the URI like this is much nicer, but I think we still need to `URI.encode(data, /[^#{URI::PATTERN::UNRESERVED}]/)`.
```ruby
irb(main):001:0> MIQEncode.base64Encode("a%b&c")
=> "a%25b%26c"
irb(main):002:0> URI::Generic.build(:scheme => "https", :host => "aaa", :path => "/zzz/", :query => "dsname=a%b&c").to_s
=> "https://aaa/zzz/?dsname=a%b&c"
```
Also, do you know what special characters can be in the `filepath`?
```ruby
irb(main):003:0> URI::Generic.build(:scheme => "https", :host => "aaa", :path => "/zzz/x%y^z", :query => "dsname=a%b&c").to_s
URI::InvalidComponentError: bad component(expected absolute path component): /zzz/x%y^z
	from /home/bdunne/.rubies/ruby-2.3.3/lib/ruby/2.3.0/uri/generic.rb:769:in `check_path'
	from /home/bdunne/.rubies/ruby-2.3.3/lib/ruby/2.3.0/uri/generic.rb:817:in `path='
	from /home/bdunne/.rubies/ruby-2.3.3/lib/ruby/2.3.0/uri/generic.rb:191:in `initialize'
	from /home/bdunne/.rubies/ruby-2.3.3/lib/ruby/2.3.0/uri/generic.rb:136:in `new'
	from /home/bdunne/.rubies/ruby-2.3.3/lib/ruby/2.3.0/uri/generic.rb:136:in `build'
	from (irb):4
```

agrare: > Also, do you know what special characters can be in the filepath?

It is a VM name so some of the special characters are already URI escaped, but you're right it is definitely safer to `URI.encode()` first

PR #172, opened 2017-05-22
Extract discovery

chessbyte:
Extracted repo in https://github.com/ManageIQ/manageiq-network_discovery

Waiting on https://github.com/ManageIQ/manageiq/pull/15230 to be merged.


bdunne:
@chessbyte do you want to take `MiqIPMI` out with this?

chessbyte:
@bdunne not sure because https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_host_provision/ipmi.rb uses it.  Thoughts?

PR #173, opened 2017-05-23
Move ThumbPrint into VMwareWebService

agrare:
`ThumbPrint` is only used by `VcenterThumbPrint` and `ESXThumbPrint` for `VixDiskLib` connection so move it out of `gems/pending/util/` and into `VMwareWebService`


agrare:
cc @bdunne

PR #174, opened 2017-05-23
[WIP] Extract postgres_ha_admin into separate repo

chessbyte:
Extracted repo in https://github.com/chessbyte/postgres_ha_admin


Fryguy:
@carbonin Please review.

Fryguy:
I'm wondering if it makes sense for this to be postgres-ha_admin or pg-ha_admin (to mirror the pg-pglogical)

carbonin:
This is interesting because nothing would really "depend" on this. I'm not sure how we should handle that.

We could add it to the manageiq Gemfile, but that's not a real dependency of manageiq...

carbonin:
> I'm wondering if it makes sense for this to be postgres-ha_admin or pg-ha_admin (to mirror the pg-pglogical)

It's not really as close to postgres as the other one was. This is more about repmgr and failing over db clients.

I called my first pass at gemfying this `repmgr_client_failover`

chessbyte:
@carbonin https://github.com/ManageIQ/manageiq-appliance/blob/master/LINK/usr/bin/evm_failover_monitor.rb depends on PostgresHaAdmin

carbonin:
@chessbyte Yes, but I meant a Gemfile or gemspec dependency. In other words, how will this get installed? If it is removed from `manageiq-gems-pending` it will not be in our bundle.

carbonin:
(Unless we add it to the `manageiq-gems-pending` gemspec @Fryguy ?

chessbyte:
@carbonin `manageiq-gems-pending` will soon disappear.  Each thing should require its own dependencies.

carbonin:
> Each thing should require its own dependencies.

@chessbyte Right, which is why I raised the question originally.  If this is merged, this gem will not be an explicit or implicit dependency of any other in our application. It's almost as if `manageiq-appliance` needs a Gemfile to require system service dependencies.

Fryguy:
@carbonin I think the point is to add it to the respective gemfile where appropriate.  That being said, the way I'm thinking this be better laid out is to either

-  Include this file directly in the manageiq-appliance repo, which is the only consumer

- Extract evm_failover_monitor.rb to this new gem and rename as pg_ha_admin_failover_monitor (or something) and put it in the bin directory.  Then this gem would have to be included as part of the build, perhaps in the same place that we might require a future appliance_console gem.  Maybe manageiq-appliance needs a Gemfile of some sort?  @simaishi Thoughts?

carbonin:
When this gem is fully extracted and released, the dependency should be listed either in the file added in ManageIQ/manageiq-appliance#125 or as a dependency of the `appliance_console` gem as it controls stopping and starting the failover monitor.

chessbyte:
@bdunne Should I close this or can we use any of it?

bdunne:
Closing in favor of https://github.com/ManageIQ/manageiq-gems-pending/pull/300

PR #175, opened 2017-05-23
Remove dependency on Log4r from VMwareWebService

agrare:
Remove custom console formatter and `Log4r` dependencies and just use a standard `Logger.new(STDOUT)` for manual `VMwareWebService` tests


agrare:
cc @bdunne

PR #181, opened 2017-05-26
DB primary and standby feature

aljesusg:
Set the database like a primary o standby configuration
Issue https://github.com/ManageIQ/manageiq-gems-pending/issues/54

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -3,6 +3,8 @@ require 'pathname'
 require 'appliance_console/utilities'
 require 'appliance_console/logging'
 require 'appliance_console/database_configuration'
+require 'appliance_console/database_replication_primary'
+require 'appliance_console/database_replication_standby'
 require 'appliance_console/internal_database_configuration'
 require 'appliance_console/external_database_configuration'
 require 'appliance_console/external_httpd_authentication'
@@ -24,6 +26,8 @@ end
 module ApplianceConsole
   class Cli
     attr_accessor :options
+
+    NETWORK_INTERFACE = "eth0".freeze
 
     # machine host
     def host
@@ -53,6 +57,10 @@ module ApplianceConsole
 
     def local_database?
       database? && local?(hostname)
+    end
+
+    def replication?
+      options[:username] && options[:password] && options[:dbname] && options[:replication] && options[:replication_node]
     end
 
     def certs?
@@ -118,6 +126,8 @@ module ApplianceConsole
         opt :password, "Database Password",  :type => :string,  :short => "p"
         opt :dbname,   "Database Name",      :type => :string,  :short => "d", :default => "vmdb_production"
         opt :standalone, "Run this server as a standalone database server", :type => :bool, :short => 'S'
+        opt :replication, "Replication type", :type => :string
+        opt :replication_node, "Node of DB Cluster", :type => :integer
         opt :key,      "Create encryption key",  :type => :boolean, :short => "k"
         opt :fetch_key, "SSH host with encryption key", :type => :string, :short => "K"
         opt :force_key, "Forcefully create encryption key", :type => :boolean, :short => "f"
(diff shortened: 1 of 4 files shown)


aljesusg:
@miq-bot add_label bug, enhancement, appliance/console

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -55,6 +59,14 @@ def local_database?
       database? && local?(hostname)
     end
 
+    def db_credentials?
```
carbonin: I don't see where this is being used.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -25,6 +27,8 @@ module ApplianceConsole
   class Cli
     attr_accessor :options
 
+    NETWORK_INTERFACE = "eth0".freeze
```
carbonin: I see why you would need this, but I would rather change the `DatabaseReplicationStandby` class so that this was a parameter rather than have to define a constant in any class that uses it.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -15,7 +15,7 @@ class DatabaseReplicationStandby < DatabaseReplication
 
     attr_accessor :disk, :standby_host, :run_repmgrd_configuration, :resync_data, :force_register
 
-    def initialize
+    def initialize(_hash = {})
```
carbonin: For this to work you need to actually initialize the parameters with the values in the hash.

Alternatively because they are already `attr_accessor`s you can just set them on the instance.

lib/gems/pending/appliance_console/database_replication_primary.rb
```diff
@@ -9,7 +9,7 @@ class DatabaseReplicationPrimary < DatabaseReplication
 
     REGISTER_CMD = 'repmgr master register'.freeze
 
-    def initialize
+    def initialize(_hash = {})
```
carbonin: Same thing here. You need to either do something with the hash contents or set the attributes using the accessors on the instance.

lib/gems/pending/appliance_console/cli.rb
aljesusg: In replication? function

carbonin: If that's the only place it's used we can keep it all in the same method.

Having `#replication?` enumerate all the dependencies of that task is fine.

aljesusg:
I did tghe changes @carbonin

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -13,7 +13,7 @@ class DatabaseReplicationStandby < DatabaseReplication
     REGISTER_CMD    = 'repmgr standby register'.freeze
     REPMGRD_SERVICE = 'rh-postgresql95-repmgr'.freeze
 
-    attr_accessor :disk, :standby_host, :run_repmgrd_configuration, :resync_data, :force_register
+    attr_accessor :node_number, :database_name, :database_user, :database_password, :disk, :standby_host, :run_repmgrd_configuration, :resync_data, :force_register
```
carbonin: These are inherited from the base class. No need to add them here.

lib/gems/pending/appliance_console/database_replication_primary.rb
```diff
@@ -9,6 +9,8 @@ class DatabaseReplicationPrimary < DatabaseReplication
 
     REGISTER_CMD = 'repmgr master register'.freeze
 
+    attr_accessor :node_number, :database_name, :database_user, :database_password
```
carbonin: These are inherited from the base class. No need to add them here.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -229,6 +240,34 @@ def set_external_db
...
+    def set_replication_db
+      db_opts = {
+        :node_number       => options[:replication_node],
+        :database_name     => options[:dbname],
+        :database_user     => options[:username],
+        :database_password => options[:password]
+      }
+      case options[:replication]
+      when "primary"
+        db_replication = ApplianceConsole::DatabaseReplicationPrimary.new(db_opts)
```
carbonin: The initialize method for these classes does not take any options. This will fail in production.

aljesusg: I don't know why will fail I need to read the documentation about ^^ I'll make the changes thanks @carbonin

chessbyte:
@aljesusg The appliance console has been extracted to a new [repo](https://github.com/ManageIQ/manageiq-appliance_console).  Please move this PR there.

PR #187, opened 2017-05-31
Remove dependency on linux_block_device gem

chessbyte:
Dependent on https://github.com/ManageIQ/manageiq-smartstate/pull/6


chessbyte:
Removed `wip` label now that ManageIQ/manageiq-smartstate#6 has been merged.

PR #189, opened 2017-05-31
Remove dependency on ovirt gem

chessbyte:
Only callers seem to be in https://github.com/ManageIQ/manageiq-providers-ovirt and https://github.com/ManageIQ/manageiq-network_discovery which already declare their dependencies in their respective gemspec files.


jrafanie:
Looks good to me!  😎  👍

PR #190, opened 2017-05-31
Remove dependency on image-inspector-client gem

chessbyte:
Dependent on https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/28


chessbyte:
Removed `wip` label since ManageIQ/manageiq-providers-kubernetes#28 was merged

PR #191, opened 2017-05-31
Remove dependency on rubyzip gem

chessbyte:
Found no dependencies nor callers of this.


bdunne:
@chessbyte I think this needs to be added to `manageiq-automation_engine` and `manageiq` first.

jrafanie:
> @chessbyte I think this needs to be added to manageiq-automation_engine and manageiq first.

Wow, good point @bdunne.  I assumed incorrectly that they were there already.

chessbyte:
Removed label `wip` now that ManageIQ/manageiq-automation_engine#36 and ManageIQ/manageiq#15299 have been merged

PR #192, opened 2017-05-31
Remove dependency on kubeclient gem

chessbyte:
Dependent on https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/29


chessbyte:
Removing `wip` label now that ManageIQ/manageiq-providers-kubernetes#29 has been merged

PR #193, opened 2017-05-31
Remove code moved to manageiq-smartstate repo

chessbyte:
See https://github.com/ManageIQ/manageiq-smartstate/blob/master/Rakefile

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,16 +7,6 @@ RSpec::Core::RakeTask.new(:spec) do |t|
   t.rspec_opts = "--options #{File.expand_path(".rspec_ci", __dir__)}" if ENV['CI']
 end
 
-Rake::TestTask.new do |t|
-  t.test_files = FileList['test/ts_*.rb'] - ['test/ts_mdfs.rb', 'test/ts_metadata.rb']
-end
-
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -7,16 +7,6 @@ RSpec::Core::RakeTask.new(:spec) do |t|
...
-
-namespace :test do
-  Rake::TestTask.new(:extract)  { |t| t.test_files = ['test/ts_extract.rb'] }
-  Rake::TestTask.new(:metadata) { |t| t.test_files = ['test/ts_metadata.rb'] }
-  Rake::TestTask.new(:miq_disk) { |t| t.test_files = ['test/ts_mdfs.rb'] }
-end
-
 task :default do
   Rake::Task["spec"].invoke
   Rake::Task["test"].invoke
```
bdunne: I think this should go too, right?

jrafanie: Yes!  Good point @bdunne

chessbyte: No - there are still XML tests in `test` dir.

jrafanie: Oh, you didn't extract those yet 😊

PR #194, opened 2017-05-31
Remove dependency on memory_buffer gem

chessbyte:
Depends on https://github.com/ManageIQ/manageiq-smartstate/pull/8


chessbyte:
Removing `wip` label now that https://github.com/ManageIQ/manageiq-smartstate/pull/8 is merged.

PR #196, opened 2017-05-31
Remove dependency on vmware_web_service gem

chessbyte:
It is a direct dependency in https://github.com/ManageIQ/manageiq/blob/master/Gemfile#L87, https://github.com/ManageIQ/manageiq/blob/master/Gemfile#L87 and https://github.com/ManageIQ/manageiq/blob/master/Gemfile#L87.

Waiting on https://github.com/ManageIQ/manageiq-providers-vmware/pull/63


chessbyte:
Removing `wip` label now that https://github.com/ManageIQ/manageiq-providers-vmware/pull/63 is merged.

jrafanie:
:shipit:

PR #197, opened 2017-06-01
[FINE] Remove files that do not seem to be used

borod108:
This files depend on the ovirt-gem and seem not be used so it would be nice to remove them.


chessbyte:
@borod108 is there an upstream/master PR that this is derived from?

chessbyte:
@borod108 bump

chessbyte:
@borod108 bump

borod108:
This is not relevant anymore.

PR #199, opened 2017-06-01
Allow configuring appliance as dedicated database with appliance_console_cli

rsevilla87:
Add option to appliance_console_cli that allows configuring appliance as dedicated database

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -113,6 +113,7 @@ module ApplianceConsole
         opt :username, "Database Username",  :type => :string,  :short => 'U', :default => "root"
         opt :password, "Database Password",  :type => :string,  :short => "p"
         opt :dbname,   "Database Name",      :type => :string,  :short => "d", :default => "vmdb_production"
+        opt :standalone, "Run this server as a standalone database server", :type => :bool, :short => 'S'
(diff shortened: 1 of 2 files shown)


lib/gems/pending/appliance_console/cli.rb
```diff
@@ -113,6 +113,7 @@ def parse(args)
         opt :username, "Database Username",  :type => :string,  :short => 'U', :default => "root"
         opt :password, "Database Password",  :type => :string,  :short => "p"
         opt :dbname,   "Database Name",      :type => :string,  :short => "d", :default => "vmdb_production"
+        opt :dedicated, "Configure this server as a dedicated database instance", :type => :bool, :short => 'D', :default => false
```
carbonin: This wording was deemed confusing in prior changes. I think we should be consistent and use something like "Run this server as a standalone database server" which comes directly from https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/appliance_console/internal_database_configuration.rb#L54

carbonin: Also, I don't think we should default this option. It will be true if it is specified otherwise `InternalDatabaseConfiguration` will deal with the default value. Similar to how we do `--key`

rsevilla87: You're right @carbonin, I'll commit your proposed changes :)

PR #200, opened 2017-06-05
Remove dependency on azure-armrest gem

chessbyte:
Dependent on https://github.com/ManageIQ/manageiq-smartstate/pull/11


chessbyte:
Removing `wip` label, now that ManageIQ/manageiq-smartstate#11 has been merged.

PR #201, opened 2017-06-05
Remove dependency on openscap gem

chessbyte:
Dependent on https://github.com/ManageIQ/manageiq/pull/15300


chessbyte:
Removed label `wip` now that ManageIQ/manageiq#15300 has been merged

PR #202, opened 2017-06-06
Add 600 second cache timeout to pickup directory changes.

jvlcek:
This PR creates a 10 minute cache timeout which will result in changes made on the
remote directory replacing cached values in 10 minutes.

The IdM team replied with the following regarding the impact to MiQ of reducing entry_cache_timeout, from the default of of 90 minutes, to a few seconds, or few a minutes:

> For Web applications, the performance impact should be really low
> since the user attributes and group membership gets consulted only
> during logon, not afterwards (unlike typical operations on the OS).

Also, this change is consistent with what we document for external auth via manually configured
[SSSD/LDAP](http://manageiq.org/docs/reference/latest/auth/ldap)

Addresses:
https://bugzilla.redhat.com/show_bug.cgi?id=1447064

### DIFF
--- a/lib/gems/pending/appliance_console/external_httpd_configuration.rb
+++ b/lib/gems/pending/appliance_console/external_httpd_configuration.rb
@@ -106,6 +106,9 @@ module ApplianceConsole
           pattern = "[domain/#{Regexp.escape(domain)}].*(\n)"
           config[/#{pattern}/, 1] = "\nldap_user_extra_attrs = #{ldap_user_extra_attrs}\n"
         end
+
+        pattern = "[domain/#{Regexp.escape(domain)}].*(\n)"
+        config[/#{pattern}/, 1] = "\nentry_cache_timeout = 600\n"
       end
 
       def configure_sssd_service(config)


jvlcek:
@abellotti Please review

jvlcek:
@miq-bot add_labels appliance/console, bug, enhancement

lib/gems/pending/appliance_console/external_httpd_configuration.rb
```diff
@@ -106,6 +106,9 @@ def configure_sssd_domain(config, domain)
           pattern = "[domain/#{Regexp.escape(domain)}].*(\n)"
           config[/#{pattern}/, 1] = "\nldap_user_extra_attrs = #{ldap_user_extra_attrs}\n"
         end
+
+        pattern = "ldap_user_extra_attrs.*(\n)"
+        config[/#{pattern}/, 1] = "\nentry_cache_timeout = 600\n"
```
abellotti: just for correctness, I wonder if we shouldn't match that pattern in the domain section specifically.

```
pattern = "[domain/#{Regexp.escape(domain)}](\n.*)+ldap_user_extra_attrs.*(\n)"
config[/#{pattern}/,2] = "\nentry_cache_timeout = 600\n" 
```

jvlcek: @abellotti I had thought of that but I realized that since we just added the ldap_user_extra_attrs line to the domain section there seems to be no chance that the it will be in a different section.

abellotti: 👍

jvlcek: @abellotti Better safe than sorry. Done as suggested. Thank you for the help!

jvlcek:
@abellotti  Just FYI. I tested on a live VM.

abellotti:
LGTM!! Thanks @jvlcek for enhancing this.

PR #206, opened 2017-06-12
Remove `util/vmdb-logger` from appliance_console

NickLaMuro:
This is never used by the appliance console, as it ends up just using the default ruby logger for it's `default_logger`:

```ruby
def self.default_logger
  @default_logger ||= begin
    require 'logger'
    l = Logger.new(LOGFILE)
    l.level = Logger::INFO
    l
  end
end
```

`default_logger` is also a attr_writer on the class, so if `util/vmdb-logger` is wanted for the logger, a user can simply require that when loading the bin, and then assign the logger themselves onto the class variable.

QA
--
Confirm that the appliance console still functions just fine with this removed.

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -20,7 +20,6 @@ require 'highline/system_extensions'
 require 'rubygems'
 require 'bcrypt'
 require 'linux_admin'
-require 'util/vmdb-logger'
 require 'util/postgres_admin'
 require 'awesome_spawn'
 include HighLine::SystemExtensions


Fryguy:
I'm good with this, though I might argue the reverse approach that the default_logger should be changed to the vmdb-logger.  @chessbyte Please review as you are extracting appliance_console, and this will collide with you.

jrafanie:
I'm ok with this.  I don't think appliance_console should be required to use the vmdb_logger.

carbonin:
Just tested this out. The console still logs successfully to `/var/www/miq/vmdb/log/appliance_console.log`.
It very much seems like this isn't actually necessary unless there were only some options that were logging to somewhere else ...

I tested this on an appliance with configuring a database.

NickLaMuro:
@carbonin Thanks.  Wasn't sure how to do this myself.

But yes, kinda looked like it was just there for fun, or some where there was a setup that actually assigned the `ApplianceConsole::Logging.logger` to `Vmdb::Logger`, and they didn't want to explicitly include the `util/vmdb-logger` in the script.

Fryguy:
LGTM, then... letting @chessbyte coordinate the merge.

PR #209, opened 2017-06-13
Removes psych as a dependency

NickLaMuro:
The psych dependency was added back when some monkey patches still existed in our codebase, were merged in my Aaron Patterson, and the one from ruby core was not updated to the latest version:

https://github.com/ManageIQ/manageiq/commit/907a679

Now that we support ruby 2.3+, the psych is on version 2.1+, so this dependency is no longer needed:

```console
$ irb
irb(main):001:0> Psych
NameError: uninitialized constant Psych
from (irb):1 from /Users/nicklamuro/.rubies/ruby-2.3.3/bin/irb:11:in `<main>'
irb(main):002:0> require 'psych'
=> true
irb(main):003:0> Psych.methods
=> [:dump, :load, :quick_emit, :parse, :load_file, ... ]
irb(main):004:0> Psych.method(:dump).source_location
=> ["/Users/nicklamuro/.rubies/ruby-2.3.3/lib/ruby/2.3.0/psych.rb", 411]
irb(main):005:0>  Psych::VERSION
=> "2.1.0"
irb(main):006:0>
```

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -38,7 +38,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "nokogiri",                "~>1.7.2"
   s.add_runtime_dependency "pg",                      "~>0.18.2"
   s.add_runtime_dependency "pg-dsn_parser",           "~>0.1.0"
-  s.add_runtime_dependency "psych",                   "~>2.0.12"
   s.add_runtime_dependency "rake",                    ">=11.0"
   s.add_runtime_dependency "sys-proctable",           "~>1.1.3"
   s.add_runtime_dependency "sys-uname",               "~>1.0.1"


NickLaMuro:
Looks like I was using ruby 2.3.3, so I probably should check this with 2.3.1 since that is what is in our ManageIQ/manageiq `.travis.yml`.

NickLaMuro:
@jrafanie @Fryguy thoughts on this one?

I mostly have been having to comment out this line in development mode since I use [zeus](https://github.com/burke/zeus) for testing, and it loads psych outside of bundler, so that becomes a problem since there are conflicting versions.  #selfishPR

jrafanie:
hmm

```
05:27:02 ~/Code/manageiq (split_ui_from_backend2) (2.4.1) + git grep Psych
app/models/dialog_import_validator.rb:  rescue Psych::SyntaxError
app/models/miq_request_task/state_machine.rb:        # HACK: Psych is unable to deserialize an Exception that contains a
app/models/mixins/yaml_import_export_mixin.rb:      rescue Psych::SyntaxError => err
app/models/orchestration_template_hot.rb:  rescue Psych::SyntaxError => err
app/models/orchestration_template_vnfd.rb:  rescue Psych::SyntaxError => err
app/models/widget_import_validator.rb:  rescue Psych::SyntaxError
config/initializers/yaml_autoloader.rb:Psych::Visitors::ToRuby.prepend Module.new {
lib/patches/database_configuration_patch.rb:      rescue Psych::SyntaxError => e
lib/vmdb/config.rb:      rescue StandardError, Psych::SyntaxError => err
lib/vmdb/permission_stores/yaml.rb:        @permissions = Psych.load_file file
spec/initializers/yaml_autoloader_spec.rb:describe Psych::Visitors::ToRuby do
```

jrafanie:
I guess technically, we have a direct dependency on psych

NickLaMuro:
@jrafanie I think you are a bit confused about what this is doing.  This isn't removing it, but using the one that comes packaged with ruby by default (and is now newer), instead of an out dated one.

jrafanie:
Yeah, so why not just bump the version number?  I mean, we depend on it, so we should really list it as a dependency... we probably need it in the main repo as a dependency too

NickLaMuro:
@jrafanie Oh, I guess I see what you are getting at.  If I am understanding your thought process correctly, you think that because we have it in our Gemfile, it becomes something that is required through bundler.  Though, I think from an ManageIQ perspective, bundler won't autoload a dependency of a dependency, so it won't get loaded via a `Bundler.require` if I am not mistaken.

I guess from my perspective, what I see is that we are just having bundler download a redundant version of `psych` to fill this dependency, when the one that comes with Ruby is handling our needs just fine.

But, not really tied to that thought process, just my pedantic self thinking that it is a bit silly to do that.  Could go either way.

jrafanie:
> @jrafanie Oh, I guess I see what you are getting at. If I am understanding your thought process correctly, you think that because we have it in our Gemfile, it becomes something that is required through bundler. Though, I think from an ManageIQ perspective, bundler won't autoload a dependency of a dependency, so it won't get loaded via a Bundler.require if I am not mistaken.

It's more that we use `Psych` directly and not the `YAML` namespace so we depend on the psych gem and not the stdlib yaml library that happens to be psych today.  In other words, we actually depend on it so we should list it in our Gemfile.  It's going to get required anyway, requires has nothing to do with what I'm thinking.

NickLaMuro:
@jrafanie Okay, fine, then I guess my follow up statement to that is "why" then do we have a specific version of `psych` that we are requiring?  We use `bundler`, for example, that HEAVILY depends on rubygems, and the only dependency on a version for it that it has is through bundler's gemspec [require_rubygems_version](https://github.com/bundler/bundler/blob/master/bundler.gemspec#L31) constraint, which only really really enforcing [yelling at the user on install](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/installer.rb#L618-L627) when the version is too low.

So the thought that we "depend on it" doesn't make sense to me because it comes with ruby by default (like many other stdlib libraries we depend on), bundler isn't doing us a favor and auto-requiring it (as mentioned previously), and previous commits bumping this version are dealing with patches on older versions of the gem that should be a part of the stdlib version.

Now, if we think there going to be incompatibilities with ruby versions with psych between the different ruby versions we have developers using, I can understand that rational.  For example, ruby 2.3.1 has psych 2.0.17 (which is still newer then what we are depending on currently) in it's stdlib instead of the 2.1.0 version of psych in ruby 2.3.3.  Of note, if we still support ruby anything around 2.2.3 and lower, then we do have a problem, because the version of psych in standard lib is less than what is required here (2.0.8), but I don't think that is officially the case any more.  If we see that that discrepancy as a problem, then I am fine with hard locking a version of your choosing, but the previous commits bumping this version over the years don't lead me to believe that we have had issues with being too "bleeding edge" with psych, but become to out of date.

To further emphasize that point, of the things `git grep`'d earlier, most are `rescue Psych::SyntaxError`, and that practice predates the earlier commit that I referenced that added `psych` as a dependency in the first place (one example of that [here](https://github.com/ManageIQ/manageiq/commit/2708ba5#diff-753064975c353f02f1dbd11bb21eb7b0R6)).  Of the remaining 3, the `# HACK` comment from @Fryguy, that references a dead link, was added in 2012 (can't show here), the `.prepend` monkey patch was something you actually added in 2012, and was just converted to to use `.prepend Module` in a rubocop sweep commit, and `lib/vmdb/permission_stores/yaml.rb` uses `Psych.load_file` for some reason instead of `YAML.load_file` (would have to ask the man himself why that decision was made), and even that was done shortly after the change in https://github.com/ManageIQ/manageiq/commit/907a679.  `psych` has been upgraded since then and worked just fine, so failing to see how any of these are really a "hard dependency" on `Pysch`.

That all said, by requiring this directly, we are adding to the `$LOAD_PATH`, and increasing the number dependencies needing to be resolved and downloaded by bundler.  Technically speaking, have a indirect minimum `psych` version through our minimum supported ruby version for ManageIQ, so that does limit the surface area of versions we have to support as well.

What I am trying to determine with this questioning is if there is a reason anymore we are going out of our way at this point to say "yes, we need this version of psych to function", or if that thought process is now outdated.  It seems to be the rational for adding it to the Gemfile in the first place (again, see https://github.com/ManageIQ/manageiq/commit/907a679) was some patches were merged in the core gem that we needed and weren't in stdlib then, but now stdlib has now caught up to our needs.  Maybe something has changed that I am unaware of in my investigation that I am missing.


P.S.  Sorry this is long, and became kind of "stream of conciousness" like as I was writing it, but I don't like being wrong on the internet... :trollface:

jrafanie:
> @jrafanie Okay, fine, then I guess my follow up statement to that is "why" then do we have a specific version of psych that we are requiring?

I'm fine with not putting a version constraint or a less restrictive number that satisfies the yaml_autoloader.  If ruby 2.3.0+ gives a psych version that doesn't break it, I'm good with dropping the gem entirely.

simaishi:
@bdunne @jrafanie Adding `fine/yes` and `euwe/yes` as we can't go to ruby 2.3.6 (which comes with psych 2.1.0) for downstream unless we remove or update this dependency. Don't think removing will be a problem?

bdunne:
¯\\_(ツ)_/¯ Not sure if this comes into play https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/extensions/miq-yaml.rb cc @jrafanie @Fryguy

Fryguy:
I think it will be ok, since psych is a core dependency

jrafanie:
you should be fine @simaishi 👍

NickLaMuro:
@jrafanie 😑

simaishi:
Thanks everyone!

simaishi:
Fine backport details:

```
$ git log -1
commit ee44ce0034cc15fb2860e514db2b0ed678f50d57
Author: Chris Arcand <chrisarcand@users.noreply.github.com>
Date:   Tue Jun 13 23:19:30 2017 -0500

    Merge pull request #209 from NickLaMuro/remove_psych_dependency
    
    Removes psych as a dependency
    (cherry picked from commit c53dfb8eff383bc621bf9c1e0c408a80a301734c)
```

simaishi:

```
$ git log -1
commit db15cc8d6cbf78b0203d8dc0c1ec4dff66b98571
Author: Chris Arcand <chrisarcand@users.noreply.github.com>
Date:   Tue Jun 13 23:19:30 2017 -0500

    Merge pull request #209 from NickLaMuro/remove_psych_dependency
    
    Removes psych as a dependency
    (cherry picked from commit c53dfb8eff383bc621bf9c1e0c408a80a301734c)
```

PR #211, opened 2017-06-15
Fix bug: existed v2_key deleted even if fetch/create new v2_key fails

ailisp:
**ISSUE**: When user decide to fetch a new v2_key from appliance console, and choose overwrite existing encryption key. If the user fails to get a new v2_key, the old one is still lost, and no way to recover.

BZ: [https://bugzilla.redhat.com/show_bug.cgi?id=1430701](https://bugzilla.redhat.com/show_bug.cgi?id=1430701
                                                          )

\cc @yrudman @gtanzillo

@miq-bot add-label wip, bug

### DIFF
--- a/lib/gems/pending/appliance_console/key_configuration.rb
+++ b/lib/gems/pending/appliance_console/key_configuration.rb
@@ -7,7 +7,7 @@ require 'util/miq-password'
 module ApplianceConsole
   CERT_DIR = ENV['KEY_ROOT'] || RAILS_ROOT.join("certs")
   KEY_FILE = "#{CERT_DIR}/v2_key".freeze
-  KEY_FILE_TEMP = "#{KEY_FILE}.tmp".freeze
+  NEW_KEY_FILE = "#{KEY_FILE}.tmp".freeze
 
   class KeyConfiguration
     attr_accessor :host, :login, :password, :key_path, :action, :force
@@ -47,11 +47,11 @@ module ApplianceConsole
 
     def activate
       if !key_exist? || force
-        success_get_new = fetch_key? ? fetch_key : create_key
+        success_get_new = get_new_key
         if success_get_new
-          save_temp_key
+          save_new_key
         else
-          remove_temp_key_if_any
+          remove_new_key_if_any
           false
         end
       else
@@ -66,15 +66,15 @@ module ApplianceConsole
       end
     end
 
-    def save_temp_key
-      FileUtils.mv(KEY_FILE_TEMP, KEY_FILE)
+    def save_new_key
+      FileUtils.mv(NEW_KEY_FILE, KEY_FILE, :force => true)
     rescue StandardError => e
       say("Failed to overwrite original key, original key kept. #{e.message}")
       return false
     end
 
-    def remove_temp_key_if_any
-      FileUtils.rm(KEY_FILE_TEMP) if File.exist?(KEY_FILE_TEMP)
+    def remove_new_key_if_any
+      FileUtils.rm(NEW_KEY_FILE) if File.exist?(NEW_KEY_FILE)
     end
 
     def key_exist?
@@ -86,15 +86,15 @@ module ApplianceConsole
     end
 
     def create_key
-      MiqPassword.generate_symmetric(KEY_FILE_TEMP) && true
+      MiqPassword.generate_symmetric(NEW_KEY_FILE) && true
     end
 
     def fetch_key
       # use :verbose => 1 (or :debug for later versions) to see actual errors
       Net::SCP.start(host, login, :password => password) do |scp|
-        scp.download!(key_path, KEY_FILE_TEMP)
+        scp.download!(key_path, NEW_KEY_FILE)
       end
-      File.exist?(KEY_FILE_TEMP)
+      File.exist?(NEW_KEY_FILE)
     rescue => e
       say("Failed to fetch key: #{e.message}")
       false
@@ -105,6 +105,10 @@ module ApplianceConsole
     def ask_for_action(default_action)
       options = {'Create key' => :create, 'Fetch key from remote machine' => :fetch}
       ask_with_menu("Encryption Key", options, default_action, false)
+    end
+
+    def get_new_key
+      fetch_key? ? fetch_key : create_key
     end
   end
 end
--- a/lib/gems/pending/appliance_console/key_configuration.rb
+++ b/lib/gems/pending/appliance_console/key_configuration.rb
@@ -6,7 +6,8 @@ require 'util/miq-password'
 
 module ApplianceConsole
   CERT_DIR = ENV['KEY_ROOT'] || RAILS_ROOT.join("certs")
-  KEY_FILE = "#{CERT_DIR}/v2_key"
+  KEY_FILE = "#{CERT_DIR}/v2_key".freeze
+  KEY_FILE_TEMP = "#{KEY_FILE}.tmp".freeze
 
   class KeyConfiguration
     attr_accessor :host, :login, :password, :key_path, :action, :force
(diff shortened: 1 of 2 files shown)


ailisp:
Ready for review now

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -7,6 +7,7 @@
 module ApplianceConsole
   CERT_DIR = ENV['KEY_ROOT'] || RAILS_ROOT.join("certs")
   KEY_FILE = "#{CERT_DIR}/v2_key"
+  KEY_FILE_BACKUP = "#{CERT_DIR}/vk.bak"
```
yrudman: I think backup file should be named something like `v2_key.backup`.
Also, I would suggest to freeze string constant (as Travis recommends)

ailisp: OK. I'll rename it. Should I also add a freeze to KEY_FILE. It's not suggested by travis but rubymine.

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -45,11 +46,15 @@ def ask_question_loop
     end
 
     def activate
+      if key_exist? && force
+        backup_key
+      end
+
```
yrudman: What happened if we have good `v2_key.backup` and bad `v2_key`and fetching would fail ? - we would loose good backup.  I think flow should not alter existing back-up if fetching failing.

ailisp: If fetching fail, the fetch_key would return false, success_get_new will be false, so `remove_backup_key_if_any` and `return true` won't execute. Then we will get to the end of if and `restore_key_if_any` so the good backup will replace bad v2_key.
```ruby
      if remove_key(force)
        success_get_new = fetch_key? ? fetch_key : create_key
        if success_get_new
          remove_backup_key_if_any
          return true
        end
      else
        # probably only got here via the cli
        $stderr.puts
        $stderr.puts "Only generate one encryption key (v2_key) per installation."
        $stderr.puts "Chances are you did not want to overwrite this file."
        $stderr.puts "If you do this all encrypted secrets in the database will not be readable."
        $stderr.puts "Please backup your key and run this command again with --force-key."
        $stderr.puts
      end
      restore_key_if_any
```

ailisp: Oh. Sorry I misunderstand your means. If the backup is created by this program, it will never exist after a success or a fail fetch. If it's backup by user manually with the `v2_key.backup` name, hmm, this `backup_key` call will overwrite user's backup, that's a problem

yrudman: right, we should not alter existing files in fetch failed
may be saving current `v2_key` to temp file first ...

yrudman: another scenario: current file is good and back-up is bad, failing fetch will erase good one

ailisp: well, this scenario needs to do more changes. I need to check if the `File.cp` for backup success, if not, ask user whether to continue or not (and suggest user do a manual backup before continue)

yrudman: I was thinking that putting existing `v2_key` in temp file first and moving it to back-up only when fetch successful would be sufficient (and removing temp file in `ensure`

spec/appliance_console/key_configuration_spec.rb
```diff
@@ -80,21 +82,34 @@
...
+        it "keeps original v2_key if fetch new fails" do
+          subject.force = true
+          mock_key = Tempfile.new('v2_key')
+          mock_key.close
+          stub_const("ApplianceConsole::KEY_FILE", mock_key.path)
+          stub_const("ApplianceConsole::KEY_FILE_TEMP", mock_key.path + ".tmp")
+          expect(subject).to receive(:fetch_key).and_return(false)
+          expect(subject.activate).to be_falsey
+          expect(FileUtils).not_to receive(:mv)
+          expect(File.exist?(ApplianceConsole::KEY_FILE)).to be_truthy
```
yrudman: check for existence of file does not prove that file is the same

spec/appliance_console/key_configuration_spec.rb
```diff
@@ -80,21 +82,34 @@
...
         it "fails if key exists (no force)" do
           expect($stderr).to receive(:puts).at_least(2).times
           subject.force = false
           v2_exists(true)
-          expect(FileUtils).not_to receive(:rm)
+          expect(FileUtils).not_to receive(:mv)
           expect(Net::SCP).not_to receive(:start)
           expect(subject.activate).to be_falsey
         end
+
```
yrudman: @ailisp I think one more test should be added in the ```context "with existing key"``` (to test new behavior): creating back-up file

yrudman: @ailisp my mistake, you have removed back-up in latest edition, disregard last comment

ailisp:
@miq-bot remove-label wip

yrudman:
👍  looks good

Fryguy:
@bdunne @carbonin Please review.

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -63,6 +66,17 @@ def activate
       end
     end
 
+    def save_temp_key
+      FileUtils.mv(KEY_FILE_TEMP, KEY_FILE)
```
carbonin: Should we use `:force => true` here? (ref: https://ruby-doc.org/stdlib-2.3.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-mv)

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -63,6 +66,17 @@ def activate
...
+    rescue StandardError => e
```
carbonin: No need to specify `StardardError` here `rescue => e` will work fine (like [here](https://github.com/ManageIQ/manageiq-gems-pending/pull/211/files#diff-f53388092c798ee64087388537f48064R98))

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -45,11 +46,13 @@ def ask_question_loop
     end
 
     def activate
-      if remove_key(force)
-        if fetch_key?
-          fetch_key
+      if !key_exist? || force
+        success_get_new = fetch_key? ? fetch_key : create_key
```
carbonin: I think this would read a bit better if we move the `fetch_key? ? fetch_key : create_key` into a private method. Then we won't need the local variable.

Maybe one called `#get_new_key`?

ailisp: Yes. this make sense.

lib/gems/pending/appliance_console/key_configuration.rb
ailisp: Added, Thank you

lib/gems/pending/appliance_console/key_configuration.rb
ailisp: You suggested was the one I originally used. But Yuri has different idea on this. He suggest to use a common parent class of the exception we may encountered. And we made sure all exceptions FileUtils.mv made is a subclass of StandardError, so I'll leave this unchanged. Thank you.

lib/gems/pending/appliance_console/key_configuration.rb
```diff
@@ -45,11 +46,13 @@ def ask_question_loop
     end
 
     def activate
-      if remove_key(force)
-        if fetch_key?
-          fetch_key
+      if !key_exist? || force
+        success_get_new = get_new_key
+        if success_get_new
```
carbonin: This can just be:
```ruby
if get_new_key
  save_new_key
else
...
```

PR #212, opened 2017-06-21
Consider database.yml not configured if we don't have a v2_key

carbonin:
This becomes an issue in the unlikely scenario where a v2_key is not present, but there are v2 encrypted strings in database.yml

This allows the user to create a new v2_key using the console which would previously fail to even start in this situation.

Follow up to #211

https://bugzilla.redhat.com/show_bug.cgi?id=1430701


yrudman:
👍

PR #214, opened 2017-06-23
Enter password once when connecting to external (existed) DB

ailisp:
**ISSUE**: When configure database in appliance console, password needs to be enter twice, which is unnecessary.

**BZ**: https://bugzilla.redhat.com/show_bug.cgi?id=1462032

\cc @gtanzillo @yrudman 

@miq-bot add-label bug,wip

### DIFF
--- a/lib/gems/pending/appliance_console/database_configuration.rb
+++ b/lib/gems/pending/appliance_console/database_configuration.rb
@@ -114,7 +114,7 @@ module ApplianceConsole
       raise MiqSignalError if warn && !agree(CREATE_REGION_AGREE)
     end
 
-    def ask_for_database_credentials
+    def ask_for_database_credentials(password_twice = true)
       self.host     = ask_for_ip_or_hostname("database hostname or IP address", host) if host.blank? || !local?
       self.port     = ask_for_integer("port number", nil, port) unless local?
       self.database = just_ask("name of the database on #{host}", database) unless local?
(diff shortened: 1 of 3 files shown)


lib/gems/pending/appliance_console/database_configuration.rb
```diff
@@ -147,7 +138,7 @@ def friendly_inspect
 Host:     #{host}
 Username: #{username}
 Database: #{database}
-FRIENDLY
+      FRIENDLY
```
Fryguy: This isn't necessary, but doesn't hurt

ailisp: eh, this is auto reindent by rubymine :rofl:, next time I'll `git diff` before commit

lib/gems/pending/appliance_console/database_configuration.rb
```diff
@@ -114,7 +114,7 @@ def create_new_region_questions(warn = true)
       raise MiqSignalError if warn && !agree(CREATE_REGION_AGREE)
     end
 
-    def ask_for_database_credentials
+    def ask_for_database_credentials(password_once = true)
```
carbonin: I think I would rather default to the previous behavior, then change only the callers that were causing the issue before.

ailisp:
@carbonin changed, please have a look. If OK, I'll squash it into 3 commits.

lib/gems/pending/appliance_console/external_database_configuration.rb
```diff
@@ -22,7 +22,7 @@ def ask_questions
       create_new_region_questions if action == :create
       clear_screen
       say("Database Configuration\n")
-      ask_for_database_credentials
+      ask_for_database_credentials false
```
carbonin: Use parenthesis for method calls

yrudman:
Title should be changed -  we are still entering password twice when creating new DB,
may be something like "enter password once when connecting to external DB" ?

ailisp:
@carbonin OK, ready to merge

PR #215, opened 2017-06-24
can use arrow to navigate input in appliance console

ailisp:
**ISSUE**: User can't navigate input using arrow keys when answering question to appliance console

**BZ**: https://bugzilla.redhat.com/show_bug.cgi?id=1446249

\cc @gtanzillo @yrudman 

@miq-bot add-label bug,wip

### DIFF
--- a/spec/appliance_console/prompts_spec.rb
+++ b/spec/appliance_console/prompts_spec.rb
@@ -8,14 +8,14 @@ require 'highline/import'
 require "linux_admin"
 
 describe ApplianceConsole::Prompts do
-  let(:input) {
+  let(:input) do
     temp_stdin = Tempfile.new("temp_stdin")
     File.open(temp_stdin.path, 'w+')
-  }
-  let(:readline_output){
+  end
+  let(:readline_output) do
     temp_stdout = Tempfile.new("temp_stdout")
     File.open(temp_stdout.path, 'w+')
-  }
+  end
   let(:output) { StringIO.new }
   let(:prompt) { "\n?  " }
 
--- a/spec/appliance_console/prompts_spec.rb
+++ b/spec/appliance_console/prompts_spec.rb
@@ -1,4 +1,6 @@
-require "stringio"
+require 'stringio'
+require 'tempfile'
+require 'readline'
 require "appliance_console/errors"
 require "appliance_console/prompts"
 require "active_support/all"
@@ -6,12 +8,16 @@ require 'highline/import'
 require "linux_admin"
 
 describe ApplianceConsole::Prompts do
-  let(:input)  { StringIO.new }
+  let(:temp_stdin) { Tempfile.new "temp_stdin" }
+  let(:temp_stdout) { Tempfile.new "temp_stdout" }
+  let(:input) { File.open(temp_stdin.path, 'w+') }
+  let(:output_with_input) { File.open(temp_stdout.path, 'w+') }
   let(:output) { StringIO.new }
-
   let(:prompt)  { "\n?  " }
 
   subject do
+    Readline.input = input
+    Readline.output = output_with_input
     Class.new(HighLine) { include ApplianceConsole::Prompts }.new(input, output)
   end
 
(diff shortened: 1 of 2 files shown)


ailisp:
Ready for review now.

This spec change looks ugly, but I can't find another way to work around to make spec pass. The problem here is, the Highline gem we use will mixup input and output if working in readline mode (Highline.readline=true). That means it's input and output will mess up to be the output. While our spec is designed to have separate input and output StringIO. Also, It needs manually set up both `Readline.input`, `Readline.output` and `Highline.new(input,output)` to make our spec works. Fail to set one of them cause spec stuck in getting input or just put output to the stdout. And although the input is the same, the output is actually not. The `Readline.output` is a messed up version of input and output. Highline's output is output only but the first one of our `expect_heard` only appears in `Readline.output` but not Highline's output. This should be considered as a bug of Highline. In our running appliance console, input is stdin and output is stdout so it's good. So I treat first of our `expect_heard strs` separately, to find it in `Readline.output` and rest ones can be found in `Readline.output`. If not doing so, I think we have to rewrite all our prompts related tests to send one input each time, then expect the output with that input and not possible to send all input and output. I think it's not reasonable to rewrite all that tests. Another way maybe we construct all output and input then compare with the messed up `Readline.output` but I've tried this approach also have several different cases, and also looks ugly. Maybe you have other better ways for this prompt spec?

Ref:
[https://github.com/JEG2/highline/blob/master/test/test_highline.rb](https://github.com/JEG2/highline/blob/master/test/test_highline.rb)
[https://fossies.org/linux/ruby/test/readline/test_readline.rb]https://fossies.org/linux/ruby/test/readline/test_readline.rb

yrudman:
good job! I think it is time to remove 'wip'

yrudman:
looks good to me 👍

yrudman:
@miq-bot   assign @carbonin

spec/appliance_console/prompts_spec.rb
```diff
@@ -602,7 +608,17 @@ def expect_cls
...
+    readline_output = output_with_input.read
+    if readline_output.empty?
+      strs = Array(strs).collect { |s| s == "" ? "\n" : s }.join
+    else
+      strs = Array(strs)
+      str1 = strs[0]
+      strs = strs[1..-1].collect { |s| s == "" ? "\n" : s }.join
+      first_output_correct = readline_output.include?(str1)
+      expect(first_output_correct).to be_truthy
+    end
```
carbonin: I think this can all be simplified to something like this:

```ruby
readline_output = output_with_input.rewind.read

strs = Array(strs)
expect(readline_output).to_include(strs.shift) unless readline_output.empty?
strs = strs.collect { |s| s == "" ? "\n" : s }.join
```

carbonin:
Also do we care that this doesn't fix the issues for menu prompts or Y/N prompts?

spec/appliance_console/prompts_spec.rb
ailisp: Quite simplistic 👍  Make all these changes. Except that  rewind returns an integer so can't combine rewind.read together. For menu prompts or Y/N prompts I think it's acceptable since user is not likely to move cursor to change one character. What do you think?

spec/appliance_console/prompts_spec.rb
```diff
@@ -1,17 +1,27 @@
...
 require "appliance_console/prompts"
 require "active_support/all"
 require 'highline/import'
 require "linux_admin"
 
 describe ApplianceConsole::Prompts do
-  let(:input)  { StringIO.new }
+  let(:input) do
+    temp_stdin = Tempfile.new("temp_stdin")
+    File.open(temp_stdin.path, 'w+')
```
carbonin: Does this not work if we just use the `Tempfile` object directly? Why reopen it as a `File` type?

carbonin:
>  For menu prompts or Y/N prompts I think it's acceptable since user is not likely to move cursor to change one character. What do you think?

Yes, I think it's fine not to include these in the fix.

spec/appliance_console/prompts_spec.rb
ailisp: Unfortunately we can't. If so, it complains 
 ```
Failure/Error: Readline.input = input
      
      TypeError:
        wrong argument type Tempfile (expected File)
```

PR #217, opened 2017-06-28
Disable connection reuse for WebSocket connections in Apache

skateman:
This is a temporary workaround for the issue described here:
https://bugzilla.redhat.com/show_bug.cgi?id=1404354

This can be reverted after httpd is updated to 2.4.25 or newer

cc @psav, @gtanzillo

### DIFF
--- a/lib/gems/pending/util/miq_apache/miq_apache.rb
+++ b/lib/gems/pending/util/miq_apache/miq_apache.rb
@@ -191,7 +191,13 @@ module MiqApache
 
       ports = Array(ports).sort.reverse
       ports.each do |port|
-        @raw_lines.insert(index + 1, "BalancerMember #{protocol}://0.0.0.0:#{port}\n")
+        # Temporarily disable the connection reuse for WebSockets for httpd version < 2.4.25
+	# https://bugzilla.redhat.com/show_bug.cgi?id=1404354
+        if protocol == 'ws'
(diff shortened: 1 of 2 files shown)


lib/gems/pending/util/miq_apache/miq_apache.rb
```diff
@@ -191,15 +191,20 @@ def add_ports(ports, protocol)
 
       ports = Array(ports).sort.reverse
       ports.each do |port|
-        @raw_lines.insert(index + 1, "BalancerMember #{protocol}://0.0.0.0:#{port}\n")
+        # Temporarily disable the connection reuse for WebSockets for httpd version < 2.4.25
```
jrafanie: Can we link to a BZ so we know when this is fixed and released?

jrafanie: (if they backport and fix it before 2.4.25 for example...)

skateman: @jrafanie done

carbonin:
This change needs to live in this file https://github.com/ManageIQ/manageiq-appliance/blob/master/COPY/etc/httpd/conf.d/manageiq-balancer-websocket.conf.

We don't write these dynamically anymore.

jrafanie:
Good find @carbonin, the final caller of `add_ports` was removed [here](https://github.com/ManageIQ/manageiq/commit/955d03fe46c6804119cbc49a0c52a3d6dc1fdda3)

skateman:
Okay, closing this and creating a new PR there.

PR #222, opened 2017-07-05
handle PG::ConnectionBad and appliance console won't crash when provide invalid IP address

ailisp:
**ISSUE**: The PG::ConnectionBad error is not handled, so if an invalid IP address is provided in database replication, standby, appliance console will crashed. Now it's fixed.
**BZ**: https://bugzilla.redhat.com/show_bug.cgi?id=1445819

\cc @yrudman 

@miq-bot add-label wip, bug

### DIFF
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -136,6 +136,12 @@ module ApplianceConsole
 
       say("An #{node_state} #{rec["type"]} node (#{rec["name"]}) with the node number #{node_number} already exists")
       ask_yn?("Would you like to continue configuration by overwriting the existing node", "N")
+
+    rescue PG::Error => e
+      error_msg = "Failed to validate node number #{node_number}. #{e.message}"
+      say(error_msg)
+      Logging.logger.error(error_msg)
+      return false
     end
 
     private
--- a/spec/appliance_console/database_replication_standby_spec.rb
+++ b/spec/appliance_console/database_replication_standby_spec.rb
@@ -306,27 +306,42 @@ describe ApplianceConsole::DatabaseReplicationStandby do
       double("nil PG::Result", :map_types! => mapped_result)
     end
 
-    before do
-      expect(PG::Connection).to receive(:new).and_return(connection)
-      expect(PG::BasicTypeMapForResults).to receive(:new).and_return(type_map)
-      subject.node_number = node_number
+    context "connection successful and node_number got" do
+      before do
+        expect(PG::Connection).to receive(:new).and_return(connection)
+        expect(PG::BasicTypeMapForResults).to receive(:new).and_return(type_map)
+        subject.node_number = node_number
+      end
+
+      it "returns true if no node is found" do
+        expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(nil_result)
+        expect(subject.node_number_valid?).to be_truthy
+      end
+
+      it "returns true if a node is found and overwrite is confirmed" do
+        expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(one_result)
+        expect(subject).to receive(:ask_yn?).and_return(true)
+        expect(subject.node_number_valid?).to be_truthy
+      end
+
+      it "returns false if a node is found and overwrite is not confirmed" do
+        expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(one_result)
+        expect(subject).to receive(:ask_yn?).and_return(false)
+        expect(subject.node_number_valid?).to be_falsey
+      end
     end
 
-    it "returns true if no node is found" do
-      expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(nil_result)
-      expect(subject.node_number_valid?).to be_truthy
-    end
+    context "unable to get node_number" do
+      it "returns false when the connection fails" do
+        expect(PG::Connection).to receive(:new).and_raise(PG::ConnectionBad)
+        expect(subject.node_number_valid?).to be_falsey
+      end
 
-    it "returns true if a node is found and overwrite is confirmed" do
-      expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(one_result)
-      expect(subject).to receive(:ask_yn?).and_return(true)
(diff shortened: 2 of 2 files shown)


lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -146,6 +147,11 @@ def record_for_node_number
         SELECT type, name, active
         FROM repl_nodes where id = $1
       SQL
+    rescue PG::ConnectionBad => e
+      error_msg = "Failed to record for node number #{node_number}. #{e.message}"
```
carbonin: I think we can do better with this message, maybe:

```ruby
"Failed to connect to primary database to check node number: #{e.message}"
```

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -130,6 +130,7 @@ def stop_repmgrd
 
     def node_number_valid?
       rec = record_for_node_number
+      return false if rec == false
```
carbonin: This seems a bit odd especially because we're returning `true` if rec is `nil`. I think we're overloading the return value of `record_for_node_number` too much here.

Maybe it would be better to rescue the exception in this method. What do you think?

carbonin:
Also, is this still WIP? Are you going to write a spec for this?

lib/gems/pending/appliance_console/database_replication_standby.rb
ailisp: Good idea. I also think it's weird: nil is correct, false is error.

ailisp:
@yrudman @carbonin Do you think we need a new spec for this?

yrudman:
I think you should add test to cover your changes (```say(error_msg)```)

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -129,7 +129,14 @@ def stop_repmgrd
     end
 
     def node_number_valid?
-      rec = record_for_node_number
+      begin
+        rec = record_for_node_number
+      rescue PG::Error => e
```
carbonin: Nothing else in this method will raise a PG error. I would rather still have this at the bottom of the method.

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -129,7 +129,14 @@ def stop_repmgrd
...
+        error_msg = "Failed to record for node number #{node_number}. #{e.message}"
```
carbonin: My previous comment about this message still applies:

>I think we can do better with this message, maybe:
>```ruby
>"Failed to connect to primary database to check node number: #{e.message}"
>```

spec/appliance_console/database_replication_standby_spec.rb
```diff
@@ -306,27 +306,44 @@
...
     end
 
-    it "returns false if a node is found and overwrite is not confirmed" do
-      expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(one_result)
-      expect(subject).to receive(:ask_yn?).and_return(false)
-      expect(subject.node_number_valid?).to be_falsey
+    context "unable to get node_number" do
+      it "error to connect database" do
+        expect(PG::Connection).to receive(:new).and_raise(PG::ConnectionBad)
+        expect(subject).to receive(:say).with(/Failed to record for node number/)
```
carbonin: Not a fan of expecting on error messages, I think the return value is what we are really testing here.

spec/appliance_console/database_replication_standby_spec.rb
```diff
@@ -306,27 +306,44 @@
...
+        expect(subject.node_number_valid?).to be_falsey
+      end
     end
 
-    it "returns false if a node is found and overwrite is not confirmed" do
-      expect(connection).to receive(:exec_params).with(instance_of(String), [1]).and_return(one_result)
-      expect(subject).to receive(:ask_yn?).and_return(false)
-      expect(subject.node_number_valid?).to be_falsey
+    context "unable to get node_number" do
+      it "error to connect database" do
```
carbonin: These specs should describe expected behavior so this should probably be something like:

```ruby
it "returns false when the connection fails" do
```

spec/appliance_console/database_replication_standby_spec.rb
```diff
@@ -306,27 +306,44 @@
...
+        expect(PG::Connection).to receive(:new).and_raise(PG::ConnectionBad)
+        expect(subject).to receive(:say).with(/Failed to record for node number/)
+        expect(subject.node_number_valid?).to be_falsey
+      end
+
+      it "error to query from repl_nodes" do
```
carbonin: Same here:

```ruby
it "returns false when the query fails" do
```

lib/gems/pending/appliance_console/database_replication_standby.rb
ailisp: All agree but have doubt on this one. Error can be error on connection or execute the `SELECT ... FROM repl_nodes` in `record_for_node_number`. It's a `PG::UndefinedTable` Error in my vm. Do we still use `Failed to connect`?

lib/gems/pending/appliance_console/database_replication_standby.rb
```diff
@@ -136,6 +136,12 @@ def node_number_valid?
 
       say("An #{node_state} #{rec["type"]} node (#{rec["name"]}) with the node number #{node_number} already exists")
       ask_yn?("Would you like to continue configuration by overwriting the existing node", "N")
+
+    rescue PG::Error => e
+      error_msg = "Failed to record for node number #{node_number}. #{e.message}"
```
carbonin: `Failed to validate node number #{node_number}`

ailisp:
@miq-bot remove-label wip

simaishi:
Fine backport details:

```
$ git log -1
commit 0084c06085c1a303ff1938a779bc5c11fbce8078
Date:   Thu Jul 6 18:13:06 2017 -0400

    Merge pull request #222 from ailisp/bug-1445819
    
    handle PG::ConnectionBad and appliance console won't crash when provide invalid IP address
    (cherry picked from commit 15f1b740dfd4517dda5e286d914da835326e0af8)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1496900
```

PR #232, opened 2017-07-11
Remove the start and stop methods from PostgresAdmin

carbonin:
They are no longer used after https://github.com/ManageIQ/manageiq/pull/15542


jrafanie:
Test failed... kicking it after https://github.com/ManageIQ/manageiq/pull/15542 was merged

PR #239, opened 2017-07-14
Remove Win32API

djberg96:
Since we don't support running the app on Windows we can remove Win32API and cleanup these OS checks.

Followup to https://github.com/ManageIQ/manageiq/issues/1404

### DIFF
--- a/lib/gems/pending/util/extensions/miq-file.rb
+++ b/lib/gems/pending/util/extensions/miq-file.rb
@@ -1,5 +1,4 @@
 require 'sys-uname'
-require 'Win32API' if Sys::Platform::OS == :windows
 require 'util/runcmd'
 require 'uri'
 require 'util/MiqSockUtil'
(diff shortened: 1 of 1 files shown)


lib/gems/pending/util/extensions/miq-file.rb
```diff
@@ -1,32 +1,16 @@
...
 require 'uri'
 require 'util/MiqSockUtil'
 require 'addressable'
 
 class File
   def self.paths_equal?(f1, f2)
-    if Sys::Platform::OS == :windows
-      # Note: The file needs to exist and be accessable for getShortFileName to work.
-      f1 = File.getShortFileName(f1).downcase
-      f2 = File.getShortFileName(f2).downcase
```
Fryguy: The only callers of `getShortFileName` across the code base are these lines here, so we can remove `getShortFileName` entirely now.

Fryguy:
I'm going to merge, and follow up with the other deletions. I believe paths_equal can also be deleted.

PR #242, opened 2017-07-17
Extract hash sorting to more_core_extensions

bdunne:
https://github.com/ManageIQ/more_core_extensions/pull/46


Fryguy:
Merged but needs a gem release.

PR #244, opened 2017-07-18
Don't clear the cache until we have a new value to cache

jrafanie:
The prior cache[:value] was being cleared too early.   If block.call
raised an exception, we lost the prior cached value.

A subsequent method call was checking only if the cache was non-empty, which it was,
since the cache[:timeout] was set.  We would then return a nil because cache[:value]
was never set when the exception occurred.

https://bugzilla.redhat.com/show_bug.cgi?id=1469307
https://bugzilla.redhat.com/show_bug.cgi?id=1468898

### DIFF
--- a/lib/gems/pending/util/extensions/miq-module.rb
+++ b/lib/gems/pending/util/extensions/miq-module.rb
@@ -26,15 +26,17 @@ class Module
         cache = $miq_cache_with_timeout[key]
 
         old_timeout = cache[:timeout]
-        cache.clear if force_reload || (old_timeout && Time.now.utc > old_timeout)
-        break cache[:value] unless cache.empty?
+        timeout_expired = !!old_timeout && (Time.now.utc > old_timeout)
+        break cache[:value] if (!force_reload && !timeout_expired && cache.key?(:value))
 
         new_timeout = timeout || 300 # 5 minutes
         new_timeout = new_timeout.call if new_timeout.kind_of?(Proc)
         new_timeout = Time.now.utc + new_timeout
+        new_value   = block.call
+        cache.clear
 
         cache[:timeout] = new_timeout
-        cache[:value]   = block.call
+        cache[:value]   = new_value
       end
     end
 
--- a/spec/util/extensions/miq-module_spec.rb
+++ b/spec/util/extensions/miq-module_spec.rb
@@ -41,6 +41,22 @@ describe Module do
       expect(test_module.default).to eq(value)
     end
 
+    it 'will return a prior cached value if resetting the value fails' do
+      $already_run = false
+      test_class.cache_with_timeout(:flapping_method) do
+        if $already_run == false
+          $already_run = true
+          5
+        else
+          raise "OOPS"
+        end
+      end
+
+      test_class.flapping_method(true)             # sets cache to 5
+      test_class.flapping_method(true) rescue nil  # blows up, doesn't reset cache
+      expect(test_class.flapping_method).to eq(5)  # returns prior cached value of 5
+    end
+
     it 'will not return the cached value when passed force_reload = true' do
       value = test_class.default(true)
       expect(test_class.default(true)).not_to eq(value)


jrafanie:
Please throw 🍅

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,22 @@
       expect(test_module.default).to eq(value)
     end
 
+    it 'will return a prior cached value if resetting the value fails' do
+      $already_run = false
```
jrafanie: I hate this test but it recreates it... any ideas how to do this without a global variable?

lib/gems/pending/util/extensions/miq-module.rb
```diff
@@ -26,15 +26,17 @@ def cache_with_timeout(method, timeout = nil, &block)
         cache = $miq_cache_with_timeout[key]
 
         old_timeout = cache[:timeout]
-        cache.clear if force_reload || (old_timeout && Time.now.utc > old_timeout)
-        break cache[:value] unless cache.empty?
+        timeout_expired = !!old_timeout && (Time.now.utc > old_timeout)
+        break cache[:value] if (!force_reload && !timeout_expired && cache.key?(:value))
```
Fryguy: If a request is made to clear the cache we should clear it regardless, so we should keep the force_reload in place.  However, the cache.key?(:value) part should stay.

lib/gems/pending/util/extensions/miq-module.rb
```diff
@@ -26,15 +26,17 @@ def cache_with_timeout(method, timeout = nil, &block)
...
 
         new_timeout = timeout || 300 # 5 minutes
         new_timeout = new_timeout.call if new_timeout.kind_of?(Proc)
         new_timeout = Time.now.utc + new_timeout
+        new_value   = block.call
+        cache.clear
```
Fryguy: You don't need the cache.clear here, because it was already done above.

Fryguy:
I am 👎  on changing when the cache is cleared.  IMO the design is:

1. blow away the cache if time has passed or asked to via force_reload
2. refill the cache based on the block

2 raising an exception should having nothing to do with 1 as they are discrete steps.

The problem is not that we clear the cache before we have something to replace it with, but that we filled the cache with a `:timeout` key when the `:cache` key may not exist.  That's an ordering issue that is fixed by looking for the presence of the `:cache` key specifically instead of "any key".

Fryguy:
Ah...I just saw the name of the test and now I see the disconnect:

    it 'will return a prior cached value if resetting the value fails' do

I disagree.  I think that it should *NOT* return a prior cached value as that could put us in a state where the cached value is cached forever, defeating the purpose of the method.  When time is up (or forced), thats it.  If the block fails after that we just re-raise the exception.  The bug is that we effectively cached a "nil" on an exception, which is very bad.

lib/gems/pending/util/extensions/miq-module.rb
jrafanie: Yeah, I think this code is doing exactly what you describe...

Basically, we only return from the cache:
If it's not force reload, and, the timeout hasn't expired, and the cache has the value key.

jrafanie: Note, we don't want to clear it regardless before checking for the new value... it will cache a nil if the block.call raises an exception.

lib/gems/pending/util/extensions/miq-module.rb
jrafanie: I problem was that we were clearing the cache before checking for a new value to cache, therefore this clear was moved down here, after we get the `new_timeout` and `new_value` and before we set these values into the cache.

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,22 @@
...
+      test_class.cache_with_timeout(:flapping_method) do
+        if $already_run == false
+          $already_run = true
+          5
+        else
+          raise "OOPS"
```
jrafanie: Note, there are two scenarios I want to test but don't know how to properly write.
1) a method returns in the follow pattern:
  * 5
  * raises
  * raises
  * raises
  * ...

In this case, which is described currently in the test, we want to return the previously cached value of 5.  It makes no sense to clear the cache if we can't store a new value in the cache. We'd cache a nil.

2) a method returns in the follow pattern:
  * 5
  * raises
  * 5
  * raises
  * 5

In this case, we want to return only the cached value of 5 and never cache the nil.

jrafanie:
> I am 👎 on changing when the cache is cleared. IMO the design is:
> 
> blow away the cache if time has passed or asked to via force_reload
> refill the cache based on the block
> 2 raising an exception should having nothing to do with 1 as they are discrete steps.

@Fryguy it's not clear which edge cases are affected here.  My thought was `MiqServer.my_server(true)` should be atomic... change the cached value to the "current" value if that operation succeeds.  I can see now based on your comments above that this is slightly different behavior.  Let me see if I can draft some tests that describe the  edge cases and see what makes sense there.  Cache invalidation is 🍰 .

jrafanie:
Closing this PR since I'm taking @Fryguy's side... even if `MiqServer.my_server(true)` should be an atomic operation that does the "cache clearing" and "cache setting", that's new behavior and shouldn't be changed.  I'll open a subsequent PR with a different approach to just fix the problem.

spec/util/extensions/miq-module_spec.rb
Fryguy: > It makes no sense to clear the cache if we can't store a new value in the cache. We'd cache a nil.

No, the cache would not exist (it would be no different that the value before the very first time calling the method).

PR #246, opened 2017-07-19
Extract Range extension to more_core_extensions

bdunne:
Depends on https://github.com/ManageIQ/more_core_extensions/pull/47 and a new gem release


Fryguy:
Merged  ManageIQ/more_core_extensions#47  and waiting on a new gem release.

PR #247, opened 2017-07-19
Only return the cached value if it exists.

jrafanie:
If the cache was cleared and an exception was raised when trying to set
the new `cache[:value]`, we'd end up setting the `cache[:timeout]` but
not the new value.  Any subsequent call would return a nil from
`cache[:value]` because `cache[:timeout]` exists, making `cache` non-empty.

The new code only returns from the cache if the :value key exists in the
`cache`.

Additionally, for consistency, we only set the `cache[:timeout]`
and `cache[:value]` after we've calculated these values, therefore the
`cache` doesn't get partially set.  This is not required but makes me
feel better.

https://bugzilla.redhat.com/show_bug.cgi?id=1469307
https://bugzilla.redhat.com/show_bug.cgi?id=1468898

This replaces https://github.com/ManageIQ/manageiq-gems-pending/pull/244

### DIFF
--- a/spec/util/extensions/miq-module_spec.rb
+++ b/spec/util/extensions/miq-module_spec.rb
@@ -41,6 +41,33 @@ describe Module do
       expect(test_module.default).to eq(value)
     end
 
+    it 'clears the cache even if the block raises an exception' do
+      test_class.define_singleton_method(:call_count)  { @call_count ||= 0 }
+      test_class.define_singleton_method(:call_count=) { |val| @call_count = val }
+      test_class.cache_with_timeout(:flapping_method) do
+        begin
+          if test_class.call_count.even?
+            test_class.call_count
+          else
+            raise "OOPS"
+          end
+        ensure
+          test_class.call_count += 1
+        end
+      end
+
+      expect(test_class.flapping_method(true)).to eq(0) # initializes cache to 0
+      expect(test_class.flapping_method).to eq(0)       # returns from cache
+
+      test_class.flapping_method(true) rescue nil       # blows up, clears cache for next request
+      expect(test_class.flapping_method).to eq(2)       # sets new value since cache is cleared
+      expect(test_class.flapping_method).to eq(2)       # returns from cache
+
+      test_class.flapping_method(true) rescue nil       # blows up, clears cache for next request
+      expect(test_class.flapping_method).to eq(4)       # sets new value since cache is cleared
+      expect(test_class.flapping_method).to eq(4)       # returns from cache
+    end
+
     it 'will not return the cached value when passed force_reload = true' do
       value = test_class.default(true)
       expect(test_class.default(true)).not_to eq(value)
--- a/lib/gems/pending/util/extensions/miq-module.rb
+++ b/lib/gems/pending/util/extensions/miq-module.rb
@@ -27,14 +27,15 @@ class Module
 
         old_timeout = cache[:timeout]
         cache.clear if force_reload || (old_timeout && Time.now.utc > old_timeout)
-        break cache[:value] unless cache.empty?
+        break cache[:value] if cache.key?(:value)
 
         new_timeout = timeout || 300 # 5 minutes
         new_timeout = new_timeout.call if new_timeout.kind_of?(Proc)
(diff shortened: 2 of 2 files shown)


spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,30 @@
       expect(test_module.default).to eq(value)
     end
 
+    it 'clears the cache even if the block raises an exception' do
+      $call_count = 0
```
Fryguy: Maybe you can use an instance variable in the test class?

```ruby
def test_class.call_count
  @call_count ||= 0
end

test_class.cache_with_timeout...
```

Fryguy:
In the other PR you mentioned a test for 5, raise, raise, etc.  Do you think you still need that?

Fryguy:
:+1: on the code...just need to clean up the specs.

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,30 @@
...
+            $call_count
+          else
+            raise "OOPS"
+          end
+        ensure
+          $call_count += 1
+        end
+      end
+
+      test_class.flapping_method(true)             # initializes cache to 0
```
Fryguy: Add an expectation here as well that it's 0

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,30 @@
...
+      expect(test_class.flapping_method).to eq(0)  # returns cached 0
+
+      test_class.flapping_method(true) rescue nil  # blows up, clears cache for next request
+      expect(test_class.flapping_method).to eq(2)  # sets new value since cache is cleared
```
Fryguy: You might want to throw another call after this one to show that it's cached since technically this line is a fresh call.

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,30 @@
...
+
+      test_class.flapping_method(true) rescue nil  # blows up, clears cache for next request
+      expect(test_class.flapping_method).to eq(4)  # sets new value since cache is cleared
```
Fryguy: Same here...you might want a second call to verify the caching.

jrafanie: haha, I contemplated these too...

spec/util/extensions/miq-module_spec.rb
```diff
@@ -41,6 +41,33 @@
...
+        ensure
+          test_class.call_count += 1
+        end
+      end
+
+      expect(test_class.flapping_method(true)).to eq(0) # initializes cache to 0
+      expect(test_class.flapping_method).to eq(0)       # returns from cache
+
+      test_class.flapping_method(true) rescue nil       # blows up, clears cache for next request
+      expect(test_class.flapping_method).to eq(2)       # sets new value since cache is cleared
```
jrafanie: > In the other PR you mentioned a test for 5, raise, raise, etc. Do you think you still need that?

@Fryguy No.  This expectation shows that we are actually setting a new value and returning it, NOT the previously cached value, 0.  So, I think this test encompasses the desired test cases.

simaishi:
Fine backport details:

```
$ git log -1
commit 8f7d63f441d54ff41247c7605e2bf5986133f9f1
Date:   Fri Jul 21 12:06:41 2017 -0400

    Merge pull request #247 from jrafanie/cache_invalidation_is_harder
    
    Only return the cached value if it exists.
    (cherry picked from commit 6bdd64449485857526c968a119d55a7edba6e015)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1473787
```

simaishi:
@jrafanie The BZs referenced here is for Embedded Ansible. Since that's not available in Euwe branch, I assume this bug will show up in some different way... Can you create a BZ that describes the problem in Euwe?

jrafanie:
@simaishiThanks!  I'll remove the euwe label.  While it's possible for this bug on euwe, it's very unlikely without the embedded ansible thread worker.

PR #248, opened 2017-07-19
Extract Symbol#to_i to more_core_extensions

bdunne:
Depends on https://github.com/ManageIQ/more_core_extensions/pull/49 and a new gem version


Fryguy:
Merged ManageIQ/more_core_extensions#49 , so waiting on a new gem version.

PR #139, opened 2017-04-25
Use pid + thread id to index into the benchmark results

jrafanie:
https://bugzilla.redhat.com/show_bug.cgi?id=1424716

Forks inherit the `@@realtime_by_tid` and parent/child `Thread.current.object_id`
are equal, so we need to index into the hash with the pid too.

The following test recreates and verifies this change but because it
forks and doesn't properly bubble up exit codes, I wrote it just to
confirm the problem.

```ruby
  it "doesn't get parent's realtime_block in forked child" do
    p1_result, p1_timings = Benchmark.realtime_block(:parent_task1) do
      Timecop.travel(1000)
      p2_result, p2_timings = Benchmark.realtime_block(:parent_task2) do
        Timecop.travel(500)
      end

      expect(p2_timings[:parent_task2]).to be_within(0.5).of(500)
      expect(p2_timings.keys).to eq([:parent_task2])

      pid = fork do
        c_result, c_timings = Benchmark.realtime_block(:child_task) do
          Timecop.travel(250)
        end
        expect(c_timings[:child_task]).to be_within(0.5).of(250)
        expect(c_timings.keys).to eq([:child_task])
      end
    end

    expect(p1_timings[:parent_task1]).to be_within(0.5).of(1500)
    expect(p1_timings.keys.sort).to eq([:parent_task1, :parent_task2])
  end
```

The tests pass because I'm not handling exit codes properly, but I
see the expectation error pop up when I roll back my code change:

```
/Users/joerafaniello/.gem/ruby/2.3.4/gems/rspec-support-3.5.0/lib/rspec/support.rb:87:in
`block in <module:Support>':
(RSpec::Expectations::ExpectationNotMetError)
expected: [:child_task]
     got: [:parent_task2, :child_task]

     (compared using ==)

     Diff:
     @@ -1,2 +1,2 @@
     -[:child_task]
     +[:parent_task2, :child_task]
```

PR #255, opened 2017-08-07
realtime_block: log timings on exception in outermost block

cben:
When realtime_block is aborted with exception, the timings are still there in the hash, so a higher level may log them.
Unless that is this was the outermost block, in which case they're lost.
This PR logs them in such case.

Example:
```
[----] E, [2017-08-07T15:04:08.145415 #14296:2aaca3b530cc] ERROR -- : MIQ(ManageIQ::Providers::Openshift::ContainerManager::Refresher#refresh) EMS: [scale-ocp-36-c03], id: [3] Refresh failed
[----] E, [2017-08-07T15:04:08.145736 #14296:2aaca3b530cc] ERROR -- : [KubeException]: Timed out connecting to server  Method:[block in method_missing]
[----] E, [2017-08-07T15:04:08.145855 #14296:2aaca3b530cc] ERROR -- : /home/bpaskinc/myenv/rbenv/versions/2.3.4/lib/ruby/gems/2.3.0/gems/kubeclient-2.4.0/lib/kubeclient/common.rb:117:in `rescue in handle_exception'
...
bin/rails:4:in `<main>'
[----] E, [2017-08-07T15:04:08.145951 #14296:2aaca3b530cc] ERROR -- : MIQ(ManageIQ::Providers::Openshift::ContainerManager::Refresher#refresh) EMS: [scale-ocp-36-c03], id: [3] Unable to perform refresh for the following targets:
[----] E, [2017-08-07T15:04:08.146105 #14296:2aaca3b530cc] ERROR -- : MIQ(ManageIQ::Providers::Openshift::ContainerManager::Refresher#refresh)  --- ManageIQ::Providers::Openshift::ContainerManager [scale-ocp-36-c03] id [3]
[----] I, [2017-08-07T15:04:08.236129 #14296:2aaca3b530cc]  INFO -- : MIQ(ManageIQ::Providers::Openshift::ContainerManager::Refresher#refresh) Refreshing all targets...Complete
[----] I, [2017-08-07T15:04:08.236201 #14296:2aaca3b530cc]  INFO -- : Exception in realtime_block :ems_total_refresh - Timings: {:collect_inventory_for_targets=>60.113999128341675, :ems_refresh=>60.11402702331543, :ems_total_refresh=>60.208136320114136}
```

* This is not water-tight against `Timeout.timeout`.
  Timeout inside the block logged:
  ```
  Benchmark.realtime_block(:sleep) { Timeout.timeout(3) { sleep(10) } }
  ```
  but outside it didn't:
  ```
  Timeout.timeout(3) { Benchmark.realtime_block(:sleep) { sleep(10) } }
  ```
  (See also https://github.com/ManageIQ/manageiq/issues/9642 for more Timeout woes)

### DIFF
--- a/lib/gems/pending/util/extensions/miq-benchmark.rb
+++ b/lib/gems/pending/util/extensions/miq-benchmark.rb
@@ -34,6 +34,10 @@ module Benchmark
         begin
           ret = realtime_store(hash, key, &block)
           return ret, hash
+        rescue Exception
+          # Don't let timings be lost on exception when there is nobody else to pick them up.
+          Logger.new($stderr).info("Exception in realtime_block #{key.inspect} - Timings: #{hash.inspect}")
+          raise
         ensure
           delete_current_realtime
         end
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -50,6 +50,6 @@ Gem::Specification.new do |s|
   s.add_development_dependency "rspec",                     "~> 3.5.0"
   s.add_development_dependency "simplecov"
   s.add_development_dependency "test-unit"
-  s.add_development_dependency "timecop",                   "~> 0.7.3"
+  s.add_development_dependency "timecop",                   "~> 0.9.1"
   s.add_development_dependency "xml-simple",                "~> 1.1.0"
 end
--- a/spec/util/extensions/miq-benchmark_spec.rb
+++ b/spec/util/extensions/miq-benchmark_spec.rb
@@ -64,6 +64,43 @@ describe Benchmark do
     expect(Benchmark.in_realtime_block?).to be_falsey
   end
 
+  it '.realtime_block with an Exception aborting outermost block' do
+    expect do
+      Benchmark.realtime_block(:test1) do
+        Timecop.travel(2.1)
+        Benchmark.realtime_block(:test2) do
+          Timecop.travel(5.1)
+          raise Exception
+        end
+      end
+    end.to raise_exception(Exception)
+      .and output(/Exception in realtime_block :test1 - Timings: {:test2=>5\.\d*, :test1=>7\.\d*}/).to_stderr
+
+    expect(Benchmark.in_realtime_block?).to be_falsey
+  end
+
+  it '.realtime_block with an Exception caught in inner block' do
+    result = timings = nil
+    expect do
+      result, timings = Benchmark.realtime_block(:test1) do
+        begin
+          Timecop.travel(500)
+          Benchmark.realtime_block(:test2) do
+            Timecop.travel(500)
+            raise Exception
+          end
+        rescue Exception
+          "value"
+        end
+      end
+    end.to_not output.to_stderr
+
+    expect(result).to eq("value")
+    expect(timings[:test1]).to be_within(0.5).of(1000)
+    expect(timings[:test2]).to be_within(0.5).of(500)
+    expect(Benchmark.in_realtime_block?).to be_falsey
+  end
+
   it "Timeout raising within .realtime_block" do
     expect(Benchmark.in_realtime_block?).to be_falsey
 


cben:
@moolitayer asked if this needs an issue / BZ.
Don't think so, "good logging" is not a well defined feature we promise users :-)
It's when other issues happen and we debug from logs that this might help...

- https://bugzilla.redhat.com/show_bug.cgi?id=1440950 is one example of "normal" exception where this could have improved logs.
- https://bugzilla.redhat.com/show_bug.cgi?id=1428570 is one example of `Timeout` killing the thread, where PR probably won't help.

cben:
@jrafanie @Fryguy @agrare please review.

moolitayer:
@cben sorry my comment was not clear, I don't think we need a general logging issue. (see https://github.com/ManageIQ/manageiq-providers-kubernetes/pull/82#issuecomment-320956526)

lib/gems/pending/util/extensions/miq-benchmark.rb
```diff
@@ -29,6 +29,10 @@ def self.realtime_block(key, &block)
     begin
       ret = realtime_store(hash, key, &block)
       return ret, hash
+    rescue
+      # Don't let timings be lost on exception when there is nobody else to pick them up.
+      $log.info("Exception in realtime_block #{key.inspect} - Timings: #{hash.inspect}") if outermost
```
jrafanie: There are no other log lines in this file.  It probably shouldn't expect `$log` to be defined and instead be initialized with a logger... But, for a surgical change, maybe add `if $log && outermost`

jrafanie:
Tangentially related:  https://github.com/ManageIQ/manageiq-gems-pending/pull/139

lib/gems/pending/util/extensions/miq-benchmark.rb
Fryguy: The Benchmark extension is meant to be non-miq-specific.  In fact, it's on the list to be moved to more_core_extensions.  Imagine it's in more_core, how would this change work?

cben: I'm clueless about how all the `$log`,  `_log` thingies come to be (storks?), please advise...

The OOP answer is not use Benchmark class methods but an object intialized with a logger.  meh — much hassle, and pushes similar problem onto the users that would want a global instance...

- `realtime_block` could take a logger option, but again too much hassle to pass it everywhere, especially for an edge case.
- could have some accessor `Benchmark.realtime_block_logger = ...` to be initialized once in apps that have logging.  (need advice where to set in MIQ?)
- just default to logging to stderr, might be good enough for a rare edge case?

I also wanted to propose a convenience `realtime_log_once` function that logs timing IFF it's the outermost, and only returns the result, so callers don't need to discard `result, _ = ...`.
Same considerations, except it's standard flow, not an edge case.

jrafanie: @cben I agree.  

* Creating an instance that you pass a logger seems like a hassle.

* Using a class accessor on Benchmark to set a logger might work but then the caller knows about benchmark... I guess.  It might work?  See [here](https://github.com/ManageIQ/manageiq/blob/aad44397f41df5a2dbb1b4691307887518b3c2f8/lib/vmdb/loggers.rb#L63) where we create the loggers and setup the external loggers.

* stderr should work too.  Try that first and see if the workers log it.  I think that's the simplest approach.  We should be keeping stderr.

```
app/models/miq_worker.rb:    pid = Kernel.spawn(self.class.build_command_line(guid), [:out, :err] => [Rails.root.join("log", "evm.log"), "a"])
lib/tasks/evm_application.rb:    pid = Kernel.spawn(env_options, command_line, :pgroup => true, [:out, :err] => [Rails.root.join("log/evm.log"), "a"])
```

Fryguy: The only other thing I can think of is a top-level accessor that has the "last" value's run... Something like a last_realtime_block_result.  Downside is you would have this thing hanging around forever, but it isn't very big.

Fryguy: > Try that first and see if the workers log it.

Again, you have to put on your non-manageiq hat, because that's where the class is going.  This makes no sense if some random project uses it.

jrafanie: It makes perfect sense for a library to write to stderr on errors if they don't provide a facility for specifying a logger.

chessbyte:
@jrafanie @Fryguy @agrare please review.

PR #256, opened 2017-08-08
Fix restart network error when set host name in pure ipv6 network

ailisp:
ISSUE: In pure ipv6 network, after set hostname, appliance console will restart network. Currently we're using `systemctl restart network`, which can't successfully start network after stop it and cause appliance console crushed. But using NetworkManager is fine. As in https://access.redhat.com/solutions/783533 , network is considered as depreicate and we can update to use NetworkManager instead

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1475804

\cc @gtanzillo @yrudman 

@miq-bot add-label wip, bug

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -307,7 +307,7 @@ Static Network Configuration
             system_hosts.hostname = new_host
             system_hosts.set_loopback_hostname(new_host)
             system_hosts.save
-            LinuxAdmin::Service.new("network").restart
+
             press_any_key
           end
         end


yrudman:
@bdunne could you review

ailisp:
I'm not sure it's a good.way to fix it. Will remove wip.

ailisp:
@miq-bot remove-label wip

ailisp:
@carbonin Can we merge this? Thanks.

carbonin:
I don't think we have `NetworkManager` installed on the appliance. I remember doing some things to explicitly disable it so that we use the `network` service directly.

Can you pin down *why* the restart is failing when we are not using `NetworkManager`?

ailisp:
@carbonin I've find the error log for `systemctl network restart`. Let me find it. For `NetworkManager`, I think it's installed because I've tried in that appliance.

lib/gems/pending/appliance_console.rb
```diff
@@ -307,7 +307,7 @@ module ApplianceConsole
             system_hosts.hostname = new_host
             system_hosts.set_loopback_hostname(new_host)
             system_hosts.save
-            LinuxAdmin::Service.new("network").restart
+            LinuxAdmin::Service.new("NetworkManager").restart
```
bdunne: Honestly, I'm not sure why this restart is here.  Can we just remove it?

ailisp: In my search, some documents said we need to restart the machine, while some docs require to at least restart network service to make hostname change take effect: https://www.howtogeek.com/50631/how-to-change-your-linux-hostname-without-rebooting/

bdunne: The hostnamectl command which we're running seems to take care of everything and not require a reboot.
https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/Networking_Guide/sec_Configuring_Host_Names_Using_hostnamectl.html

ailisp: @bdunne Thanks. `hostnamectl set-hostname name` works. And it's much faster than `systemctl restart network`. But the bug disappear today, `systemctl restart network` works and set hostname in ap won't crash. Do we still want to change that to `hostnamectl`? @carbonin

carbonin: @ailisp We are already using `hostnamectl` in [`LinuxAdmin::Hosts#hostname=`](https://github.com/ManageIQ/linux_admin/blob/master/lib/linux_admin/hosts.rb#L39). That's why @bdunne suggested just getting rid of this line.

simaishi:
Fine backport details:

```
$ git log -1
commit 0fb020cfec87b51eec36ebc82a8c8d0e2431fa7c
Date:   Thu Aug 17 09:38:18 2017 -0400

    Merge pull request #256 from ailisp/ipv6-set-hostname
    
    Fix restart network error when set host name in pure ipv6 network
    (cherry picked from commit cbc714fa4235bfb34436e4ce7344a4200194a2e9)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1482669
```

PR #257, opened 2017-08-08
Allow httpd password to have dollar sign in it

yrudman:
Having `$` sign inside double quotation for echo argument is interpreted as variable which need to be resolved.

https://bugzilla.redhat.com/show_bug.cgi?id=1438974

@miq-bot add-label bug

\cc @bdunne

### DIFF
--- a/lib/gems/pending/appliance_console/external_httpd_authentication.rb
+++ b/lib/gems/pending/appliance_console/external_httpd_authentication.rb
@@ -128,7 +128,7 @@ module ApplianceConsole
 
     def configure_ipa_http_service
       say("Configuring IPA HTTP Service and Keytab ...")
-      AwesomeSpawn.run!("/bin/echo \"#{@password}\" | /usr/bin/kinit #{@principal}")
+      AwesomeSpawn.run!("/usr/bin/kinit", :params => [@principal], :stdin_data => @password)
       service = Principal.new(:hostname => @host, :realm => realm, :service => "HTTP", :ca_name => "ipa")
       service.register
       AwesomeSpawn.run!(IPA_GETKEYTAB, :params => {"-s" => @ipaserver, "-k" => HTTP_KEYTAB, "-p" => service.name})
--- a/spec/appliance_console/external_httpd_authentication_spec.rb
+++ b/spec/appliance_console/external_httpd_authentication_spec.rb
@@ -1,6 +1,7 @@
 require "active_support/all"
 require "appliance_console/external_httpd_authentication"
 require "appliance_console/prompts"
+require "appliance_console/principal"
 require "linux_admin"
 
 describe ApplianceConsole::ExternalHttpdAuthentication do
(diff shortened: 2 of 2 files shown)


yrudman:
@miq-bot add-label fine/yes

lib/gems/pending/appliance_console/external_httpd_authentication.rb
```diff
@@ -128,7 +128,7 @@ def configure_sssd
 
     def configure_ipa_http_service
       say("Configuring IPA HTTP Service and Keytab ...")
-      AwesomeSpawn.run!("/bin/echo \"#{@password}\" | /usr/bin/kinit #{@principal}")
+      AwesomeSpawn.run!("/bin/echo '#{@password}' | /usr/bin/kinit #{@principal}")
```
bdunne: @yrudman Is it possible to add some tests for this?

Fryguy:
Then you can't have `"` in your password either. The problem here is that awesome_spawn is being used incorrectly...This should be

```ruby
AwesomeSpawn.run!("/usr/bin/kinit", :params => [@principal], :stdin_data => @password)
```

yrudman:
thank you @Fryguy

spec/appliance_console/external_httpd_authentication_spec.rb
```diff
@@ -162,4 +163,26 @@
       subject.post_activation
     end
   end
+
+  context "#configure_ipa_http_service" do
+    before do
```
bdunne: Why use a `before` block when there is only one test?

spec/appliance_console/external_httpd_authentication_spec.rb
```diff
@@ -162,4 +163,26 @@
...
+      allow(subject).to receive(:say)
```
bdunne: Shouldn't we `expect` all of these things to happen?

spec/appliance_console/external_httpd_authentication_spec.rb
```diff
@@ -162,4 +163,26 @@
...
+      service = double("Principal")
+      allow(ApplianceConsole::Principal).to receive(:new).and_return(service)
+      allow(service).to receive(:register)
+      allow(service).to receive(:name)
+      allow(service).to receive(:name)
```
bdunne: Prefer `expect(service).to receive(:name).twice`

spec/appliance_console/external_httpd_authentication_spec.rb
yrudman: I think it looks more clean to separate what we are testing from what is not important

spec/appliance_console/external_httpd_authentication_spec.rb
yrudman: my mistake, forgot to delete that line, we need that only once

spec/appliance_console/external_httpd_authentication_spec.rb
yrudman: I was thinking that `before` block is setting-up environment for executing expectation in `it' block. I am not sure what is the right way (`expect` vs `allow`) to mock some behavior  which is not tested in `it`. 
@bdunne, do you think it make sense to remove `before` and replace all `allow` with `expect` here?

simaishi:
Fine backport details:

```
$ git log -1
commit 5be8085f865e889b33b2f6871318dcf97b0f9305
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Tue Aug 15 11:57:23 2017 -0400

    Merge pull request #257 from yrudman/allow-httpd-password-to-have-dollar-sign
    
    Allow httpd password to have dollar sign in it
    (cherry picked from commit 7bea4f3d0ebda0ef479110b7c61e42e9dbf72834)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1481846
```

PR #259, opened 2017-08-10
Bump sys-proctable

chrisarcand:
To force getting rid of the very annoying Fixnum warnings for Ruby 2.4

cc/ @djberg96


djberg96:
👍

PR #260, opened 2017-08-10
Configuring database hourly maintenance with appliance_console_cli tool

rsevilla87:
Configure database hourly maintenance with _appliance_console_cli_ tool. This feature is available using _appliance_console_, and could be useful for automated deployments.

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -13,6 +13,7 @@ require 'appliance_console/principal'
 require 'appliance_console/certificate'
 require 'appliance_console/certificate_authority'
 require 'appliance_console/logfile_configuration'
+require 'appliance_console/database_maintenance_hourly'
 
 # support for appliance_console methods
 unless defined?(say)
@@ -87,6 +88,10 @@ module ApplianceConsole
       options[:server]
     end
 
+    def db_hourly_maintenance?
+      options[:db_hourly_maintenance]
+    end
+
     def initialize(options = {})
       self.options = options
     end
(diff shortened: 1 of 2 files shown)


lib/gems/pending/appliance_console/cli.rb
```diff
@@ -117,6 +122,7 @@ def parse(args)
         opt :username, "Database Username",  :type => :string,  :short => 'U', :default => "root"
         opt :password, "Database Password",  :type => :string,  :short => "p"
         opt :dbname,   "Database Name",      :type => :string,  :short => "d", :default => "vmdb_production"
+        opt :db_hourly_maintenance, "Configure database hourly maintenance", :type => :bool, :short => 'b'
```
carbonin: I don't think a short option here makes sense.

What we're doing is a bit too complicated for a once character abbreviation I think.

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -117,6 +122,7 @@ def parse(args)
         opt :username, "Database Username",  :type => :string,  :short => 'U', :default => "root"
         opt :password, "Database Password",  :type => :string,  :short => "p"
         opt :dbname,   "Database Name",      :type => :string,  :short => "d", :default => "vmdb_production"
+        opt :db_hourly_maintenance, "Configure database hourly maintenance", :type => :bool, :short => :none
```
carbonin: You can just not specify the `:short` key, we do it with things like `:logdisk` and `:dbdisk` if you need an example.

rsevilla87: I have a strange behavior with the trollop Gem, If i don't specify the ```none``` flag, the short option help is displayed.

rsevilla87: @carbonin as I can see in the trollop code, the case when only the short option is displayed occurs when all letters from the option name are being used by other short options.

PR #263, opened 2017-08-22
Kludge: doubly `ensure` clearing outermost realtime_block on Timeout

cben:
**Based on #269**

Fixes https://github.com/ManageIQ/manageiq/issues/9642.
Timeout::Error is just one regular exception, it can be caught.  But it can strike anywhere.
`ensure` is not broken, if it hits inside `begin..ensure` that's fine, the cleanup runs.
BUT if it strikes after `ensure`, it can abort the cleanup!
Also it can strike between performing `current_realtime=` and even entering begin..ensure.
(Instrumenting the test to log stacktrace where exception hit confirms these are the 2 problematic points:
https://gist.github.com/cben/d5ea134a1f556023905488f1a6a756bf)

The Hobbit - Bain gives Bard the Black Arrow "Timeout Errow"

Good news is there is only one Timeout::~~Arrow~~Error!
(Unless we do nested Timeouts, and they expire in same millisecond.)
In this code the cleanup is safe to attempt twice.  At first I wanted to do something like this, which still had weak spots:
```diff
      self.current_realtime = hash if outermost
# 🏹 ➸ Weak spot here!
      begin
        ret = realtime_store(hash, key, &block)
        return ret, hash
      ensure
# 🏹   ➸ Weak spot here!  not covered by either begin..ensure
+       begin
          delete_current_realtime if outermost
+       ensure
+         delete_current_realtime if outermost && self.in_realtime_block?
+       end
      end
```
*This is why dragons have **overlapping** scales.*  A second begin..ensure covering the whole thing works, at least one `ensure` will complete uninterrupted. :-)
Successfully run 10M iterations :heavy_check_mark: 

Spec shamelessly stolen from @jrafanie's POC in https://github.com/ManageIQ/manageiq/issues/9642.
100k iterations take ~7seconds, doubling this repo's test suite :-(
10k usually missed the bug, it averages ~1/30k for me.
Can't narrow the time range down much (see gist above, wide time distribution to hit either spot).

@miq-bot add-label bug

### DIFF
--- a/lib/gems/pending/util/extensions/miq-benchmark.rb
+++ b/lib/gems/pending/util/extensions/miq-benchmark.rb
@@ -22,15 +22,27 @@ module Benchmark
   # call completes.  If the hash already has a value for that key, the time is
   # accumulated.
   def self.realtime_block(key, &block)
-    outermost = !self.in_realtime_block?
     hash = current_realtime
-    self.current_realtime = hash if outermost
 
-    begin
+    if in_realtime_block?
       ret = realtime_store(hash, key, &block)
       return ret, hash
-    ensure
-      delete_current_realtime if outermost
+    else
+      # Outermost block.
+      begin
+        self.current_realtime = hash
+        begin
+          ret = realtime_store(hash, key, &block)
+          return ret, hash
+        ensure
+          delete_current_realtime
+        end
+      ensure
+        # A second layer of protection in case TimeoutError struck right after
+        # setting self.current_realtime, or right before `delete_current_realtime`.
+        # In those cases, current_realtime might (wrongly) still exist.
+        delete_current_realtime if in_realtime_block?
+      end
     end
   end
 
--- a/spec/util/extensions/miq-benchmark_spec.rb
+++ b/spec/util/extensions/miq-benchmark_spec.rb
@@ -67,12 +67,16 @@ describe Benchmark do
   it "Timeout raising within .realtime_block" do
     expect(Benchmark.in_realtime_block?).to be_falsey
 
-    places_outcomes = {}
-    1000.times do |i|
+    # If something left over from previous tests gets GC'd during Timeout, its finalizer may
+    # be interrupted.  For example, interrupted Temfile cleanup may get stuck forever!
+    # This reduces chance of nasty interactions...
+    GC.start
+
+    20.times do |i|
       begin
         # keep entering/exiting, abort ASAP
         Timeout.timeout(1e-9) do
-          loop do
+          2_000_000.times do
             Benchmark.realtime_block(:test1) do
               Benchmark.realtime_block(:test2) do
               end
@@ -80,25 +84,12 @@ describe Benchmark do
             end
           end
         end
+        raise("Completed without Timeout!?  Tip: may result from unclosed Tempfile object.")
       rescue Timeout::Error => e
-        # sortable compact stacks: realtime_block and children, outer first, strip directories, pad line numbers
-        interesting_depth = e.backtrace.rindex { |s| s =~ /in `realtime_block'/ } || 5
-        where = e.backtrace[0..interesting_depth].reverse
-        where = where.collect { |s| s.sub(%r{/.*/}, '').sub(/\d+:/) { |linenum| '%03d:' % linenum.to_i } }
-        places_outcomes[where] ||= {:cleaned => 0, :failed => 0}
-        if Benchmark.in_realtime_block?
-          places_outcomes[where][:failed] += 1
-          Benchmark.delete_current_realtime
-        else
-          places_outcomes[where][:cleaned] += 1
-        end
-      else
-        fail "impossible: escaped infinite loop without exception"
+        where = e.backtrace.first(10).join("\n")
+        expect(Benchmark.in_realtime_block?).to be_falsey,
+                                                "failed on #{i}th Timeout, #{e.inspect} interrupted in:\n#{where}"
       end
-    end
-
-    places_outcomes.sort.each do |where, outcomes|
-      puts "cleaned\t#{outcomes[:cleaned]}\tfailed\t#{outcomes[:failed]}\t#{where.join(' -> ')}"
     end
   end
 end
--- a/spec/util/extensions/miq-benchmark_spec.rb
+++ b/spec/util/extensions/miq-benchmark_spec.rb
@@ -1,8 +1,13 @@
 require 'util/extensions/miq-benchmark'
 require 'timecop'
+require 'timeout'
 
 describe Benchmark do
   after(:each) { Timecop.return }
+  after(:each) do
+    # Isolate other tests
+    Benchmark.delete_current_realtime if Benchmark.in_realtime_block?
+  end
 
   it '.realtime_store' do
     timings = {}
@@ -57,5 +62,43 @@ describe Benchmark do
       expect(Benchmark.in_realtime_block?).to be_truthy
     end
     expect(Benchmark.in_realtime_block?).to be_falsey
+  end
+
+  it "Timeout raising within .realtime_block" do
+    expect(Benchmark.in_realtime_block?).to be_falsey
+
+    places_outcomes = {}
+    1000.times do |i|
+      begin
+        # keep entering/exiting, abort ASAP
+        Timeout.timeout(1e-9) do
+          loop do
+            Benchmark.realtime_block(:test1) do
+              Benchmark.realtime_block(:test2) do
+              end
+              "result"
+            end
+          end
+        end
+      rescue Timeout::Error => e
+        # sortable compact stacks: realtime_block and children, outer first, strip directories, pad line numbers
+        interesting_depth = e.backtrace.rindex { |s| s =~ /in `realtime_block'/ } || 5
+        where = e.backtrace[0..interesting_depth].reverse
+        where = where.collect { |s| s.sub(%r{/.*/}, '').sub(/\d+:/) { |linenum| '%03d:' % linenum.to_i } }
+        places_outcomes[where] ||= {:cleaned => 0, :failed => 0}
+        if Benchmark.in_realtime_block?
+          places_outcomes[where][:failed] += 1
+          Benchmark.delete_current_realtime
+        else
+          places_outcomes[where][:cleaned] += 1
+        end
+      else
+        fail "impossible: escaped infinite loop without exception"
+      end
+    end
+
+    places_outcomes.sort.each do |where, outcomes|
+      puts "cleaned\t#{outcomes[:cleaned]}\tfailed\t#{outcomes[:failed]}\t#{where.join(' -> ')}"
+    end
   end
 end


Fryguy:
"Here be dragons" 😬 

I'm good with this, and I don't mind that it adds 7 seconds.  Since this happen 1/30k can you make the number 50k iteration?

lib/gems/pending/util/extensions/miq-benchmark.rb
```diff
@@ -22,15 +22,20 @@ def self.realtime_store(hash, key)
...
+      begin
+        ret = realtime_store(hash, key, &block)
+        return ret, hash
+      ensure
+        delete_current_realtime if outermost
+      end
+    # A second layer of protection in case TimeoutError struck before or after `begin..ensure`
     ensure
-      delete_current_realtime if outermost
+      delete_current_realtime if outermost && in_realtime_block?
```
Fryguy: I'm trying to grok this, but I don't understand.  `outermost = !in_realtime_block`, so this code is saying `delete_current_realtime if !in_realtime_block? && in_realtime_block?`, which is always, right?

cben: `outermost` is cached `!in_realtime_block?` at begining; `in_realtime_block?` is a call checking current state (original had `self.in_realtime_block?` which might have stressed it, drop because rubocop).
At this point it _should_ have original value restored, yes, so this && should never happen — unless Timeout::Error...

Rewritten as 2 branches — outermost or not — which IMO reads clearer? PTAL.

Fryguy:
@jrafanie Would like you 👀  on this as well.

lib/gems/pending/util/extensions/miq-benchmark.rb
```diff
@@ -22,15 +22,27 @@ def self.realtime_store(hash, key)
...
+          ret = realtime_store(hash, key, &block)
+          return ret, hash
+        ensure
+          delete_current_realtime
+        end
+      ensure
+        # A second layer of protection in case TimeoutError struck right after
+        # setting self.current_realtime, or right before `delete_current_realtime`.
+        # In those cases, current_realtime might (wrongly) still exist.
+        delete_current_realtime && in_realtime_block?
```
jrafanie: @cben I don't understand what the `&& in_realtime_block?` is doing.

Is that needed?

jrafanie: Was it intended to be `if in_realtime_block?` ?

jrafanie:
@cben This looks good to me.  It's a simple solution to a very complicated problem.  You're 100% right that it's perfectly fine for us to try to `delete_current_realtime` twice and at least one of them will always complete.

lib/gems/pending/util/extensions/miq-benchmark.rb
cben: Oops bad rewrite, yes, s/&&/if/.
This means it's safe to do without checking :-)
Huh.  TIL ruby Hash#delete just returns nil if no such key.  I should have guessed but I keep getting surprised...

IMHO it's more readable with the check.

jrafanie: Ok, cool.  I'm fine with keeping the `if in_realtime_block?` check.

spec/util/extensions/miq-benchmark_spec.rb
```diff
@@ -58,4 +63,26 @@
     end
     expect(Benchmark.in_realtime_block?).to be_falsey
   end
+
+  it "Timeout raising within .realtime_block" do
+    expect(Benchmark.in_realtime_block?).to be_falsey
+    50_000.times do |i|
+      timeout = 10**Random.rand(-13.0..-10.0)
```
cben: Wait this is funny, 1e-10 is already 0.1ns, less than a cpu cycle!  (Moore proof haha).
My "distribution" showed that timeouts in this range do trigger the bug, sometimes, but I think it just expired immediately and depended on thread context switch...

I'll measure with longer values.

cben:
Turned out, whatever the timeout number, less than 1 in 1000 Timeouts were hitting the realtime_block at all, most blocks were completing sucessfully.
=> Rewrote test to loop entering and exiting blocks until the Timeout hits.  Now about 1 in 10 Timeouts catches the bug, but due to loop count it's only mildly faster (1st and 2nd commit have details).  Whatever.
=> Prettier stacks tally https://gist.github.com/cben/d5ea134a1f556023905488f1a6a756bf but nothing new.

What I'm stuck on now, is that running the full suite rather than just this test, the infinite loop never finishes.  I could limit the loop but then the test is ineffective.  Anyway, how come Timeout doesn't work?!
It's an order dependency, I've bisected it to Tempfile.new.  If I add `Tempfile.new("x")` at the start of this spec, and run just this spec, it loops forever.
Any ideas?

jrafanie:
> It's an order dependency, I've bisected it to Tempfile.new. If I add Tempfile.new("x") at the start of this spec, and run just this spec, it loops forever.
> Any ideas?

@cben Weird.  I recreated it "sporadically" and ran it with debug and it's definitely the cleanup of the tempfile that is hanging...

Here's a good run:

```
11:57:14 ~/Code/manageiq-gems-pending (cben-overlapping-timeout-armor) (2.3.4) - RUBYOPT=-d be rspec spec/util/extensions/miq-benchmark_spec.rb:69
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems.rb:1241 - cannot load such file -- rubygems/defaults/operating_system
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems.rb:1250 - cannot load such file -- rubygems/defaults/ruby
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55 - cannot load such file -- did_you_mean
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55 - cannot load such file -- bundler
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
...
Exception `Errno::EEXIST' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:254 - File exists @ dir_s_mkdir - /Users/joerafaniello/.gem/ruby/2.3.4
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:558 - method `gem' not defined in #<Class:Kernel>
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:556 - undefined method `find_spec_for_exe' for class `#<Class:Gem>'
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:556 - undefined method `activate_bin_path' for class `#<Class:Gem>'
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/rspec-core-3.5.4/lib/rspec/core/pending.rb:20 - uninitialized constant RSpec::Core::Pending::Test
Using Ext extension for JSON.
Run options: include {:locations=>{"./spec/util/extensions/miq-benchmark_spec.rb"=>[69]}}

Randomized with seed 7990

Benchmark
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
removing /var/folders/fq/blrz820d3qz7nm7vj8mbtfs40000gq/T/x20170906-57250-s683fs...
done
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
...
  Timeout raising within .realtime_block

Finished in 1.96 seconds (files took 0.15031 seconds to load)
1 example, 0 failures
```

Here's when it hangs:

```
11:57:22 ~/Code/manageiq-gems-pending (cben-overlapping-timeout-armor) (2.3.4) - RUBYOPT=-d be rspec spec/util/extensions/miq-benchmark_spec.rb:69
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems.rb:1241 - cannot load such file -- rubygems/defaults/operating_system
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems.rb:1250 - cannot load such file -- rubygems/defaults/ruby
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55 - cannot load such file -- did_you_mean
Exception `LoadError' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/rubygems/core_ext/kernel_require.rb:55 - cannot load such file -- bundler
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
...
Exception `TypeError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/vendor/thor/lib/thor/base.rb:637 - can't dup FalseClass
Exception `Errno::EEXIST' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/fileutils.rb:254 - File exists @ dir_s_mkdir - /Users/joerafaniello/.gem/ruby/2.3.4
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:558 - method `gem' not defined in #<Class:Kernel>
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:556 - undefined method `find_spec_for_exe' for class `#<Class:Gem>'
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/bundler-1.15.4/lib/bundler/rubygems_integration.rb:556 - undefined method `activate_bin_path' for class `#<Class:Gem>'
Exception `NameError' at /Users/joerafaniello/.gem/ruby/2.3.4/gems/rspec-core-3.5.4/lib/rspec/core/pending.rb:20 - uninitialized constant RSpec::Core::Pending::Test
Using Ext extension for JSON.
Run options: include {:locations=>{"./spec/util/extensions/miq-benchmark_spec.rb"=>[69]}}

Randomized with seed 61986

Benchmark
Exception `Timeout::Error' at /Users/joerafaniello/.rubies/ruby-2.3.4/lib/ruby/2.3.0/timeout.rb:112 - execution expired
removing /var/folders/fq/blrz820d3qz7nm7vj8mbtfs40000gq/T/x20170906-57351-jfeu6l...
```

jrafanie:
@cben when it hangs, it's here:  https://github.com/ruby/ruby/blob/820605ba3c10b9f4dafc4e5d6e09765b8b31cbea/lib/tempfile.rb#L255-L257

jrafanie:
@cben So, tl;dr:  Timeout is horrible and appears to be raising the exception in the finalizer when the tempfile object is being GC'd.

I'm able to pretty reliably recreate it on osx with this script:

```ruby
require 'tempfile'
require 'timeout'
Tempfile.new("x")

def stuff
  ""
end

20.times do |i|
  begin
    Timeout.timeout(1e-9) do
      loop do
        begin
          "result"
        ensure
        end
      end
    end
  rescue Timeout::Error => e
  end
end
```

So, I'm leaning towards not testing this explicitly.

jrafanie:
@cben I don't expect they'll find a way to reliably fix this, since it's Timeout's thread unsafe code at fault but I opened a bug anyway: https://bugs.ruby-lang.org/issues/13876

cben:
Awesome!  I mean horrible.
I'll fix some other specs to close and unlink their Tempfile, good idea
anyway.
Will also make the loop bounded with helpful message if exceeded.

Also GC.start before running tons of `Timeout` may be good idea to cleanup
remains from other tests.  (That alone does help but not 100%...)

cben:
Rebased on #269.  Added GC.start, made loop bounded.  PTAL.

Fryguy:
Meregd #269

Fryguy:
Why fail, as opposed to raise?

spec/util/extensions/miq-benchmark_spec.rb
```diff
@@ -58,4 +63,33 @@
...
+        Timeout.timeout(1e-9) do
+          100_000.times do
+            Benchmark.realtime_block(:test1) do
+              Benchmark.realtime_block(:test2) do
+              end
+              "result"
+            end
+          end
+        end
+        fail("Completed without Timeout!?  Tip: may result from unclosed Tempfile object.")
```
jrafanie: Good 👁 @Fryguy, yeah, I'm not sure why we fail here.  From testing this, we get to this line when the block is fast and the timeout is never raised OR when it's slow and the tempfile is cleaned up without the exception being raised.

jrafanie: @cben ^

cben: My intent was get error from this test when it's stuck due to tempfile, instead of having Travis kill job at 10min.
When the block is fast — I tried to set enough loops so it never happens.  For me timeout normally hits around 10,000 loops.
But 100,000 is <2sec, so yeah, bumped to 2,000,000 which is ~30sec.
Anyway if for any reason it completes without Timeout, that means this test is not testing anything, and I wanted an indication.  Is failing too harsh?

["fail" because that's how stack overflow told me how to fail an rspec test, sounded right :-)  changed to "raise".]

jrafanie:
LGTM.  Let's see how it goes 👍

PR #269, opened 2017-09-07
Cleanup after specs: Tempfiles, Readline redirection

cben:
Both close and unlink all Tempfiles created in specs.
Helpful for #263 to avoid https://bugs.ruby-lang.org/issues/13876, and good practice anyway.
@jrafanie please review


jrafanie:
Nice, thanks @cben

PR #276, opened 2017-09-29
Force to use different partition for db

ailisp:
ISSUE:
We should not allow using same partition for OS and database.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1425153

\cc @gtanzillo @yrudman @carbonin 

@miq-bot add-label wip,bug

### DIFF
--- a/lib/gems/pending/appliance_console/prompts.rb
+++ b/lib/gems/pending/appliance_console/prompts.rb
@@ -143,12 +143,13 @@ module ApplianceConsole
       just_ask(prompt, default, INT_REGEXP, "an integer", Integer) { |q| q.in = range if range }
     end
 
-    def ask_for_disk(disk_name, verify = true)
+    def ask_for_disk(disk_name, verify = true, allow_none = true)
       require "linux_admin"
       disks = LinuxAdmin::Disk.local.select { |d| d.partitions.empty? }
 
       if disks.empty?
         say "No partition found for #{disk_name}. You probably want to add an unpartitioned disk and try again."
+        raise MiqSignalError unless allow_none
       else
         default_choice = disks.size == 1 ? "1" : nil
         disk = ask_with_menu(
--- a/lib/gems/pending/appliance_console/database_replication_standby.rb
+++ b/lib/gems/pending/appliance_console/database_replication_standby.rb
@@ -30,7 +30,7 @@ module ApplianceConsole
       clear_screen
       say("Establish Replication Standby Server\n")
       return false if !data_dir_empty? && !confirm_data_resync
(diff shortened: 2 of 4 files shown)


ailisp:
@yrudman I'm not sure with what to do with replication db. Please make a confirmation so I can write specs.

yrudman:
I think it should be the same for replication too

ailisp:
@miq-bot remove-label wip

lib/gems/pending/appliance_console/prompts.rb
```diff
@@ -143,7 +143,7 @@ def ask_for_integer(prompt, range = nil, default = nil)
       just_ask(prompt, default, INT_REGEXP, "an integer", Integer) { |q| q.in = range if range }
     end
 
-    def ask_for_disk(disk_name, verify = true)
+    def ask_for_disk(disk_name, verify = true, force = false)
```
carbonin: I feel like this is using the word "force" in the opposite way I would expect.

If someone calls `#ask_for_disk` I would think that "force" would mean "Give me an answer no matter what". What you're doing here feels more like "Abort if we can't get a disk".

I think this would be better if the parameter had a more clear name, maybe `allow_none`?

ailisp: you're right, thanks

ailisp:
@carbonin Now the naming is much better 😄

lib/gems/pending/appliance_console/prompts.rb
```diff
@@ -162,6 +162,7 @@ def ask_for_disk(disk_name, verify = true)
 
       if verify && disk.nil?
         say ""
+        raise MiqSignalError unless allow_none
```
carbonin: I feel like this won't work if the method is called without verify.

So as an example `ask_for_disk("disk", false, false)` would not raise.

Also, we have a check for `disks.empty?` above, should this raise be there rather than down here?

ailisp: I thought verify is to ask manageiq to check disk. verify=false is only used in case we're sure disk is valid or invalid disk doesn't matter, in either case no need to verify. So allow_none only make sense and should be used when verify =true. But your comment seems better. I'll change it.

ailisp:
@carbonin This is now solved by https://github.com/ManageIQ/manageiq-gems-pending/pull/277
Shall we close this?

carbonin:
Yes, I think the approach in #277 is better. But this PR also changes the behavior of the replication standby disk configuration.

I think we should close this PR, and maybe open another for handling the replication standby config in a similar way.

PR #277, opened 2017-09-29
internal database must be set with dbdisk option in ap and in cli

ailisp:
ISSUE:
When setting internal database in appliance console cli, the user can set without given a --dbdisk option, which is not supported.
BZ:
https://bugzilla.redhat.com/show_bug.cgi?id=1431815
https://bugzilla.redhat.com/show_bug.cgi?id=1425153

\cc @yrudman @gtanzillo 

@miq-bot add-label wip,bug

### DIFF
--- a/lib/gems/pending/appliance_console/internal_database_configuration.rb
+++ b/lib/gems/pending/appliance_console/internal_database_configuration.rb
@@ -3,6 +3,7 @@ require "appliance_console/logical_volume_management"
 require "pathname"
 require "util/postgres_admin"
 require "pg"
+require "linux_admin"
 
 RAILS_ROOT ||= Pathname.new(__dir__).join("../../../")
 
@@ -49,6 +50,7 @@ module ApplianceConsole
 
     def ask_questions
       choose_disk
+      check_disk_is_mount_point
       self.run_as_evm_server = !ask_yn?(<<-EOS.gsub!(/^ +/m, ""), "N")
 
         Should this appliance run as a standalone database server?
@@ -67,6 +69,11 @@ module ApplianceConsole
 
     def choose_disk
       @disk = ask_for_disk("database disk")
+    end
+
+    def check_disk_is_mount_point
+      error_message = "The disk for database must be a mount point"
+      raise error_message unless disk || pg_mount_point?
     end
 
     def initialize_postgresql_disk
@@ -108,7 +115,7 @@ module ApplianceConsole
     private
 
     def mount_point
-      Pathname.new(ENV.fetch("APPLIANCE_PG_MOUNT_POINT"))
+      PostgresAdmin.mount_point
     end
 
     def copy_template(src, src_dir = self.class.postgresql_template, dest_dir = PostgresAdmin.data_directory)
@@ -119,6 +126,10 @@ module ApplianceConsole
       else
         FileUtils.cp full_src, dest_dir
       end
+    end
+
+    def pg_mount_point?
+      LinuxAdmin::LogicalVolume.mount_point_exists?(mount_point)
     end
 
     def run_initdb
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -200,6 +200,7 @@ module ApplianceConsole
         :disk              => disk_from_string(options[:dbdisk]),
         :run_as_evm_server => !options[:standalone]
       }.delete_if { |_n, v| v.nil? })
+      config.check_disk_is_mount_point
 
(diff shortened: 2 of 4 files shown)


ailisp:
@miq-bot remove-label wip

yrudman:
@ailisp there is another check requested in the same BZ- do not allow blank password

carbonin:
I believe this will break ManageIQ appliance builds.

See https://github.com/ManageIQ/manageiq-appliance-build/blob/master/kickstarts/partials/post/db_init.ks.erb#L7

I think we should come up with a way to allow this behavior through the CLI.

For example if someone configures a logical volume and wants to put the database there (as we do on the ManageIQ appliance), we should allow them to do that.

As I see it, the real issue is that we want the database on a separate *filesystem* not necessarily a separate disk. Unfortunately, we don't do any validation against filesystems so checking that case would be much more difficult.

carbonin:
> there is another check requested in the same BZ- do not allow blank password

@yrudman I think it would be better to do this one in a separate PR.

yrudman:
@carbonin yes,  it is better to do password present validation in different PR

ailisp:
> As I see it, the real issue is that we want the database on a separate filesystem not necessarily a separate disk. Unfortunately, we don't do any validation against filesystems so checking that case would be much more difficult.
@carbonin do you mean the device number of filesystem? In this question:
https://unix.stackexchange.com/questions/44249/how-to-check-if-two-directories-or-files-belong-to-same-filesystem
I did some try:
```
[boyao@localhost ~]$ df -h
Filesystem               Size  Used Avail Use% Mounted on
devtmpfs                 5.8G     0  5.8G   0% /dev
tmpfs                    5.8G  988K  5.8G   1% /dev/shm
tmpfs                    5.8G  1.6M  5.8G   1% /run
tmpfs                    5.8G     0  5.8G   0% /sys/fs/cgroup
/dev/mapper/fedora-root   50G  6.8G   40G  15% /
tmpfs                    5.8G   88K  5.8G   1% /tmp
/dev/sda1                477M  149M  299M  34% /boot
/dev/mapper/fedora-home  180G   26G  145G  16% /home
tmpfs                    1.2G   24K  1.2G   1% /run/user/42
tmpfs                    1.2G   32K  1.2G   1% /run/user/1000
[boyao@localhost ~]$ stat -c "%d" /
64769
[boyao@localhost ~]$ stat -c "%d" /var
64769
[boyao@localhost ~]$ stat -c "%d" /dev
6
```

ailisp:
@yrudman Thanks for the note. Seems I should wait this PR merged then open another PR for checking password. otherwise seems some conflict will happen in prompts.rb

carbonin:
@ailisp Generally, yes, but I don't think the place for that is in the cli class.

If we want a more complete fix, we could edit the `InternalDatabaseConfiguration` class so that we have an option to verify that the data directory is on a separate filesystem, alternatively you can check that `APPLIANCE_PG_MOUNT_POINT` is indeed a mount point.

yrudman:
@ailisp you can create tracking  branch:
`git checkout -b feature_2 --track feature_1`

ailisp:
@miq-bot add-label wip

ailisp:
@carbonin Yuri and I think this https://github.com/ManageIQ/manageiq-gems-pending/pull/276 can be closed now, because your suggestion to check `APPLIANCE_PG_MOUNT_POINT` should solve both BZs. What do you think

lib/gems/pending/appliance_console/internal_database_configuration.rb
```diff
@@ -121,6 +123,11 @@ def copy_template(src, src_dir = self.class.postgresql_template, dest_dir = Post
       end
     end
 
+    def check_pg_mount_point
+      is_mount_point = Class.new.include(LinuxAdmin::Mountable).mount_point_exists?(mount_point)
```
carbonin: Can you not just use `LinuxAdmin::LogicalVolume` here?

lib/gems/pending/appliance_console/internal_database_configuration.rb
```diff
@@ -82,6 +83,7 @@ def initialize_postgresql_disk
 
     def initialize_postgresql
       log_and_feedback(__method__) do
+        check_pg_mount_point
```
carbonin: I think this would make more sense as a check in `activate` or even in ask_questions.

Right after choose disk we could print some error message `unless disk || have_pg_mount_point?`

ailisp:
@miq-bot remove-label wip

ailisp:
@yrudman @carbonin 
works fine in both appliance console and cli now

PR #278, opened 2017-10-02
Update nokogiri to a more secure version.

NickLaMuro:
With the introduction of the following security vulnerability on libxml2 (a huge dependency of nokogiri):

https://access.redhat.com/security/cve/cve-2017-9050

`manageiq-gems-pending` should strive to try and use the latest version of the nokogiri gem to avoid including this vulnerability in dependencies of it.


simaishi:
Backported to Fine via https://github.com/ManageIQ/manageiq-gems-pending/pull/280

PR #281, opened 2017-10-04
must give password to set db in cli

ailisp:
ISSUE: Currently we allow user to set up database in appliance console cli without given a password, which should be disallowed. This is part of BZ https://bugzilla.redhat.com/show_bug.cgi?id=1431815
Another part (disallow --internal without --dbdisk) is fixed in https://github.com/ManageIQ/manageiq-gems-pending/pull/277

BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1431815#c2

\cc @yrudman @carbonin @gtanzillo 

@miq-bot add-label wip, bug

### DIFF
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -182,11 +182,16 @@ module ApplianceConsole
 
     def set_db
       raise "No encryption key (v2_key) present" unless key_configuration.key_exist?
+      raise "Must provide a password to set database" unless password?
       if local?
         set_internal_db
       else
         set_external_db
       end
+    end
+
+    def password?
+      options[:password] && !options[:password].strip.empty?
     end
 
     def set_internal_db
--- a/lib/gems/pending/appliance_console/cli.rb
+++ b/lib/gems/pending/appliance_console/cli.rb
@@ -182,7 +182,7 @@ module ApplianceConsole
 
     def set_db
       raise "No encryption key (v2_key) present" unless key_configuration.key_exist?
-      raise "Must provide a password to set database" unless password?
+      raise "A password is required to configure a database" unless password?
       if local?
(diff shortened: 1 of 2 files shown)


ailisp:
@miq-bot remove-label wip

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -182,13 +182,18 @@ def run
 
     def set_db
       raise "No encryption key (v2_key) present" unless key_configuration.key_exist?
+      raise "Must provide a password to set database" unless password?
```
carbonin: I think we can improve this sentence a bit ...

Maybe "A password is required to configure a database"

yrudman: 👍

lib/gems/pending/appliance_console/cli.rb
```diff
@@ -182,13 +182,18 @@ def run
...
         set_internal_db
       else
         set_external_db
       end
     end
 
+    def password?
+      options[:password] && !options[:password].strip.empty?
+    end
+
```
yrudman: would below look more complicated or easer to read ? : 
`options[:password] ? !options[:password].strip.empty? : false`

carbonin: I think it's fine the way it is.

carbonin:
Looks good @ailisp. Just fix up the rubocop complaints and we should be good to go.

In this case you can either just add the spaces or (as the rest of this file seems to do), use `expect` with the `do ... end` block notation. I'm fine with either.

lib/gems/pending/appliance_console/cli.rb
ailisp: Thanks @yrudman and @carbonin . Now should be good.

PR #284, opened 2017-10-12
Move miq-file extension to manageiq-smartstate

Fryguy:
SmartState related methods are the only callers, so this can live in the
SmartState gem as dedicated helpers.

WIP for now as there are still callers that need to be converted over...will un-WIP when done.


Fryguy:
Depends on
- ManageIQ/manageiq-providers-vmware#120
- ManageIQ/manageiq-providers-ovirt#105
- ManageIQ/manageiq-providers-scvmm#21

Fryguy:
ManageIQ/manageiq-smartstate#33 was already merged, released as 0.2.0 and bumped in manageiq core in ManageIQ/manageiq#16194

Fryguy:
All dependencies merged.  @agrare Please review, as I think this is good to go.

PR #285, opened 2017-10-13
Allow ipv6 address as dns when setting static ipv4 

ailisp:
ISSUE:
When setting ipv4 static network in appliance console, we didn't allow ipv6 as dns. Now it's allowed.
BZ:
allow ipv6 address as dns when setting static ipv4 

\cc @yrudman @gtanzillo 
@miq-bot add-label wip, bug

### DIFF
--- a/lib/gems/pending/appliance_console.rb
+++ b/lib/gems/pending/appliance_console.rb
@@ -195,8 +195,8 @@ To modify the configuration, use a web browser to access the management page.
           new_ip   = ask_for_ipv4("IP Address", ip)
           new_mask = ask_for_ipv4("Netmask", mask)
           new_gw   = ask_for_ipv4("Gateway", gw)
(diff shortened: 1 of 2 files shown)


ailisp:
@miq-bot remove-label wip

carbonin:
Why are we doing this? Is there a use case for mixing IP protocol versions for DNS?

It does look like it's worth investigating as the `static6` case seems to allow either IPv4 or IPv6 for the primary DNS server, but then restricts to IPv6 only for the secondary.

ref: https://github.com/ManageIQ/manageiq-gems-pending/blob/111571beef5f60f35e6268dc4d83a476cb85286b/lib/gems/pending/appliance_console.rb#L250-L251

yrudman:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1501260

PR #292, opened 2017-10-17
Extract the appliance console

bdunne:
https://github.com/ManageIQ/manageiq-appliance_console

Depends on https://github.com/ManageIQ/manageiq/pull/16222


abellotti:
😍 NICE!!!

PR #296, opened 2017-10-18
[WIP] Extract manageiq postgres admin

bdunne:
Waiting on v1.0 release


bdunne:
We're leaving this in manageiq-gems-pending for now

PR #297, opened 2017-10-18
Move database_in_recovery? from PostgresAdmin to PostgresHaAdmin

bdunne:


spec/postgres_ha_admin/failover_monitor_spec.rb
```diff
@@ -168,4 +168,24 @@ def stub_monitor_constants
   def active_databases_conninfo
     [{}, {}, {}]
   end
+
+  context "private" do
+    describe "#database_in_recovery?" do
+      before do
+        begin
+          @connection = PG::Connection.open(:dbname => 'travis', :user => 'travis')
```
Fryguy: Every database should have a dbname of template1, and you can probably use postgres as the user.

Fryguy: Oh it's a copy...nevermind.

PR #298, opened 2017-10-19
Allow Rails 5.1

chrisarcand:
For https://github.com/ManageIQ/manageiq/issues/15036

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -20,9 +20,9 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "> 2.2.2"
 
-  s.add_runtime_dependency "actionpack",              ">= 5.0", "< 5.1"
-  s.add_runtime_dependency "activerecord",            ">= 5.0", "< 5.1"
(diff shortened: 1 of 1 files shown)


manageiq-gems-pending.gemspec
```diff
@@ -20,9 +20,9 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "> 2.2.2"
 
-  s.add_runtime_dependency "actionpack",              ">= 5.0", "< 5.1"
-  s.add_runtime_dependency "activerecord",            ">= 5.0", "< 5.1"
-  s.add_runtime_dependency "activesupport",           ">= 5.0", "< 5.1"
+  s.add_runtime_dependency "actionpack",              ">= 5.0", "< 5.2"
+  s.add_runtime_dependency "activerecord",            ">= 5.0", "< 5.2" # used by appliance_console
```
bdunne: The `appliance_console` comment should be removed since that has been extracted

bdunne:
It's getting pretty late to introduce this into G-release, isn't it?

Fryguy:
😍 

But, this should be for after we branch gaprindashvili.  It's too late in the cycle to introduce it now.

PR #312, opened 2017-11-15
[POC] Store unique set size (USS) in the PSS column

jrafanie:
https://bugzilla.redhat.com/show_bug.cgi?id=1479356

Unique set size is a better way to detect workers that are growing
unbounded since any memory/reference leaks would be shown in their
uss.  If the server process is large when forking, new workers would
inherit a big pss immediately.

We should really rename the column/hash key to uss.

Related PRs:
[master](https://github.com/ManageIQ/manageiq-gems-pending/pull/312)
[gaprindashvili](https://github.com/ManageIQ/manageiq-gems-pending/pull/314)
[fine](https://github.com/ManageIQ/manageiq-gems-pending/pull/313)
[euwe](https://github.com/ManageIQ/manageiq/pull/16480)


jrafanie:
For gaprindashvili and master, we should use https://github.com/ManageIQ/manageiq/pull/16570 instead

PR #315, opened 2017-11-21
Truncating log messages when they are too large

juliancheal:


juliancheal:
@Fryguy @agrare wip right now, just want to make sure I'm on the right path.

juliancheal:
Oh and I realised the `max_message_size` is not 1MB it's 1KB I think, whoops. Can change that.

juliancheal:
@miq-bot assign @Fryguy

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -157,7 +157,7 @@ class Formatter < Logger::Formatter
 
     def call(severity, time, progname, msg)
       msg = prefix_task_id(msg2str(msg))
-
+      msg = truncate_large_lines(msg)
```
agrare: How about just `msg = prefix_task_id(msg2str(msg)).truncate(max_log_line_length)` ?

juliancheal: Guess so, was worried about making an incomplete char at the end of the line, then again that could happen with `byteslice` too.

And I guess we're truncating the line anyway so a bust char wouldn't be the end of the world?

agrare: yeah and it automatically appends `...` so you can tell that it was truncated

juliancheal:
ha so the test fails as

```
       expected: 1048576
            got: 1048647
```
We expected 1.megabyte, but of course there is the other data in the log file. So need to change the test to account for that, but almost there.

juliancheal:
Ha didn't account for `…` at the end of the message.

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -156,8 +156,9 @@ class Formatter < Logger::Formatter
     FORMAT = "[----] %s, [%s#%d:%x] %5s -- %s: %s\n"
 
     def call(severity, time, progname, msg)
-      msg = prefix_task_id(msg2str(msg))
+      max_log_line_length = 1.megabyte # TODO: (julian) Move to config
```
Fryguy: Don't worry about config, because I can't see a user configuring this.

juliancheal: Ok thanks @Fryguy.

Fryguy: You might need to add some activesupport extensions at the top of this file...it can run standalone without the rails env.  See how we use core_ext above.

spec/util/vmdb-logger_spec.rb
```diff
@@ -95,6 +95,18 @@
     expect(VMDBLogger.contents("mylog.log")).to eq("")
   end
 
+  context "long messages" do
+    before(:each) do
```
Fryguy: The `:each` is redundant, so you can remove it... however, you can also just make this a let

```ruby
let(:logger) { VMDBLogger.new(@log) }
```

Fryguy:
I think my other comment got lost;

> You might need to add some activesupport extensions at the top of this file...it can run standalone without the rails env. See how we use core_ext above.

spec/util/vmdb-logger_spec.rb
Fryguy: See way up above for an example of using let

juliancheal: I was future proofing for when we add more tests :)

juliancheal:
🍏 ❓

spec/util/vmdb-logger_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'util/vmdb-logger'
 require 'util/miq-password'
+require 'active_support/core_ext/string'
```
agrare: Think you need this in `lib/gems/pending/util/vmdb-logger.rb` too

Fryguy: Preferably in the vmdb-logger and *not* in the spec.

lib/gems/pending/util/vmdb-logger.rb
```diff
@@ -156,8 +157,9 @@ class Formatter < Logger::Formatter
     FORMAT = "[----] %s, [%s#%d:%x] %5s -- %s: %s\n"
 
     def call(severity, time, progname, msg)
-      msg = prefix_task_id(msg2str(msg))
+      max_log_line_length = 1.megabyte
```
Fryguy: Since this call will happen very frequently, I'm thinking this might be better moved to a constant.  That way it doesn't have to re-evaluate 1.megabyte over and over.

spec/util/vmdb-logger_spec.rb
juliancheal: Oh I see, so `active_support/core_ext/string` should be included via `util/vmdb-logger`? @Fryguy

lib/gems/pending/util/vmdb-logger.rb
Fryguy: Also, is there an active_support/core_ext for this megayute method?

juliancheal: Good point it is.

spec/util/vmdb-logger_spec.rb
juliancheal: @fryguy all done.

juliancheal:
@Fryguy changes all done.

juliancheal:
Ping @Fryguy

spec/util/vmdb-logger_spec.rb
Fryguy: Yes, since that's where you actually use it (for the truncate call)

juliancheal:
@Fryguy I thought I'd pushed my changes, guess not. Sorry my bad! Just made the change you asked for. Should hopefully be green.

juliancheal:
@Fryguy 🍏 We good to go?

simaishi:

```
$ git log -1
commit 36602cedbf23f88d9e7d8e78d7517e70e36ed227
Date:   Thu Jan 4 10:33:47 2018 -0500

    Merge pull request #315 from juliancheal/fix_logger_size_limit
    
    Truncating log messages when they are too large
    (cherry picked from commit d09f86aa78fdaf68b832f28c69cd8024df416bca)
```

PR #318, opened 2017-12-01
Add valid_encoding checking before inserting into xml object.

hsong-rh:
Some special control characters, here '\u000F' and '\u001F', are valid chars in UTF-8 encoding, but unacceptable when we try to convert them into XML. A check is added to remove them first before xml conversion.

https://bugzilla.redhat.com/show_bug.cgi?id=1518249

### DIFF
--- a/lib/gems/pending/util/xml/miq_rexml.rb
+++ b/lib/gems/pending/util/xml/miq_rexml.rb
@@ -148,7 +148,7 @@ module REXML
         rescue => err
           if err.class == ::Encoding::CompatibilityError
             second_utf8 = second.to_s.force_encoding('UTF-8')
-            @value = Text.unnormalize(second_utf8, nil)
+            @value = Text.unnormalize(second_utf8.valid_encoding? ? second_utf8 : "Invalid encoding found", nil)
           else
             $log.error "Encoding error: #{second_utf8}" if $log
           end
@@ -223,7 +223,7 @@ module REXML
     def add_element(element, attrs = nil)
       return add_element_orig(element) if element.kind_of?(REXML::Element)
       attrs.delete_if { |_k, v| v.nil? } unless attrs.nil?
-      add_element_orig(element.to_s, XmlHelpers.stringify_keys(attrs))
+      add_element_orig(element.to_s, XmlHelpers.validate_attrs(attrs))
     end
 
     alias_method :add_attribute_orig, :add_attribute
--- a/lib/gems/pending/util/xml/miq_rexml.rb
+++ b/lib/gems/pending/util/xml/miq_rexml.rb
@@ -147,8 +147,8 @@ module REXML
           @value = Text.unnormalize(second.to_s, nil)
         rescue => err
           if err.class == ::Encoding::CompatibilityError
-            second_utf8 = second.to_s.force_encoding('UTF-8')
-            @value = Text.unnormalize(second_utf8.valid_encoding? ? second_utf8 : "Invalid encoding found", nil)
+            second_utf8 = second.to_s.force_encoding('UTF-8').encode('UTF-8', :invalid => :replace, :undef => :replace, :replace => '')
+            @value = Text.unnormalize(second_utf8)
           else
             $log.error "Encoding error: #{second_utf8}" if $log
           end
--- a/lib/gems/pending/util/xml/xml_utils.rb
+++ b/lib/gems/pending/util/xml/xml_utils.rb
@@ -62,18 +62,10 @@ class XmlFind
 end
 
 class XmlHelpers
-  # reference from REXML::TEXT
-  VALID_CHAR = [
-    0x9, 0xA, 0xD,
-    (0x20..0xD7FF),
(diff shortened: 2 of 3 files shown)


hsong-rh:
@roliveri @jerryk55 Please review.

lib/gems/pending/util/xml/miq_rexml.rb
```diff
@@ -148,7 +148,7 @@ def initialize(first, second = nil, parent = nil)
         rescue => err
           if err.class == ::Encoding::CompatibilityError
             second_utf8 = second.to_s.force_encoding('UTF-8')
```
roliveri: Shouldn't this just be changed to something like this:
```ruby
second_utf8 = second.to_s.encode('UTF-8', :invalid => :replace, :undef => :replace, :replace => "")
```
The string may not be UTF-8, so you don't want to force the encoding. You want to transcode the string to UTF-8, removing the characters that can't be transcoded.

lib/gems/pending/util/xml/xml_utils.rb
```diff
@@ -62,6 +62,12 @@ def self.findElementInt(paths, ele)
 end
 
 class XmlHelpers
+  # Validate attributes before inserting into xml
+  def self.validate_attrs(h)
+    return nil if h.nil?
+    h.inject({}) { |options, (key, value)| options[key.to_s] = value.to_s.force_encoding('UTF-8').valid_encoding? ? value : "Invalid encoding found"; options }
```
roliveri: Same here as above.

lib/gems/pending/util/xml/miq_rexml.rb
hsong-rh: I tested the above replacement. It didn't work for our case. The final result still has unacceptable chars. The test string I used is the 1st partition type: \xAF=\xC6\x0F\x83\x84rG\x8Ey=i\xD8G}\xE4.

irb(main)> "\xAF=\xC6\x0F\x83\x84rG\x8Ey=i\xD8G}\xE4".encode('UTF-8', :invalid => :replace, :undef => :replace, :replace => "")
=> "=\u000FrGy=iG}"

roliveri: What makes you say it contains unacceptable characters? If you transcoded the string as I suggested, the result has to be a valid UTF-8 string. Did `vaild_encoding?` on the resultant string return false?

hsong-rh: The valid_encoding? return true. But somehow we still got error message: incompatible encoding regexp match (UTF-8 regexp with ASCII-8BIT string) in evm.log

blomquisg:
@roliveri and @hsong-rh, this is for a G-release blocker bug.  Try to resolve the discussion quickly so we can get this merged and move the BZ off the blocker list.

roliveri:
@blomquisg What's the reasoning behind making this a blocker?
As it only happens on a subset of VMs.

hsong-rh:
@roliveri I made changes based on your comments. Please review.

lib/gems/pending/util/xml/xml_utils.rb
```diff
@@ -62,6 +62,30 @@ def self.findElementInt(paths, ele)
 end
 
 class XmlHelpers
+  # reference from REXML::TEXT
+  VALID_CHAR = [
```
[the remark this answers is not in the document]
Fryguy: Why are you making a copy of something that already exists? Prefer  `VALID_CHAR = REXML::Text::VALID_CHAR`  (Or even better just don't create an alias for it at all.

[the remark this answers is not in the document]
hsong-rh: @roliveri After encode with replacement, they are UTF-8 characters. But some inside chars are still marked as invalid in REXML::Text::check method call, when try to add the as element's attribute.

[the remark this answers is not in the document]
hsong-rh: @Fryguy I'll check to see if any existing codes can help us to do the same thing.

hsong-rh:
Both REXML and Nokegiri will raise exception in this case. They don't expect such control characters in XML data. I'll add a spec test for this change.

hsong-rh:
@roliveri @Fryguy Please review.

simaishi:
`gaprindashvili/yes` ?

roliveri:
@simaishi Sorry, Yes.

simaishi:

```
$ git log -1
commit d865ee63c000e355bb07ed2c4774be2ebe4cbd5b
Author: Richard Oliveri <oliveri.richard.github@gmail.com>
Date:   Fri Dec 15 10:21:24 2017 -0500

    Merge pull request #318 from hsong-rh/xml_encoding
    
    Add valid_encoding checking before inserting into xml object.
    (cherry picked from commit e04c65fce73dbeced56f66d724cdb79c69707573)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1530726
```

PR #319, opened 2017-12-12
Always use SSH in non-interactive mode

jhernand:
The default configuration of the `net/ssh` gem uses _interactive_ mode.
That means that when authentication with a provided password fails, the
gem will retry, discarding the provided password and asking
interactively for a new one. In the ManageIQ environment this will
always fail, because there is no human user to type that alternative
password. The result is the following error:

  Errno::ENOTTY: Inappropriate ioctl for device

That happens when the gem tries to read the password using the
`IO#noecho` method, because inside the ManageIQ process `$stdin` isn't
attached to a terminal.

The net result is an exception that isn't caught by the gem, and scary
error messages in the log and in the UI.

To avoid these problems this patch adds the `:non_interactive => true`
option to the constructor of the SSH client. This instructs the gem to
never retry failed authentications interactively.

This addresses the following bug:

  Password change for a Host fails with error code 500
  https://bugzilla.redhat.com/1516849

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -9,7 +9,12 @@ class MiqSshUtil
     @password = password
     @status   = 0
     @shell    = nil
-    @options  = {:password => @password, :remember_host => false, :verbose => :warn}.merge(options)
+    @options  = {
+      :password        => @password,
+      :remember_host   => false,
+      :verbose         => :warn,
+      :non_interactive => true,
+    }.merge(options)
 
     # Seems like in 2.9.2, there needs to be blank :keys, when we are passing private key as string
     @options[:keys] = [] if options[:key_data]


jhernand:
@miq-bot add_labels bug, gaprindashvili/yes

simaishi:

```
$ git log -1
commit 5c5d3db576a2bb0dfeb2a25259998f54cc92a417
Author: Adam Grare <agrare@redhat.com>
Date:   Tue Dec 12 13:13:14 2017 -0500

    Merge pull request #319 from jhernand/ssh_non_interactive
    
    Always use SSH in non-interactive mode
    (cherry picked from commit 6372d2351537779777fcde8ce4396c187175943e)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1526060
```

Fryguy:
cc @gmcculloug

PR #322, opened 2018-01-04
[WIP][FINE] Ensure uniq array for OpenStack pagination list

aufi:
Adding filtering on array which should ensure that list returned from Fog list
method contains unique objects (not repeated in a single collection).

### DIFF
--- a/lib/gems/pending/openstack/openstack_handle/pagination/marker.rb
+++ b/lib/gems/pending/openstack/openstack_handle/pagination/marker.rb
@@ -16,7 +16,7 @@ module OpenstackHandle
           all_objects.concat(objects_on_page)
         end
 
-        all_objects
+        all_objects.uniq(&:identity)
       end
 
       private


lib/gems/pending/openstack/openstack_handle/pagination/marker.rb
```diff
@@ -16,7 +16,7 @@ def list
           all_objects.concat(objects_on_page)
         end
 
-        all_objects
+        all_objects.uniq(&:identity)
```
aufi: This change is not the fix for any issue, duplication here seems unrelevant. Tested return
```all_objects + all_objects``` with no change (no duplicated records in the app).

Comment (Ladas) [reply lib/gems/pending/openstack/openstack_handle/pagination/marker.rb]: @aufi so, can you test this scenario: "fog it returns duplicates once, then it returns uniq"

That is what I think is happening. If parser returns duplicates, we should see it twice in the DB. When parser returns uniq, we disconnect 1 of the duplicates. In this case, only 1 of the duplicates is associated to the Service.

-------

the problematic part is https://github.com/Ladas/manageiq/blob/3ec67f4d8059a1846c7a707b7ad4f589ca6a8f14/app/models/ems_refresh/save_inventory.rb#L66
then https://github.com/Ladas/manageiq/blob/3ec67f4d8059a1846c7a707b7ad4f589ca6a8f14/app/models/ems_refresh/save_inventory.rb#L126

**Comment by agrare** [comment]
@aufi think this would be better fixed in the refresh_parser instead of messing with the fog return

File: lib/gems/pending/openstack/openstack_handle/pagination/marker.rb (hunk 1 above)
Comment (mansam) [reply lib/gems/pending/openstack/openstack_handle/pagination/marker.rb]: @Ladas I tested that scenario and it works the way you expected. Trimming out the duplicates from fog or discarding them during refresh/save works, but leaves any disconnected duplicates that have already been created.

Comment (Ladas) [reply lib/gems/pending/openstack/openstack_handle/pagination/marker.rb]: @mansam right, so we should provide a rails c scrips for cleanup of the duplicates (since we can't backport a migration). And we are preparing migrations for the unique indexes, so it there, we should think about cleaning up these duplicates

**Comment by agrare** [comment]
@aufi can you put this in to https://github.com/ManageIQ/manageiq-providers-openstack/blob/master/lib/manageiq/providers/openstack/legacy/openstack_handle/pagination/marker.rb first instead of directly to fine?

**Comment by ITD27M01** [comment]
Have checked this but got the error:

```
[----] E, [2018-01-26T17:32:33.305058 #25535:aa5140] ERROR -- : MIQ(ManageIQ::Providers::Openstack::NetworkManager::Refresher#refresh) EMS: [SPB Network Manager], id: [4] Refresh failed
[----] E, [2018-01-26T17:32:33.313279 #25535:aa5140] ERROR -- : [NoMethodError]: undefined method `list_security_groups' for nil:NilClass  Method:[rescue in block in refresh]
[----] E, [2018-01-26T17:32:33.313753 #25535:aa5140] ERROR -- : /opt/rh/cfme-gemset/gems/fog-openstack-0.1.20/lib/fog/network/openstack/models/security_groups.rb:19:in `all'
/opt/rh/cfme-gemset/gems/fog-core-1.44.3/lib/fog/core/collection.rb:112:in `lazy_load'
/opt/rh/cfme-gemset/gems/fog-core-1.44.3/lib/fog/core/collection.rb:16:in `each'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/refresh_parser_common/helper_methods.rb:9:in `process_collection'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/network_manager/refresh_parser.rb:128:in `get_security_groups'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/network_manager/refresh_parser.rb:37:in `ems_inv_to_hashes'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/network_manager/refresh_parser.rb:8:in `ems_inv_to_hashes'
/var/www/miq/vmdb/app/models/manageiq/providers/openstack/network_manager/refresher.rb:6:in `parse_legacy_inventory'
/var/www/miq/vmdb/app/models/ems_refresh/refreshers/ems_refresher_mixin.rb:146:in `block in parse_targeted_inventory'
```

For various objects not only for sec groups.

**Comment by Ladas** [comment]
@ITD27M01 is that on Fine, or Master? I suppose if it fails, it would fail the same in both though. @aufi have you tried to run all specs with this?

**Comment by ITD27M01** [comment]
@Ladas that is Fine. I have tried the following and it works:
```
uniq_all_objects = []
uniq_all_objects.concat(all_objects)
uniq_all_objects.uniq(&:identity)
```

**Comment by mansam** [comment]
I observed that issue as well while debugging; calling `uniq` on the fog collection causes it to lose certain references that makes subsequent iteration fail, though converting it to an array first works correctly.

**Comment by ITD27M01** [comment]
Another caveat about openstack pagination. This example show how duplicates occurred but also that there can be pagination loop:

https://gist.github.com/ITD27M01/f6b5c7732d8547a1485fccf05e51d172

**Comment by Ladas** [comment]
@ITD27M01 in your example, does the API also return a next_page pointer every time? 

In a rapidly changing env (deleting and adding new instances quickly), the current way could have problems. So we should find a better way. Also, I suppose the API got more consistent than how it was 2.5 years ago? Back than, the pagination was quite chaotic. :-) 

@mansam if we will change a pagination, we'll need to verify it works on all new OpenStack versions. The old VCR envs were set the way, that all services had harcoded limit 2 of resources, so that way we verified that pagination worked on all 4-5 OpenStack versions, on all API queries. There were nuances in different versions, like sometimes :next_page pointer was missing, or the pagination params were not even taken, etc.

**Comment by Ladas** [comment]
@ITD27M01 @mansam hm, though I think we also look for :rel, or any next page link. There is a method in for core

https://github.com/Ladas/manageiq-providers-openstack/blob/d27e835b6e2e2075a1cca992f36d84ddb9710aea/lib/manageiq/providers/openstack/legacy/openstack_handle/pagination/marker.rb#L25

the @service, that comes from fog e.g (Fog::Compute::OpenStack::Real) list class. I can't find the ```more_pages?``` implementation in fog now, though.

**Comment by ITD27M01** [comment]
@Ladas 

```
 def more_pages?(objects_on_page)
        marker(objects_on_page) && @service.more_pages?(objects_on_page)
end
```

This return true if marker exist because @service.more_pages? always true:

https://github.com/ManageIQ/manageiq-providers-openstack/blob/bab43c922977e8226815910dba1d7714fb0159d6/lib/manageiq/providers/openstack/legacy/openstack_handle/handled_list.rb#L71

I am using Newton now and here an example for API requests:

1. All instances by one request:
https://gist.github.com/ITD27M01/00681d4079d533f1f80a125259978f82
2. First page of limited queries (limit=1)
https://gist.github.com/ITD27M01/a62ecdef1aa3c86d434c39add26cda1c

You can see the following there:

```
"servers_links": [
    {
      "href": "http://testos.billing.ru:8774/v2.1/servers/detail?limit=1&all_tenants=1&marker=5ac85d0b-78f5-4bae-9415-721b4f960ffe",
      "rel": "next"
    }
```

3. Last page of limited queries
https://gist.github.com/ITD27M01/43018064154ea09e92a82290842a226d

Ladas:
@ITD27M01 yeah, just looking at that, it always goes to https://github.com/Ladas/manageiq-providers-openstack/blob/4ffefac87b852134a18df90280a8dff526ab95a0/lib/manageiq/providers/openstack/legacy/openstack_handle/handled_list.rb#L74

So the service specific readers of metadata were never finished :-(

Ladas:
@ITD27M01 right, I was wondering if you saw that having "rel": "next", could also lead to infinite loop. Unless OpenStack computes the page and next link in atomic query, it could always return that there is a next page (when we quickly add/remove vms).

We could simply make another guard, that we expect the same amount of rows for every page. Then the page having less is the last. (we can't base it on sent limit, since we do not know the limit, that is set on server)

aufi:
Closing in favor of https://github.com/ManageIQ/manageiq-providers-openstack/pull/212

PR #327, opened 2018-01-15
[WIP] Load linux_admin once when needed

NickLaMuro:
Due to some concerns about loading `linux_admin` multiple times when calling `MiqSystem.num_cpus` causing a memory leak, this creates a helper method that no-ops after it has been required once.

This is currently a [WIP] pull request as we verify further that calling `require` multiple times is leaking memory in the `manageiq` application, but putting this out there as a proposal for a quick fix.  Fixing require to not leak is far more ideal, but will take time.

Concerns
--------

Unsure if this will break autoloading in some way, though I doubt too many are trying to do a reload with in development `linux_admin`, or calling code where this would be executed and reloaded in development.  Regardless, something to consider when reviewing.

### DIFF
--- a/lib/gems/pending/util/miq-system.rb
+++ b/lib/gems/pending/util/miq-system.rb
@@ -47,7 +47,7 @@ class MiqSystem
 
   def self.num_cpus
     return unless Sys::Platform::IMPL == :linux
-    require 'linux_admin'
+    require_linux_admin
     @num_cpus ||= LinuxAdmin::Hardware.new.total_cores
   end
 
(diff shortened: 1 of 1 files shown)


lib/gems/pending/util/miq-system.rb
```diff
@@ -228,6 +228,12 @@ def self.open_browser(url)
     when :mingw, :mswin then `start "#{url.gsub('"', '""')}"`
     end
   end
+
+  def self.require_linux_admin
+    return true if @linux_admin_required
+    require_linux_admin
```
kbrock: think you mean
```ruby
require "linux_admin"
```

NickLaMuro: Derp.  Thanks for the catch.  Will fix.

NickLaMuro:
going to close this as https://github.com/ManageIQ/manageiq/pull/16837 is more all encompassing of a fix for the entire project.  Will re-open if it turns out that this is a more preferred/surgical/safer solution (doubt it, though).

PR #342, opened 2018-04-03
Drop legacy yml load of Hash subclasses with ivars

jrafanie:
Depends on:
- [x] https://github.com/ManageIQ/manageiq-gems-pending/pull/341

This was fixed in psych 2.0.9+
https://github.com/ruby/psych/commit/8f84ad0fc711a82a1040def861cb121e8985fd4c

Psych 2.0.9+ has been included in ruby starting with ruby 2.3.0

Since we don't support ruby 2.2 on master any longer, we can drop
support for this legacy loading of yaml dumped with ruby 2.2 and lower.


Fryguy:
@jrafanie What happens to existing database records?  Is that even something we have in the database?

jrafanie:
> @jrafanie What happens to existing database records? Is that even something we have in the database?

@Fryguy I'll do more digging into existing databases if I can find VimHash objects in serialized columns.  I don't know that we store them in serialized columns.

PR #344, opened 2018-04-11
Rerecord Disk Specs

jerryk55:
Rerecord miq_disk_spec recordings due to issue with updates to the psych gem for yaml1.2 causing
travis failures in the FINE release only.  

Need to use new image names also used by the manageiq-smartstate gem in new releases.

@roliveri @hsong-rh @jrafanie please review.  Thanks.


jrafanie:
Note, this change is causing the problem:
https://github.com/ruby/psych/commit/b737f0811a9687cf86f44f0a35f61cbde9eac673#diff-4529319a033c80b7ec4bbd1a05fe0412

It's unclear if there is a bug in the regex or if our cassettes aren't probably formatting the binary strings, but it was introduced in psych 2.1.0 and is still there.  Ruby 2.3.1 was the last ruby that shipped a psych without that change, 2.0.17.  

We should probably investigate the bug that introduced the change and see if we can fix it differently:
https://bugs.ruby-lang.org/issues/11988

simaishi:
@jrafanie ok to merge?

jerryk55:
Ok let me figure that out as well.

jrafanie:
Thanks @jerryk55

simaishi:
Thank you Jerry!

jerryk55:
So it looks like we can make the same changes to the spec file and then copy the recordings from the FINE branch into EUWE and it will all work.  Making it happen.

PR #355, opened 2018-06-26
Support S3 for DB Backups

jerryk55:
Initial commit to support DB backups to S3.

For RFE in https://bugzilla.redhat.com/show_bug.cgi?id=1513520

@roliveri please review.

Coincides with PRs
https://github.com/ManageIQ/manageiq-ui-classic/pull/4300
https://github.com/ManageIQ/manageiq-schema/pull/227
https://github.com/ManageIQ/manageiq/pull/17689

### DIFF
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -246,7 +246,7 @@ class MiqGenericMountSession
     res
   end
 
-  def add(source, dest_uri)
+  def add(source, dest_uri, _object_file = nil)
     log_header = "MIQ(#{self.class.name}-add)"
 
     logger.info("#{log_header} Source: [#{source}], Destination: [#{dest_uri}]...")
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -9,6 +9,7 @@ require 'util/miq-uuid'
 
 class MiqGenericMountSession
   require 'util/mount/miq_nfs_session'
+  require 'util/mount/miq_s3_session'
   require 'util/mount/miq_smb_session'
   require 'util/mount/miq_glusterfs_session'
 
@@ -53,7 +54,7 @@ class MiqGenericMountSession
 
   def self.new_session(opts)
     klass = uri_scheme_to_class(opts[:uri])
-    session = klass.new(:uri => opts[:uri], :username => opts[:username], :password => opts[:password])
+    session = klass.new(:uri => opts[:uri], :username => opts[:username], :password => opts[:password], :region => opts[:region])
     session.connect
     session
   end
@@ -62,6 +63,8 @@ class MiqGenericMountSession
     require 'uri'
     scheme, userinfo, host, port, registry, share, opaque, query, fragment = URI.split(URI.encode(uri))
     case scheme
+    when 's3'
+      MiqS3Session
     when 'smb'
       MiqSmbSession
     when 'nfs'
@@ -462,6 +465,10 @@ class MiqGenericMountSession
     else
       raise "platform not supported"
     end
+  end
+
+  def copy_dump_to_store
+    # nothing to do
   end
 
   private
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -246,7 +246,7 @@ class MiqGenericMountSession
     res
   end
 
-  def add(source, dest_uri, _object_file = nil)
+  def add(source, dest_uri)
     log_header = "MIQ(#{self.class.name}-add)"
 
     logger.info("#{log_header} Source: [#{source}], Destination: [#{dest_uri}]...")
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -54,7 +54,7 @@ class MiqGenericMountSession
 
   def self.new_session(opts)
     klass = uri_scheme_to_class(opts[:uri])
-    session = klass.new(:uri => opts[:uri], :username => opts[:username], :password => opts[:password], :region => opts[:region])
+    session = klass.new(opts.slice(:uri, :username, :password, :region))
     session.connect
     session
   end
@@ -465,10 +465,6 @@ class MiqGenericMountSession
     else
       raise "platform not supported"
     end
-  end
-
-  def copy_dump_to_store
-    # nothing to do
   end
 
   private
--- a/lib/gems/pending/util/mount/miq_s3_session.rb
+++ b/lib/gems/pending/util/mount/miq_s3_session.rb
@@ -18,7 +18,8 @@ class MiqS3Session < MiqGenericMountSession
   end
 
   def connect
-    _scheme, _userinfo, @host, _port, _registry, @mount_path, _opaque, _query, _fragment = URI.split(URI.encode(@settings[:uri]))
+    @host       = URI(uri).host
+    @mount_path = URI(uri).path
     super
   end
 
--- a/lib/gems/pending/util/mount/miq_s3_session.rb
+++ b/lib/gems/pending/util/mount/miq_s3_session.rb
@@ -2,7 +2,6 @@ require 'util/mount/miq_generic_mount_session'
 require 'aws-sdk'
 
 class MiqS3Session < MiqGenericMountSession
-
   def initialize(log_settings)
     log_header = "MIQ(#{self.class.name}-initialize)"
     logger.debug("#{log_header} initialize: log_settings are #{log_settings}")
@@ -15,7 +14,7 @@ class MiqS3Session < MiqGenericMountSession
   end
 
   def s3
-    @s3 ||= Aws::S3::Resource.new(region: @settings[:region], access_key_id: @settings[:username], secret_access_key: @settings[:password])
+    @s3 ||= Aws::S3::Resource.new(:region => @settings[:region], :access_key_id => @settings[:username], :secret_access_key => @settings[:password])
   end
 
   def connect
@@ -47,20 +46,18 @@ class MiqS3Session < MiqGenericMountSession
   end
 
   def copy_dump_to_store(database_opts, connect_opts)
-    begin
-      # Strip off "s3://" prefix to get the bucket name.
-      bucket_name = connect_opts[:uri].split(':')[1].split(/\//)[2]
-      if (dump_bucket = s3.bucket(bucket_name)).exists?
-        logger.debug("Found bucket #{bucket_name}")
-      else
-        logger.debug("Bucket #{bucket_name} does not exist, creating.")
-        dump_bucket.create
-      end
-      # write dump file to s3
-      logger.debug("Writing [#{database_opts[:local_file]}] to Bucket #{bucket_name}.")
-      dump_bucket.object(database_opts[:object_file]).upload_file(database_opts[:local_file])
-    rescue => err
-      logger.error("Error copying dump from temporary location to S3 object store #{err}")
+    # Strip off "s3://" prefix to get the bucket name.
+    bucket_name = connect_opts[:uri].split(':')[1].split(/\//)[2]
+    if (dump_bucket = s3.bucket(bucket_name)).exists?
+      logger.debug("Found bucket #{bucket_name}")
+    else
+      logger.debug("Bucket #{bucket_name} does not exist, creating.")
+      dump_bucket.create
     end
+    # write dump file to s3
+    logger.debug("Writing [#{database_opts[:local_file]}] to Bucket #{bucket_name}.")
+    dump_bucket.object(database_opts[:object_file]).upload_file(database_opts[:local_file])
+  rescue => err
+    logger.error("Error copying dump from temporary location to S3 object store #{err}")
   end
 end
--- a/lib/gems/pending/util/mount/miq_s3_session.rb
+++ b/lib/gems/pending/util/mount/miq_s3_session.rb
@@ -18,8 +18,7 @@ class MiqS3Session < MiqGenericMountSession
   end
 
   def connect
-    _scheme, _userinfo, @host, _port, _registry, @mount_root, _opaque, _query, _fragment = URI.split(URI.encode(@settings[:uri]))
-    @mount_path = @mount_root.split("/")[0..1].join("/")
+    _scheme, _userinfo, @host, _port, _registry, @mount_path, _opaque, _query, _fragment = URI.split(URI.encode(@settings[:uri]))
     super
   end
 
@@ -28,26 +27,19 @@ class MiqS3Session < MiqGenericMountSession
     FileUtils.rm_rf(@mnt_point) if File.exist?(@mnt_point)
   end
 
-  def mount_root
-    File.join(@mnt_point, (@mount_root.split("/") - @mount_path.split("/")))
-  end
-
   def uri_to_local_path(remote_file)
     # Strip off the leading "s3:/" from the URI"
-    log_header = "MIQ(#{self.class.name}-uri_to_local_path)"
-    logger.debug("#{log_header} remote_file is [#{remote_file}]")
-    file_part = remote_file.split(':')[1].split(/\//)[1..-1].join('/')
-    File.join(@mnt_point, file_part)
+    File.join(@mnt_point, URI(remote_file).host, URI(remote_file).path)
   end
 
   def uri_to_object_path(remote_file)
     # Strip off the leading "s3://" and the bucket name from the URI"
-    remote_file.split(':')[1].split(/\//)[3..-1].join('/')
+    # Also remove the leading delimiter.
+    URI(remote_file).path[1..-1]
   end
 
   def copy_dump_to_store(database_opts, connect_opts)
-    # Strip off "s3://" prefix to get the bucket name.
-    bucket_name = connect_opts[:uri].split(':')[1].split(/\//)[2]
+    bucket_name = URI(connect_opts[:uri]).host
     if (dump_bucket = s3.bucket(bucket_name)).exists?
       logger.debug("Found bucket #{bucket_name}")
     else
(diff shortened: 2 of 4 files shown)


carbonin:
@jerryk55 You shouldn't put `Fixes` in this PR as the bug will actually require multiple PRs to fix right?

We don't want the bot to move the BZ to `POST` after just this one is merged.

jerryk55:
@carbonin ah cool yup. got it.

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,63 @@
...
+    @s3 = s3
+  end
+
+  def s3
+    @s3 ||= Aws::S3::Resource.new(:region => @settings[:region], :access_key_id => @settings[:username], :secret_access_key => @settings[:password])
+  end
+
+  def connect
+    _scheme, _userinfo, @host, _port, _registry, @mount_root, _opaque, _query, _fragment = URI.split(URI.encode(@settings[:uri]))
+    @mount_path = @mount_root.split("/")[0..1].join("/")
```
roliveri: This seems strange. What is it doing?

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,63 @@
...
+    super
+  end
+
+  def disconnect
+    return if @mnt_point.nil?
+    FileUtils.rm_rf(@mnt_point) if File.exist?(@mnt_point)
+  end
+
+  def mount_root
+    File.join(@mnt_point, (@mount_root.split("/") - @mount_path.split("/")))
```
roliveri: When dealing with file paths, wouldn't it be better to use path-specific operations like `filename`, `basename`, or even the `Pathname` or `URI` classes - as opposed to splitting on "/"? The same applies to the other path manipulations in this file.

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: Removing the URI prefix.  (s3://).  The way the UI is implemented at this point the part of the prefix prior to the "://" is what determines the file_depot subclass. I'm not saying that's the best way...

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: Yes.  I couldn't come up with anything that did what I wanted so far so I did it from scratch.  I can go back and look again.

lib/gems/pending/util/mount/miq_s3_session.rb
roliveri: Then you should `split("://")` and take the second field - no need to join.
Here again, look into the `URI` facility, I think it will accomplish this more cleanly.

jerryk55: Yup.  I'm reworking this all using URI.

carbonin: Can I suggest `Pathname.new(@mount_root).each_filename.first`?

```ruby
irb(main):023:0> _scheme, _userinfo, @host, _port, _registry, @mount_root, _opaque, _query, _fragment = URI.split(URI.encode("s3://host/some/path/goes/here"))
=> ["s3", nil, "host", nil, nil, "/some/path/goes/here", nil, nil, nil]
irb(main):024:0> Pathname.new(@mount_root).each_filename.first
=> "some"
```

carbonin: I'm not sure that's exactly your desired behavior, but a spec would probable help too :smile:

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -464,6 +467,10 @@ def self.raw_disconnect(mnt_point)
     end
   end
 
+  def copy_dump_to_store
```
carbonin: Would it not be better to override `#add`/`#remove`/`#download` rather than creating a specific method for this which none of the other sub-classes will use?

jerryk55: Well, I'm not against overriding if it makes sense, but as far as none of the other sub-classes implementing this method don't I have to do the same thing for OpenStack Swift? And any other object-store implementation?

carbonin: Well, yeah, that's kind of what I'm saying. I feels like the API for an instance of `MiqGenericMountSession` is supposed to be `#add`/`#download`/ etc. so if we don't have an alternate implementation for add for object stores, it feels like this fits the bill.

The only reason I though about this was that `copy_dump_to_store` isn't really a good name, right? Because you could be copying any file to the store, not just a dump. And that got me thinking that this is really just the implementation of `#add` for things which don't actually "mount".

Does that make sense?

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,55 @@
...
+    raise "region is a required value!" if @settings[:region].nil?
+    @s3 = s3
+  end
+
+  def s3
+    @s3 ||= Aws::S3::Resource.new(:region => @settings[:region], :access_key_id => @settings[:username], :secret_access_key => @settings[:password])
+  end
+
+  def connect
+    _scheme, _userinfo, @host, _port, _registry, @mount_path, _opaque, _query, _fragment = URI.split(URI.encode(@settings[:uri]))
```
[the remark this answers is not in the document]
jerryk55: I can do that.  Note that the SMB code does exactly the same thing.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
jerryk55: @carbonin note that I've reworked this area somewhat to handle your request and some of the errors encountered.  I still have to push out the Core Repo changes although those might get modified based on UI changes as well.

carbonin: I think this can be removed now, right?

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+    File.join(@mnt_point, URI(remote_file).host, URI(remote_file).path)
+  end
+
+  def uri_to_object_path(remote_file)
+    # Strip off the leading "s3://" and the bucket name from the URI"
+    # Also remove the leading delimiter.
+    URI(remote_file).path[1..-1]
+  end
+
+  def add(local_file, uri, object_file)
```
carbonin: I'm not sure we need an additional parameter here.

The original implementation of `#add` uses the filename in the destination uri (second parameter) as the remote file name. Is there any reason we couldn't do the same here?

ref: https://github.com/ManageIQ/manageiq-gems-pending/pull/355/files#diff-633c63845bfb2f447bd88a0ff73bacfdR256

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+    @mount_path = URI(uri).path
+    super
+  end
+
+  def disconnect
+    return if @mnt_point.nil?
+    FileUtils.rm_rf(@mnt_point) if File.exist?(@mnt_point)
+  end
+
+  def uri_to_local_path(remote_file)
```
carbonin: Also I think adding tests for methods like this one and `uri_to_object_path` would make it much easier to understand what they were trying to do.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -53,7 +54,7 @@ def self.in_depot_session(opts, &_block)
 
   def self.new_session(opts)
     klass = uri_scheme_to_class(opts[:uri])
-    session = klass.new(:uri => opts[:uri], :username => opts[:username], :password => opts[:password])
+    session = klass.new(:uri => opts[:uri], :username => opts[:username], :password => opts[:password], :region => opts[:region])
```
Fryguy: Can we simplify this line to `session = klass.new(opts.slice(:uri. :username, :password, :region))`

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+class MiqS3Session < MiqGenericMountSession
+  def initialize(log_settings)
+    log_header = "MIQ(#{self.class.name}-initialize)"
+    logger.debug("#{log_header} initialize: log_settings are #{log_settings}")
+    super(log_settings)
+    logger.debug("#{log_header} initialize: @settings are #{@settings}")
+    raise "username is a required value!" if @settings[:username].nil?
+    raise "password is a required value!" if @settings[:password].nil?
+    raise "region is a required value!" if @settings[:region].nil?
+    @s3 = s3
```
Fryguy: I don't understand this line.  `@s3` is set in the s3 method already, so why set it again?  On top of that `@s3` is memoized on first access, so why do that if you are going to just set it during initialize anyway.

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+
+class MiqS3Session < MiqGenericMountSession
+  def initialize(log_settings)
+    log_header = "MIQ(#{self.class.name}-initialize)"
+    logger.debug("#{log_header} initialize: log_settings are #{log_settings}")
+    super(log_settings)
+    logger.debug("#{log_header} initialize: @settings are #{@settings}")
+    raise "username is a required value!" if @settings[:username].nil?
+    raise "password is a required value!" if @settings[:password].nil?
+    raise "region is a required value!" if @settings[:region].nil?
```
Fryguy: Minor, but you could collapse this into a single raise to make it easier on the caller if they miss more than one key.

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+    @s3 = s3
+  end
+
+  def s3
```
Fryguy: Should this be a private method (as well as other methods that don't implement the public API)?

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
+require 'util/mount/miq_generic_mount_session'
+require 'aws-sdk'
```
Fryguy: Can we delay load this?

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: I'll take a look.

jerryk55: The difference is that the S3 bucketname would be the first part of the URI, but I'm also using the rest of the URI as the "directory-like" key/prefix to the the filename.  And I'm doing the same thing with the filename that is done for Samba and NFS - using the file path that is created by the actual database dump.  So for instance, if we were using NFS, the file would get copied into a directory tree named "db_backup/region_0/__adhoc_dbbackup_2018_07_06_13_41_23_0400"  underneath the NFS mount point.

carbonin: I'm not sure I understand this. In the PR which uses this class it looks like `object_file` here just ends up being the result of `#uri_to_object_file` [here](https://github.com/ManageIQ/manageiq/pull/17689#discussion_r202808292). In my mind you could simplify this by removing this parameter and extracting all the information you need from the uri inside the processing done in this class.

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,66 @@
...
+    raise "region is a required value!" if @settings[:region].nil?
+    @s3 = s3
+  end
+
+  def s3
+    @s3 ||= Aws::S3::Resource.new(:region => @settings[:region], :access_key_id => @settings[:username], :secret_access_key => @settings[:password])
+  end
+
+  def connect
+    @host       = URI(uri).host
```
carbonin: Why even bother having content here? I think it would be more clear to set `@host` and `@mount_path` on initialize and  leave this method with no content with a comment that explains that you don't really have to "connect" to the object store.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -243,7 +246,7 @@ def verify
     res
   end
 
-  def add(source, dest_uri)
+  def add(source, dest_uri, _object_file = nil)
```
carbonin: Can this unused parameter be removed?

jerryk55: I could swear I did that.  Obviously missed it.  Will do...

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,70 @@
...
+    super(log_settings)
+    logger.debug("#{log_header} initialize: @settings are #{@settings}")
+    # NOTE: This line to be removed once manageiq-ui-class region change implemented.
+    @settings[:region] = "us-east-1" if @settings[:region].nil?
+    raise "username, password, and region are required values!" if @settings[:username].nil? || @settings[:password].nil? || @settings[:region].nil?
+    @host       = URI(@settings[:uri]).host
+    @mount_path = URI(@settings[:uri]).path
+  end
+
+  def connect
```
carbonin: Shouldn't we just not define this method and let it be inherited from the base class if we're just calling `super`?

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -0,0 +1,70 @@
...
+    # No actual "connection" to S3 required here.
+    super
+  end
+
+  def disconnect
```
carbonin: Should this be `self.raw_disconnect`? It looks like the parent `.disconnect` method does some logging that we may want.

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: Here's your comment from 11 days ago (below) - "Why even bother having content here? I think it would be more clear to set @host and @mount_path on initialize and leave this method with no content with a comment that explains that you don't really have to "connect" to the object store."

carbonin: Hm, past me has a point ... can we remove the `super` call then? If we don't need it at all why call into the base class?

carbonin:
@roliveri This one looks pretty close, I think you're comments were addressed. Do you want to take another look?

lib/gems/pending/util/mount/miq_s3_session.rb
carbonin: Why do we need `super`? All `MiqGenericMountSession#connect` does is set `@mount_point` which, like you (and I) said, you're doing in `initialize`.

It sets it directly [here](https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/mount/miq_generic_mount_session.rb#L129) then in a separate method [here](https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/mount/miq_generic_mount_session.rb#L78) wouldn't that overwrite what you're doing in `initialize`?

carbonin: Ah, I see. Different variables (`mount_path` vs `mnt_point`). Well if this works then a method with just super in it is the same as the method not being present at all so we can remove it I guess.

carbonin: I still think this method should be removed. Do you think there's any reason to have a method with just `super` as the content @jerryk55?

spec/util/mount/miq_s3_session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(described_class).to receive(:`).with("sudo #{cmd} 2>&1")
+    expect($CHILD_STATUS).to receive(:exitstatus).once.and_return(1)
+
+    described_class.runcmd(cmd)
+  end
+
+  it "#@mnt_point starts with '/tmp/miq_'" do
+    result = @session.mnt_point
+    expect(result).to     be_kind_of(String)
+    expect(result).to_not be_blank
```
carbonin: Can we remove these two expectations in all these specs? Anything matching the regex will certainly be both a string and not blank.

spec/util/mount/miq_s3_session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(result).to     match(/^\/tmp\/miq_.*\/tmp\/abc\/def$/)
+  end
+
+  it "#uri_to_object_path returns a new object path" do
+    result = @session.uri_to_object_path(@uri)
+    expect(result).to     be_kind_of(String)
+    expect(result).to_not be_blank
+    expect(result).to     match(/^abc\/def$/)
```
carbonin: Can this be `expect(result).to eq("/abc/def")`?

spec/util/mount/miq_s3_session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect(result).to     be_kind_of(String)
+    expect(result).to_not be_blank
+    expect(result).to     match(/\/tmp\/miq_/)
+  end
+
+  it "#uri_to_local_path returns a new local path" do
+    result = @session.uri_to_local_path(@uri)
+    expect(result).to     be_kind_of(String)
+    expect(result).to_not be_blank
+    expect(result).to     match(/^\/tmp\/miq_.*\/tmp\/abc\/def$/)
```
carbonin: If we can know exactly what this string would be I would rather use an `eq` matcher here. It makes the desired result more understandable.

spec/util/mount/miq_s3_session_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    expect($CHILD_STATUS).to receive(:exitstatus).once.and_return(1)
+
+    described_class.runcmd(cmd)
+  end
+
+  it "#@mnt_point starts with '/tmp/miq_'" do
+    result = @session.mnt_point
+    expect(result).to     be_kind_of(String)
+    expect(result).to_not be_blank
+    expect(result).to     match(/\/tmp\/miq_/)
```
carbonin: I think `expect(result).to start_with("/tmp/miq_")` would be more readable here (especially considering the test name).

spec/util/mount/miq_s3_session_spec.rb
jerryk55: We can't.  its a generated string after the "miq_"

spec/util/mount/miq_s3_session_spec.rb
jerryk55: ok.

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: No other than your original comment.  I'll take it out...

PR #357, opened 2018-07-26
S3 db restore

jerryk55:
Part of the continuing saga of the RFE to allow backups/restores from AWS S3 and Openstack Swift.
This enables downloading the backup file from S3 prior to running the restore command.

Other PRs in this series include 
https://github.com/ManageIQ/manageiq-ui-classic/pull/4300
https://github.com/ManageIQ/manageiq/pull/17689
https://github.com/ManageIQ/manageiq-schema
https://github.com/ManageIQ/manageiq-gems-pending

One of several PRs required for RFE in https://bugzilla.redhat.com/show_bug.cgi?id=1513520
@roliveri @carbonin please review.

### DIFF
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -187,14 +187,6 @@ class MiqGenericMountSession
     connect
   end
 
-  def self.supports_objects?
-    false
-  end
-
-  def supports_objects?
-    false
-  end
-
   def with_test_file(&_block)
     raise "requires a block" unless block_given?
     file = '/tmp/miq_verify_test_file'
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -187,6 +187,14 @@ class MiqGenericMountSession
     connect
   end
 
+  def self.supports_objects?
+    false
+  end
+
+  def supports_objects?
+    false
+  end
+
   def with_test_file(&_block)
     raise "requires a block" unless block_given?
     file = '/tmp/miq_verify_test_file'
--- a/lib/gems/pending/util/mount/miq_s3_session.rb
+++ b/lib/gems/pending/util/mount/miq_s3_session.rb
@@ -43,7 +43,7 @@ class MiqS3Session < MiqGenericMountSession
     logger.debug("Writing [#{local_file}] to Bucket [#{bucket_name}] using object file name [#{object_file}]")
     begin
       dump_bucket.object(object_file).upload_file(local_file)
-    rescue Aws::S3::Errors::AccessDenied => err
+    rescue Aws::S3::Errors::AccessDenied, Aws::S3::Errors::Forbidden => err
       disconnect
       logger.error("Access to S3 bucket #{bucket_name} restricted.  Try a different name. #{err}")
       msg = "Access to S3 bucket #{bucket_name} restricted.  Try a different name. #{err}"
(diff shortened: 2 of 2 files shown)


carbonin:
@jerryk55 It looks like this still contains some of the old commits from the backup PR, can you rebase this so it's just the restore commits since the backup PR has been merged?

jerryk55:
@carbonin I'm on it..

jerryk55:
@carbonin Rebased - just shows the one commit now.

lib/gems/pending/util/mount/miq_s3_session.rb
```diff
@@ -56,6 +56,38 @@ def add(local_file, uri)
...
+    rescue => err
+      disconnect
+      logger.error("Error downloading #{remote_file} from S3. #{err}")
+      msg = "Error downloading #{remote_file} from S3. #{err}"
+      raise err, msg, err.backtrace
+    end
+    local_file
+  end
+
+  def supports_objects?
```
roliveri: @jerryk55 should there be a class `supports_objects?` as well?

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -187,6 +187,14 @@ def reconnect!
     connect
   end
 
+  def self.supports_objects?
```
carbonin: I thought we decided that we didn't need these methods and would use logic more like [this](https://github.com/ManageIQ/manageiq/pull/17689/files#diff-2824c34710b68cc8ae0f96cec49c2a2aR122)?

jerryk55: I'm not sure about *we* deciding that... :-)  You made that suggestion this morning on the other PR I believe.  My issue with that is we will definitely also need it for the Openstack Swift subclass as well.  And whatever other object-store based subclass comes up.  So that means checking on a class-by-class basis, which gets ugly.  I much prefer having a class method that controls the functionality.

lib/gems/pending/util/mount/miq_s3_session.rb
jerryk55: @roliveri pending the discussion happening one box below this one, I would think so, yes.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
carbonin: I guess I'm not sure what's stopping us from treating all the mount sessions the same. Since they all respond to `#add` I feel like we should be able to use that method no matter what subclass we're dealing with. That's the point of this kind of duck-typing right?

jerryk55: So my take on it is this - I'm not imposing any functionality on NFS or SMB or FTP or whatever other non-object-store based session we might have here.  There is an existing MiqGenericMountSession add method that does a bunch of stuff already.  I'm not sure what calls that but it certainly wasn't previously called during the backup session by the EvmDatabaseOps class.  If I don't test for an object-store based subclass in some way or another every backup for NFS and SMB, etc will invoke that add method.  They shouldn't.

carbonin: > every backup for NFS and SMB, etc will invoke that add method. They shouldn't.

I'm not sure that's true. After a quick look it seems like the `MiqGenericMountSession#add` method puts the file in the specified path relative to the mount point.

In the backup code it looks like we're doing that incidentally by virtue of being able to determine the mount point as well as the destination path. It feels like using the `#add` method would be a no-op for the other mount session types and might actually simplify adding new backup targets like FTP.

Additionally, I think not having to know what type of session you're using will really clean up the implementation here so I think it would be an overall win.

jerryk55: Ok and for the restore situation - which is the next PR which I haven't even had the chance to push out yet - EvmDatabaseOps will call the session download method - same issue - can we do that on non-object-store session types?

carbonin: I would hope so. That would make sense to me.

jerryk55: Just so everyone reading this knows - making these changes will require going back to manageiq-gems-pending and modifying the add (and probably download) methods for the MiqGenericMountSession class.

carbonin: In what way? What changes are necessary?

carbonin: Aren't we overriding that completely in the s3 mount session class?

jerryk55:
@carbonin this one should be ready for review.  Restore side of gems-pending.  Will be pushing the restore side of manageiq next.

PR #358, opened 2018-08-01
Fix issue caused by missing source file in add call

jerryk55:
Changes to EvmDatabaseOps for PR https://github.com/ManageIQ/manageiq/pull/17689
based on review comments in that PR
will require that the session add method be called on all sessions.  This allows
us to fail gracefully and fixes a spec test in THAT PR.

@carbonin and @roliveri  your proverbial can of worms.  Please review and merge this so we can move forward.  Thanks.


carbonin:
In what case would we try to add a file when the source doesn't exist? Can you link to the source or spec that tries to do this?

jerryk55:
You asked me to change the yield to always pass the session back from with_mount_session.  Now add is always called in backup.  So now the add is always called even when it wasn't before.  I tried to explain this to you and you said we were better off with the behavior this way.

PR #360, opened 2018-08-07
Adds MiqFtpLib

NickLaMuro:
Helper module for common FTP interactions that will be shared in multiple parts of the `manageiq` codebase.

Lifted from `FileDepotFtp`, and will eventually replace some of the methods there (follow up PR).

Links
-----

* Originally extracted in https://github.com/ManageIQ/manageiq/pull/17549


NickLaMuro:
@carbonin So this is just a portion of https://github.com/ManageIQ/manageiq/pull/17549 that I eventually want to pull into this repo.  I will eventually be making a `MiqFtpUploadTarget` class in this repo and some other classes to support what we discussed on Friday.

But putting this together now since it will require a follow up PR and was taking a bit of a break from main effort of getting a decent splitting interface in place around the `MiqGenericMountSession` and other upload targets.  This will probably make more sense once that interface is more than a thought in my head (and an actual PR), but feel free to ask any questions you have about it now if needed and I can provide more context for you.

cc @jerryk55 @Fryguy

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,70 @@
...
+# Will setup a `@ftp` attr_accessor to be used as the return value for
+# `.connect`, the main method being provided in this class.  Will also setup
+# logging if not already done for the particular class (that follows the
+# VmdbLogger conventions, and setup a `uri` attr_accessor if that doesn't
+# already exist.
+module MiqFtpLib
+  def self.included(klass)
+    klass.send(:attr_accessor, :ftp)
+
+    klass.send(:attr_accessor, :uri) unless klass.instance_methods.include?(:uri=)
```
carbonin: At first glance it feels like there should be a better way to check if there is already an `attr_accessor` for this. If nothing else maybe check for both the `#uri=` and `#uri` methods?

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    unless klass.instance_methods.include?(:_log)
+      klass.send(:define_method, :_log) do
+        self.class.instance_logger
```
carbonin: This makes me feel like this would be better as a class so you could assign a logger rather than a module where we need to assume some methods exist.

I'm not sure of the full direction for this so maybe it makes more sense with the full picture, but I still think it's worth considering.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: Yeah, I thought so too.  Again, this is a straight rip off of what I did in https://github.com/ManageIQ/manageiq/pull/17549 (simply changed and relocated the module), so I didn't change anything there.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: > I'm not sure of the full direction for this so maybe it makes more sense with the full picture, but I still think it's worth considering.

Okay, I have been distracted between a few things, and this response has been half completed on my computer for about 3 hours, I will give a "brief" summary of what I am planning:

* I am in the process of creating a `MiqFileStorage` and `MiqFileStorage::Interface` which is meant to be the base class for `MiqGenericMountSession` classes, and what I am currently going to call "`MiqUploadTargets`" (encompassing `FTP`, `s3`, `Swift`, etc.).
* `MiqFileStorage::Interface` will be in charge of defining top level methods and underlying implementation methods can be overwritten by subclasses per implementation (where the input from a IO stream goes when writing to a split file)
* `MiqFileStorage` will be roughly what `with_mount_session` is at the moment in `EvmDatabaseOps` in `ManageIQ/manageiq`.
* I plan on putting `FTP` and `s3` integrations under a `UploadTarget` (working title), which will be managed by `MiqFileStorage`, but conform to the `MiqFileStorage::Interface`
* the `FtpUploadTarget` will want to probably share the connection logic of `FileDepotFtp`, and I see this as the easiest way to do it.

As far as the specifics in your question:

> This makes me feel like this would be better as a class so you could assign a logger rather than a module where we need to assume some methods exist.

I am basically trying to keep `FileDepotFtp` in as much of it's current state without changing it, while being able to share it's logic without calling out to it directly.  The `_log` statements are part of that issue, and this is the best way I could think of solving it.  The `instance_logger` is the exact same interface that is used under the covers for `VmdbLogger` when defining `_log`, so this is basically me ripping that implementation off.

This most of the `if` logic here is to account for classes that are not `FileDepotFtp`, and not the other way around.

carbonin:
Live by the :book: die by the :book: I guess ...

> - I am in the process of creating a MiqFileStorage and MiqFileStorage::Interface which is meant to be the base class for MiqGenericMountSession classes, and what I am currently going to call "MiqUploadTargets" (encompassing FTP, s3, Swift, etc.).

Is the plan here to replace `MiqGenereicMountSession` or are you creating an abstraction that `MiqGenericMountSession` will also inherit from? I think some kind of hierarchy diagram might be useful in this case.

> - MiqFileStorage::Interface will be in charge of defining top level methods and underlying implementation methods can be overwritten by subclasses per implementation (where the input from a IO stream goes when writing to a split file)

Why make the interface a separate module in the namespace? I think I would rather have the `UploadTarget` base class define the interface if that's what the existing (and future) classes will be inheriting from.

> - MiqFileStorage will be roughly what with_mount_session is at the moment in EvmDatabaseOps in ManageIQ/manageiq.

I'm not sure what this means. Will `MiqFileStorage` have a class method that instantiates an `UploadTarget` (previously mount session) and yield it? That sounds fine to me.

> - I plan on putting FTP and s3 integrations under a UploadTarget (working title), which will be managed by MiqFileStorage, but conform to the MiqFileStorage::Interface

Okay, so `MiqFileStorage` will create instances of `UploadTargets`? Is the plan to do something similar to what `MiqGenericMountSession.new_session` does now? That sounds good, but see my previous comment about the interface living in the `UploadTarget` base class rather than in a module of `MiqFileStorage`

> - the FtpUploadTarget will want to probably share the connection logic of FileDepotFtp, and I see this as the easiest way to do it.

Makes sense to me. I was just trying to simplify the logging bit.

---------------------------------------------------------------------------------------------

So as I understand it the design (which is really what this conversation ended up being) sounds like this.

- `UploadTarget`: Base class of current `MiqGenericMountSession`s as well as new non-mountable storage things.
- `MiqFileStorage`: Factory of `UploadTarget`s which sounds like it will be a mashup of the current `MiqGenericMountSession.new_session` method and `EvmDatabaseOps.with_mount_session`
- `FtpUploadTarget`: subclass of `UploadTarget`, but not `MiqGenericMountSession`

This all sounds good to me. The only changes/refinements I would suggest are pushing the `new_session` factory type logic into `UploadTarget.new` (maybe adopting the null object pattern if it makes sense, I'm thinking something like the way we implemented [pdf generation](https://github.com/ManageIQ/manageiq/blob/master/lib/pdf_generator.rb) or [network interfaces](https://github.com/ManageIQ/linux_admin/blob/master/lib/linux_admin/network_interface.rb) and maybe spelling out the interface you have planned for `UploadTarget` before doing all the work.

@NickLaMuro @Fryguy If you got this far does this make sense? Should we put this conversation somewhere else?

lib/gems/pending/util/miq_ftp_lib.rb
carbonin: I'm going to move the general design discussion to a regular (non-code specific) comment

NickLaMuro:
> @NickLaMuro @Fryguy If you got this far does this make sense? Should we put this conversation somewhere else?

For those following along at home, the answer was "yes", and "we did".  Most of what we (@carbonin and myself) discussed out band should answer the questions above in the new PR (when I get something concrete to share), so I will save the summary for then, and try to remember to link back to here later (if I remember).

NickLaMuro:
@carbonin Do you have any reservations about what I am doing here still?  I realize that [your comment from before](https://github.com/ManageIQ/manageiq-gems-pending/pull/360#discussion_r208216165) and the other one (I marked that one as resolved just because it got off topic a bit) are somewhat still as they were.

https://github.com/ManageIQ/manageiq-gems-pending/pull/368 , however, now can provide some clarity (built off of https://github.com/ManageIQ/manageiq-gems-pending/pull/361 currently), so if there was any curiosity as to how this would be used, that would be the intended implementation.

* * *

But to the comments you had previously (don't read if you are just going to merge anywy... or hate :books:):

* https://github.com/ManageIQ/manageiq-gems-pending/pull/360#discussion_r208216165 : In `FileDepotFtp` (well, basically `FileDepot`), `uri` comes from `ActiveRecord` as it is a field in the DB, so that check is necessary.  The other option would be to not have this module implement `uri` as an accessor, and leave that to the including class to do.  We could even raise an error for good measure to make sure they conform to the mixin's interface.
* https://github.com/ManageIQ/manageiq-gems-pending/pull/360#discussion_r208218554 : I don't really have a good solution for this.  What was there was carry over from the logging from `FileDepotFtp` (which at least one person must have found useful), and that is all cobbled together from the coupled, but disconnected `Vmdb::Logger` and `Vmdb::Logging` that we have (a source of constant frustration for me).  I feel like that is a long winded refactoring issue that I don't want to address now, but is why that portion of this code smells so bad... :poop:.

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,84 @@
...
+require 'logger'
+
+# Helper methods for net/ftp based classes and files.
+#
+# Will setup a `@ftp` attr_accessor to be used as the return value for
+# `.connect`, the main method being provided in this class.  Will also setup
+# logging if not already done for the particular class (that follows the
+# VmdbLogger conventions, and setup a `uri` attr_accessor if that doesn't
+# already exist.
+module MiqFtpLib
```
carbonin: I still think we should make this a class around FTP interaction with its own logger. Then we also don't have to be responsible for what methods are defined on the model (or other thing) using this class.

spec/util/miq_ftp_lib_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+
+    it "creates a _log method if necessary" do
+      expect(FTPKlass).to receive(:instance_logger).once
+      subject._log
+    end
+
+    context "when the class already has a uri=" do
+      subject { OtherFTPKlass.new }
+
+      it "has a `uri` accessor" do
```
jrafanie: yikes backticks...

spec/util/miq_ftp_lib_spec.rb
```diff
@@ -0,0 +1,216 @@
...
+
+  describe "when included" do
+    it "has a `ftp` accessor" do
+      ftp_instance = Net::FTP.new
+      subject.ftp  = ftp_instance
+
+      expect(subject.ftp).to eq ftp_instance
+    end
+
+    it "has a `uri` accessor" do
```
jrafanie: single quote instead of backtick?

lib/gems/pending/util/miq_ftp_lib.rb
jrafanie: Yeah, can we simplify this.  Conditional attr_accessors feels complicated.  I don't understand why it's conditional.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: @carbonin is this in response to this:

https://github.com/ManageIQ/manageiq-gems-pending/pull/360#issuecomment-417097055

or a comment on it's own?

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,84 @@
+require 'net/ftp'
+require 'logger'
+
+# Helper methods for net/ftp based classes and files.
+#
+# Will setup a `@ftp` attr_accessor to be used as the return value for
```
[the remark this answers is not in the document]
NickLaMuro: I am treating backticks (here, and in your other comments) as you would in markdown to represent a portion of `code` with that is being talked about inside of the comment/description.

Backticks inside a string have no execution, and are simply there for emphasis for specific identifiers.

lib/gems/pending/util/miq_ftp_lib.rb
carbonin: I think it's kind of both. I read that response a few days ago then re-read the whole mixin here and I'm still not comfortable with this being an included module.

I think there would be more value and stability in making this a standalone class than trying to fit this in as a mixin. The refactoring you would have to do in FileDepotFtp seems minimal and I think it would make this class *much* more stable and readable.

TLDR;
Make it a class, take a uri as a parameter to initialize, and give it a `#logger=` method with a null logger as a default. That should take care of all of the meta-programming in `.included` here.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: @jrafanie Sorry, missed this when looking at this earlier, but please look at https://github.com/ManageIQ/manageiq-gems-pending/pull/360#issuecomment-417097055 for clarity on to why this is the way.  Specifically the first bullet point.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: > I think there would be more value and stability...

Sure, maybe long term.  But as a reminder, this came about from your original comment here:

https://github.com/ManageIQ/manageiq/pull/17549#issuecomment-398062438

In which I was trying to meet halfway in regards to making this DRY.  But, to your comment that:

> The refactoring you would have to do in FileDepotFtp seems minimal...

I disagree entirely.  These methods are _**EXACTLY**_ as they are in `FileDepotFtp`, meaning there is next to zero risk on me doing what I originally had done in there as part of https://github.com/ManageIQ/manageiq/pull/17549 which was gut these methods from `FileDepotFTP` and replace them via `include MiqFtpLib` (previously [here](https://github.com/ManageIQ/manageiq/pull/17549/files#diff-aca27294178e2b711e7629bacef08fcbR4)).

With this suggestion, it is a re-write of `FileDepotFtp` effectively, because half of the methods I am not using here still have their own `_log` method that I would need to handle (or accept disparaging loggers for).

Also, `@ftp` is no longer part of the class, as it would be a reference to the new class, and so everything that isn't a part of `MiqFtpLib` as it is now that isn't removed from `FileDepotFtp` basically going to be re-written as a result.  The "specs" for `FileDepotFtp` (which is about 90% mocking, and 10% actually testing things), would also probably completely need to be re-written as a result.



* * * 

tl; dr  No, this would not be easier to write this as a class.  Unless you can convince "the powers that be" that this is worth said re-write that I didn't estimate for, I disagree with the suggestion.

Again [[ref]](https://github.com/ManageIQ/manageiq-gems-pending/pull/360#issuecomment-417097055), if the meta-programming is an issue, I can basically remove most of the `included` block (if not all of it), and just assume this is an "interface" module, and those methods inside of `included` can then required by the "including" class.  This goes for both the `uri` bit and the `_log`.

NickLaMuro: Done.  I kept `attr_accessor :ftp` in since it has next to no meta-programming, but still tries to reach peak minimum LOC added, to appease @jrafanie ...

lib/gems/pending/util/miq_ftp_lib.rb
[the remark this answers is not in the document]
Fryguy: yard accepts markdown (and thus backticks) for comments, I believe

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,76 @@
...
+# `.connect`, the main method being provided in this class.  Will also setup
+# logging if not already done for the particular class (that follows the
+# VmdbLogger conventions, and setup a `uri` attr_accessor if that doesn't
```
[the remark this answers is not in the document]
NickLaMuro: Yup!  Good catch.

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -0,0 +1,76 @@
+require 'net/ftp'
+require 'logger'
```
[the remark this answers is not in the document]
NickLaMuro: 👍

lib/gems/pending/util/miq_ftp_lib.rb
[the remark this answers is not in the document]
NickLaMuro: @Fryguy Yes, and so does just vanilla RDoc shipped with Ruby:

http://ruby-doc.org/stdlib-2.4.1/libdoc/rdoc/rdoc/RDoc/Markup.html#class-RDoc::Markup-label-Supported+Formats

NickLaMuro:
@carbonin Just an FYI about how I "addressed" [this comment](https://github.com/ManageIQ/manageiq-gems-pending/pull/360/files/afa483213714acbec170d1d0f77fd97fb4b91d2f#diff-642ef29479a149bae7a6f164f7af0326):

I am still requiring 'logger' in the specs in the off chance it hasn't been included by something else.  In classes that include this, I will also be including it, again, just as a safety measure.

Example in a later PR:  https://github.com/NickLaMuro/manageiq-gems-pending/compare/ftp_object_storage~2...ftp_object_storage#diff-4cf23bcfce6b36adcd0045ffbd1b724e

carbonin:
Ehh, I'm not sure you can :laughing:
image

NickLaMuro:
@carbonin Yeah, looks like GHFM (GitHub Flavored Markdown) f'd with their own link:

```
https://github.com/NickLaMuro/manageiq-gems-pending/compare/ftp_object_storage~2...ftp_object_storage#diff-4cf23bcfce6b36adcd0045ffbd1b724e
```

That **DOES** work, but got fudged when it tried parsing out the `#` part in it...

... so I guess TIL on that as well...

PR #361, opened 2018-08-11
MiqFileStorage interface and subclassing (with file splitting)

NickLaMuro:
Update:  Currently built off of https://github.com/ManageIQ/manageiq-gems-pending/pull/361 (will be implementing with that code soon)
Update:  Also built off of https://github.com/ManageIQ/manageiq-gems-pending/pull/365.  #361 might not stay as part of this PR, but currently including it here as well.
Update:  Rebased in a lot of the `[WIP]`/`[FIXUP]` commits, and removed #361 from this PR (will open up a separate PR to implement FTP).

This introduces a higher level abstraction above `MiqGenericMountSession` which allows for a common API between mounted file systems, and what I have called in this branch name `UploadTarget`s (might renamed to `ObjectStorage`).

Currently Implemented
---------------------

* `MiqFileStorage` and `MiqFileStorage::Interface`
  - `MiqFileStorage` will be what is actually interacted with by the user (top level API still in the works)
  - `MiqFileStorage::Interface` is what is the super class for `MiqGenericMountSession` and `MiqObjectStore`
* Have `MiqGenericMountSession` inherit and conform to the `MiqFileStorage::Interface`
* Introduced `MiqLocalMountSession`, which allows local files to also use this interface
* File splitting support with `MiqFileStorage::Interface` classes

TODO
----

Possibly these will end up as follow up PRs, but regardless, the remaining TODO items for this repo is the following:

* [x] Add `MiqObjectStore` subclass
* [x] Move `MiqS3Session` to live under `MiqObjectStore` (`MiqS3ObjectStore` maybe?)

Decided to skip doing the following:

* [x] ~~Implement the high level interface for `MiqFileStorage` (`.add`, `.download`, etc.)~~
* [x] ~~Add `MiqFtpObjectStore`~~


NickLaMuro:
@carbonin @Fryguy Sorry for the delay on getting this out, but this is what I got done so far.  Will most likely be tackling the FTP stuff next.  Still very much `[WIP]`, but wouldn't mind your eyes and 🍅 's.

Will add some comments at some point soon around the tricky area's.

cc @jerryk55 @dmetzger57 @blomquisg

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,177 @@
...
+        puts "  @source_input: #{@source_input.inspect}"
+        puts "  @input_writer: #{@input_writer.inspect}"
+      else
+        source        = upload_args.pop
+        @source_input = source.kind_of?(IO) ? source : File.open(source, "r")
+      end
+    end
+
+    def parse_byte_value(bytes)
+      match = bytes.to_s.match(BYTE_HASH_MATCH) || return
```
carbonin: This seems a bit fragile for my tastes. I would rather go with a straight options hash or kwargs rather than determining the meaning of a parameter based on a home-grown regex.

NickLaMuro: Oh, sorry, should I have implemented the fun case logic that is included as part of `xstrtol.c`:

https://github.com/coreutils/gnulib/blob/5ff4c9ba8b2002538380e3930dfc6b0bd5f4b297/lib/xstrtol.c

Which is what is used by `coreutils` in GNU `split` to parse the `--bytes` flag?

...

:trollface: 

(I may or may not have went on a lengthy hunt to find how this was done in between travis runs just for the sake of this troll...)

* * *

Okay, but for the real ( 📖 ) answer, I assume parsing `"200M"` instead of `209715200` is not what your beef is with, but the implementation of the "`:call_sig:`" of `#add`, correct?

So, I am not married to this implementation, but the one reason I choose it was to avoid having to change every instance where `MiqGenericMountSession` (and friends) is used to conform to the new interface.  I do agree that `kwargs` probably is a much better choice, but trying to maintain the current interface for now while adding the file splitting functionality in as well, and this was the best I could come up with.

A middle ground that would avoid the `arg` shifting in `#initialize_upload_vars` would be to just allow `local_io` to be passed in as `nil`.  The reason I am not a huge fan of that is I am usually not a fan of the first arg being optionally `nil`, because then you lead to calls that look like this:

```ruby
MiqLocalMountSession.add nil, "path/to/destination/file", "200M" do |output_file|
  `pg_dump -f #{output_file}...`
end
```
And I think that just muddies the interface, and makes the caller wonder what is going on with the `nil` for the argument.

But again, not married to this implementation, just my personal preference for a ideal-ish implementation and rational.

NickLaMuro: I guess to play a bit of devil's advocate for myself, I could probably implement the `kwargs` interface for the top level class method that I want to implement anyway, and just keep `#add` conforming to the current form arg structure.  🤔

lib/gems/pending/util/mount/miq_local_mount_session.rb
```diff
@@ -0,0 +1,37 @@
...
+  #
+  # This is an override to allow methods that make use of `@mnt_point` to
+  # behave as we would expect them to.
+  #
+  # Determine the mount point relative to the destination file, and just remove
+  # the portions of the expanded path that already exist in the
+  # `remote_file_path` (might be everything if `remote_file_path` is a absolute
+  # path)
+  def mnt_point
+    @mnt_point ||= File.expand_path(remote_file_path).sub(remote_file_path, "")
```
NickLaMuro: After some thinking and looking at how to implement this in `EvmDatabaseOps`, this is a bad way to implement this via `memoization`.

Part of the reason to set `@mnt_point` was to prevent changes to `MiqGenericMountSession`, which doesn't use the accessor, but `@mnt_point` directly in a lot of spots.  But, there are plenty of times where `remote_file_path` will not be set, and so generating `@mnt_point` in this fashion will cause issues (possibly just error out) if it is set and memoized before a `#add` call.

I have some ideas on how to do this better, but know that this implementation is probably going to change.

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,217 @@
...
+        File.mkfifo(fifo_path)
+
+        # For #Reasons(TM), the reader must be opened first
+        @source_input = File.open(fifo_path.to_s, IO::RDONLY | IO::NONBLOCK)
+        @input_writer = File.open(fifo_path.to_s, IO::WRONLY | IO::NONBLOCK)
+
+        Thread.new do
+          yield fifo_path      # send the path to the block to get executed
+          @input_writer.close  # close the file so we know we hit EOF (for #add)
+        end
```
NickLaMuro: Note for myself:  We are probably going to want to do something like:

```ruby
Thread.new do
  # ...
end.tap { t.abort_on_exception = true }
```

Otherwise errors can get swallowed up in this thread and subsequent threads below it, which isn't good.

NickLaMuro:
@miq-bot remove_label unmergable

NickLaMuro:
@carbonin While I think we should probably solve https://github.com/ManageIQ/manageiq-gems-pending/pull/361#discussion_r213499935 before merge, I don't have a good idea on what I want to do with that currently, so going to leave the `[WIP]` label for a bit more while I ponder that.

But I think everything else is at a pretty stable state currently and been put through it's paces pretty heavily already.  So I wouldn't mind a review on this with that in mind, and feel give any suggestions to the `Thread` comment as well.

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -142,22 +142,8 @@ def upload_single(file) # rubocop:disable Lint/UnusedMethodArgument
 
     def upload_splits
       @position = 0
-      begin
-        until source_input.eof?
-          upload_single(next_split_filename)
-          @position += byte_count
-        end
-      rescue Errno::EIO
```
NickLaMuro: This is going to go by bye because of a rebase, but want to leave a diff here in case it becomes relevant to bring it back in the future (I don't want to spend hours looking up how to fix this again):

```diff
commit 90ce1208de367f4c6b0119d772a9b411f7a7080d
Author: Nick LaMuro <nicklamuro@gmail.com>
Date:   Wed Aug 29 16:00:49 2018 -0500

    [FIXUP] Skip unneeded PTY based rescue
    
    Now that we aren't using PTY, this rescue should no longer be needed
    (hopefully).  Running this against Travis failed it without this last
    time, so if the specs still pass with this removed, I will rebase this
    in the the original commits.

diff --git a/lib/gems/pending/util/miq_file_storage.rb b/lib/gems/pending/util/miq_file_storage.rb
index dc7d488..43d7848 100644
--- a/lib/gems/pending/util/miq_file_storage.rb
+++ b/lib/gems/pending/util/miq_file_storage.rb
@@ -142,22 +142,8 @@ class MiqFileStorage
 
     def upload_splits
       @position = 0
-      begin
-        until source_input.eof?
-          upload_single(next_split_filename)
-          @position += byte_count
-        end
-      rescue Errno::EIO
-        # Error::EIO is raised when you read from a closed PTY on Linux.  See:
-        #
-        #   https://stackoverflow.com/a/10306782/3574689
-        #   https://ruby-doc.org/stdlib-2.2.3/libdoc/pty/rdoc/PTY.html#module-PTY-label-Example
-        #
-        # Or for the neckbeards:
-        #
-        #   $ ri PTY
-        #
-        # Done with reading at this point, but update @position... just incase
+      until source_input.eof?
+        upload_single(next_split_filename)
         @position += byte_count
       end
     end
```

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: Fixed this by adding a `being`/`ensure` in the code for the thread.  Basically the same code as above, but the first line is in the main body of the `begin`, and the `.close` is part of the `ensure`.

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: Being that this comment is a month old, I am going to update and say that I didn't go through with refactoring this.  I have spent a crap ton of time validating that this does indeed work (in unit tests and [the test suite gist](https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6)), and since it is "crunch time", I am going to argue we try and address this later.

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,203 @@
...
+    require 'uri'
+    scheme, _ = URI.split(URI::DEFAULT_PARSER.escape(opts[:uri]))
+    klass = storage_interface_classes[scheme]
+    klass.new_with_opts(opts) if klass
+  end
+  private_class_method :fetch_interface_class
+
+  def self.storage_interface_classes
+    @storage_interface_classes ||= Interface.descendants.each_with_object({}) do |klass, memo|
+      memo[klass.uri_scheme] = klass if klass.uri_scheme
```
carbonin: What happens when a class's `#uri_scheme` is nil? Will we not get an instance? Should we raise an error in `.fetch_interface_class` or here?

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,203 @@
...
+    }.freeze
+
+    attr_reader :remote_file_path, :byte_count, :source_input, :input_writer
+
+    def self.new_with_opts(opts) # rubocop:disable Lint/UnusedMethodArgument
+      raise NotImplementedError, "#{name}.new_with_opts is not defined"
+    end
+
+    def self.uri_scheme
+      nil
```
carbonin: Related to my previous comment. Should this also raise `NotImplementedError` or are we expecting some classes to leave this as `nil`?

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,203 @@
...
+    #
+    # Add a file to the destination URI.
+    #
+    # In the block form of the method, only the remote_uri is required, and it
+    # is assumed the input will be a generated in the executed block (most
+    # likely an external process) to a unix pipe that can be written to.  The
+    # pipe generated by this method and passed in to the block as a file
+    # location to the `input_stream`).
+    #
+    # In the non-block form, a source must be provioded as the first argument
```
carbonin: Typo, `s/provioded/provided/`

lib/gems/pending/util/miq_file_storage.rb
carbonin: Sounds good to me, I'll resolve this.

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,203 @@
...
+      else
+        download_single(remote_file_uri, local_file)
+      end
+    ensure
+      reset_vars
+    end
+
+    private
+
+    def upload_single(file) # rubocop:disable Lint/UnusedMethodArgument
```
carbonin: Can you put a comment here describing how this should be implemented at a high level to work with splitting?

spec/util/miq_file_storage_spec.rb
```diff
@@ -0,0 +1,380 @@
+require "pty" # required inline in module, needed here for stubbing
```
carbonin: Do we still need this?

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,203 @@
...
+    def reset_vars
+      File.delete(@input_writer.path) if @input_writer
+      @position, @byte_count, @remote_file_path, @source_input, @input_writer = nil
+    end
+
+    def next_split_filename
+      "#{remote_file_path}.#{'%05d' % (@position / byte_count + 1)}"
+    end
+
+    def download_single(source, destination) # rubocop:disable Lint/UnusedMethodArgument
```
carbonin: How does the download side handle splitting? Is there some kind of detection that the file being requested is split on the remote side?

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -251,55 +251,59 @@ def verify
     res
   end
 
-  def add(source, dest_uri)
+  def add(*upload_args)
```
carbonin: It looks like most of the content was removed. Can we just inherit the whole method or are we keeping it just for the logging?

spec/util/mount/miq_local_mount_session_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        `#{Gem.ruby} -e "File.write('#{input_writer}', '0' * #{tmpfile_size})"`
+      end
+
+      expected_splitfiles.each do |filename|
+        expect(File.exist?(filename)).to be true
+        expect(Pathname.new(filename).lstat.size).to eq(2.megabytes)
+      end
+    end
+
+    context "with slightly a slightly smaller input file than 10MB" do
```
carbonin: Typo, remove the first "slightly"

lib/gems/pending/util/mount/miq_local_mount_session.rb
```diff
@@ -0,0 +1,48 @@
...
+  # :nodoc:
+  #
+  # This is an override to allow methods that make use of `@mnt_point` to
+  # behave as we would expect them to.
+  #
+  # Determine the mount point relative to the destination file, and just remove
+  # the portions of the expanded path that already exist in the
+  # `remote_file_path` (might be everything if `remote_file_path` is a absolute
+  # path)
+  def mnt_point
```
carbonin: This logic looks a bit complicated. Should we test this directly or you you think this is exercised enough through the existing tests?

lib/gems/pending/util/miq_object_storage.rb
```diff
@@ -0,0 +1,19 @@
...
+    new(opts.slice(:uri, :username, :password))
+  end
+
+  def initialize(settings)
+    raise "URI missing" unless settings.key?(:uri)
+    @settings = settings.dup
+  end
+
+  def logger
+    @logger ||= $log.nil? ? :: Logger.new(STDOUT) : $log
```
carbonin: Do we have any facility for setting the logger for this gem as a whole or is each utility kind of responsible for doing it separately? Or was everything just relying on `$log`?

lib/gems/pending/util/extensions/aws-sdk/s3_upload_stream_patch.rb
```diff
@@ -0,0 +1,281 @@
...
+# introduced in the V3 of the 'aws-sdk'
+#
+# To my knowledge, the changes from this patch should be completely backwards
+# compatible, and the API should easily support this change.
+#
+# Once upgraded, this file should be removed.
+
+require 'aws-sdk'
+
+if defined? Aws::S3::GEM_VERSION # This constant doesn't exist in v2
```
carbonin: So, I wouldn't want to push it just for this patch, but do we know if we are super far away from being able to use the new version as a whole? Would be curious to know how much effort that would be rather than forward porting this patch. @bronaghs Can you comment here or refer who would know about the aws-sdk version we are using?

lib/gems/pending/util/object_storage/miq_s3_storage.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def self.new_with_opts(opts)
+    new(opts.slice(:uri, :username, :password, :region))
+  end
+
+  def initialize(settings)
+    super(settings)
+
+    # NOTE: This line to be removed once manageiq-ui-class region change implemented.
+    @settings[:region] ||= "us-east-1"
```
carbonin: Since we're here anyway ... @jerryk55 can we resolve this `NOTE:`?

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: Yeah, good call.

I will probably point to one of the existing classes for examples, but yeah, this should definitely be documented.

spec/util/miq_file_storage_spec.rb
NickLaMuro: Oh, good catch.  Nope, not needed now that we are just using `File.mkfifo`.  Will remove.

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: Yeah, I was thinking about this, but I decided to not implement this for the first pass.

The short answer is file splitting was originally targeted for support when uploading to a endpoint that had an upload size limit (might want to have @dmetzger57 weigh in, though, as this was technically his "feature creep suggestion").  Using the split versions just happened to be something that worked for other storage backends without issue, so hooking them up was a zero cost.

I _**COULD**_ try and do auto detection, but I think that might be a bit of a stretch to squeeze this in on time.  Leaning more towards skipping that for now and seeing if this is something we can add later.  I will at the very least research what is needed to handle this.

Thoughts on that path going forward?

lib/gems/pending/util/mount/miq_generic_mount_session.rb
NickLaMuro: I figured someone added the logging for a specific purpose, and didn't want to blind side them by removing it.

But yes, I was just keeping it for the logging mostly...

...

Though, now that I look at it, I also handled the `reconnect!` and `disconnect` here, and it makes sense that it is done as part of `.add` and not trying to do it for each file in `.upload_single` (assuming the many "splits" case).

lib/gems/pending/util/mount/miq_local_mount_session.rb
NickLaMuro: It is exercised (I believe) in the `MiqLocalMountSession` tests, as I switch up what I pass to `.add` in each of the different tests.

But I can target it specifically in that spec if you want (seems like it wouldn't be a bad idea).

lib/gems/pending/util/miq_object_storage.rb
NickLaMuro: I would have to look, but I want to say it is just a cluster f...

I will report back after I do, but I want to say that this one was a copy paste from somewhere, but I don't remember off hand.

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: @carbonin So with this, I don't think the class would be an issue, since this `if` check here basically prevents nils from being part of the hash.

But, I think the question should have been:

> what happens when the `storage_interface_class[sheme]` lookup returns `nil`?

Right now, when it is called from `.fetch_interface_class`, it will bomb on `.new_with_opts` with a `NoMethodError` on `nil`.  Not the best error, I agree.

I was considering just defaulting to `MiqLocalMountSession`.  The reason I considered doing that is when we are back in `ManageIQ/manageiq` land, we could technically refactor the `evm:db:local:*` tasks to just alias to the remote ones, and then URIs that just have a local path will work through the same interface.  But wasn't sure if that was a good choice

I think we could either do that, or just raise a more friendly error.  I am fine with either.

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: Related to my [previous answer](https://github.com/ManageIQ/manageiq-gems-pending/pull/361#discussion_r216506316) (:trollface:), if this isn't implemented, this won't get added to the lookup hash.

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: One thing I was considering might be an issue with this is when we add classes after the fact.  For example, if we were to move S3 and Swift implementations their respective `manageiq-provider-*` plugins.

Probably not an issue for now, but leaving this hear as a note.

lib/gems/pending/util/miq_object_storage.rb
NickLaMuro: So, we have basically "two-ish" ways we do this:

1.  We just log to `$log` directly, which is what we do in `PostgresAdmin` for example.
2.  What we are doing here.

Which is done not only in [`miq-benchmark.rb`](https://github.com/ManageIQ/manageiq-gems-pending/blob/b6ca5f0a/lib/gems/pending/util/extensions/miq-benchmark.rb#L78), but also where I guess I pulled this from (and forgot that I did) in [`MiqGenericMountSession`](https://github.com/ManageIQ/manageiq-gems-pending/blob/b6ca5f0a/lib/gems/pending/util/mount/miq_generic_mount_session.rb#L25).

So I would say we stay consistent and keep what we have here.

lib/gems/pending/util/extensions/aws-sdk/s3_upload_stream_patch.rb
NickLaMuro: > do we know if we are super far away from being able to use the new version as a whole?

I don't know personally, or if it has been considered at all.

I want to say there was something that was being done to update `fog` in the recent future, and I thought that was potentially going to update the `aws-sdk`, but I might be remembering incorrectly (_**Update:**_  I was wrong... no surprise...  It was for https://github.com/ManageIQ/manageiq/pull/17258 and it was `fog-google`, not AWS... opps)

NickLaMuro:
This comment will be hidden immediately (since it is very bulky, but leaving this here a reference to a `[FIXUP]` commit that I rebased out since it was part of the original implementation.

```diff
commit 5f5158e51d9ae7ef58bbcfb7bf0c9d4cb3b9ef6b
Author: Nick LaMuro <nicklamuro@gmail.com>
Date:   Tue Sep 4 17:21:12 2018 -0500

    [FIXUP] Handle errors in thread on upload
    
    When a sub thread has an error, it isn't raised unless `.join` is
    called.  So for starters, we weren't doing that (and so, the process
    would hang).
    
    On top of that, if somehow the sub thread causes the subsequent
    `upload_single` or `upload_splits` to hang because the `input_writer`
    never sends a close, then the process could also hang, so we handle that
    as well by wrapping it in an include.

diff --git a/lib/gems/pending/util/miq_file_storage.rb b/lib/gems/pending/util/miq_file_storage.rb
index 43d7848..f677e34 100644
--- a/lib/gems/pending/util/miq_file_storage.rb
+++ b/lib/gems/pending/util/miq_file_storage.rb
@@ -95,12 +95,16 @@ class MiqFileStorage
     def add(*upload_args, &block)
       initialize_upload_vars(*upload_args)
       mkdir(File.dirname(@remote_file_path))
-      handle_io_block(&block)
-      if byte_count
-        upload_splits
-      else
-        upload_single(@remote_file_path)
-      end
+      thread = handle_io_block(&block)
+      result = if byte_count
+                 upload_splits
+               else
+                 upload_single(@remote_file_path)
+               end
+      # `.join` will raise any errors from the thread, so we want to do that
+      # here (if a thread exists of course).
+      thread.join if block_given?
+      result
     ensure
       reset_vars
     end
@@ -181,8 +185,11 @@ class MiqFileStorage
         @input_writer = File.open(fifo_path.to_s, IO::WRONLY | IO::NONBLOCK)
 
         Thread.new do
-          yield fifo_path      # send the path to the block to get executed
-          @input_writer.close  # close the file so we know we hit EOF (for #add)
+          begin
+            yield fifo_path      # send the path to the block to get executed
+          ensure
+            @input_writer.close  # close the file so we know we hit EOF (for #add)
+          end
         end
       end
     end
diff --git a/spec/util/miq_file_storage_spec.rb b/spec/util/miq_file_storage_spec.rb
index 1ee1ec3..170b968 100644
--- a/spec/util/miq_file_storage_spec.rb
+++ b/spec/util/miq_file_storage_spec.rb
@@ -336,21 +336,22 @@ describe MiqFileStorage do
     end
 
     describe "#handle_io_block" do
+      let(:input_writer) { Tempfile.new }
+      let(:source_input) { Tempfile.new }
+
+      after do
+        input_writer.unlink
+        source_input.unlink
+      end
+
       context "with a block" do
-        let(:input_writer) { Tempfile.new }
-        let(:source_input) { Tempfile.new }
-        let(:block)        { ->(_input_writer) { sleep 0.1 } }
+        let(:block) { ->(_input_writer) { sleep 0.1 } }
 
         before do
           expect(File).to receive(:mkfifo)
           expect(File).to receive(:open).and_return(source_input, input_writer)
         end
 
-        after do
-          input_writer.unlink
-          source_input.unlink
-        end
-
         it "creates a thread for handling the input IO" do
           thread_count = Thread.list.count
           thread       = subject.send(:handle_io_block, &block)
@@ -375,6 +376,22 @@ describe MiqFileStorage do
           expect(Thread.list.count).to eq(thread_count)
         end
       end
+
+      context "with a block that causes an error" do
+        let(:err_block) { ->(_input_writer) { raise "err-mah-gerd" } }
+
+        before do
+          expect(File).to receive(:mkfifo)
+          expect(File).to receive(:open).and_return(source_input, input_writer)
+        end
+
+        it "does not hang the process and closes the writer" do
+          expect(input_writer.closed?).to eq(false)
+          thread = subject.send(:handle_io_block, &err_block)
+          expect { thread.join }.to raise_error StandardError
+          expect(input_writer.closed?).to eq(true)
+        end
+      end
     end
   end
 end
```

NickLaMuro:
@carbonin I believe that I addressed most of your comments.

lib/gems/pending/util/miq_file_storage.rb
carbonin: Okay, that makes sense but I think I would rather an error when we can't find a class for a particular uri scheme.

Right now it actually looks like you might get `nil` from `.fetch_interface_class`.

lib/gems/pending/util/miq_file_storage.rb
carbonin: My only concern with this is if we do file splitting for database backups, not having this could make restore very difficult to implement.

So if the only file store that will have splitting by default for now is going to be FTP then this should be fine, but if we're doing splitting for other storage types we need to solve this so that we can restore a split backup.

lib/gems/pending/util/miq_object_storage.rb
NickLaMuro: > is that some syntax that I'm about to learn about?

Nope, copy pasta from `MiqGenericMountSession`...

lib/gems/pending/util/miq_file_storage.rb
NickLaMuro: > So if the only file store that will have splitting by default for now is going to be FTP then this should be fine...

@carbonin So everything is going to have splitting, it is baked into the the question set on `DatabaseAdmin` in `appliance_console` and the option params in the rake tasks, aside from the fact that it is part of the top level logic in `MiqFileStorage`, so I am not planning on doing any logic to make it available for one provider only.

So this said, I don't see this as being "easy" to implement auto detection but agree that this is worthwhile to have.  To me this sees like something I could follow up and do in a subsequent PR.  Does that sound reasonable or is this something that needs to be included for this PR to be considered merge-able?

cc @dmetzger57

carbonin: My request for this PR is to not introduce restore-side regressions for existing backup targets. If FTP can't restore backups, that's fine, but if this breaks restore from NFS, that's not fine.

So, hypothetically, if we do a backup to an NFS target, how does restore work when the backup is split on the remote? I may be missing some crucial detail, but it feels like this won't work as is.

NickLaMuro: @carbonin Sorry, I think I now see where the confusion is.  So to start off:

> My request for this PR is to not introduce restore-side regressions for existing backup targets.

There are none, but I think I misunderstood where your confusion was possibly originating from.

`MiqFileStorage#download_single` is used in both cases of `#download` (block form and not).  This differs from `#add` which will call either `#upload_splits` when a `byte_count` is passed, calling `#upload_single` multiple times, or just directly to `#upload_single`.

`#download` should work as it has previously.

> So, hypothetically, if we do a backup to an NFS target, how does restore work when the backup is split on the remote?

This, however won't work, but again, splitting is "opt-in", and isn't something you are required to do.  That means:

- You can still upload a single file backup and restore that single file backup as you had in the past.
- You can also upload files split on file size
- Both methods will be streamed through the same code

What you can't do is restore from a remote that is split automatically right now, and I am not even sure how that would look if we did (do we use a glob?  Do we try and `ls` for the file if an exact match isn't found?  Do we have to re-prompt the user if we find split files?).

You would have to cat the files together on said remote and then restore currently.

* * *

I see being able to upload a file split as a new feature, and being able to restore split files as an augment to said feature, or a new feature in it of itself.  The original feature was in place so Support could upload a part based dump automatically.  You can always skip splitting the file (by saying 'n'), as shown on how it is prompted in the integrating PR in the [`appliance_console`](https://github.com/ManageIQ/manageiq-appliance_console/pull/50):

```
Would you like to split the dump output into multiple parts? (Y/N): y
Enter the byte size to split by: |500M| 750M
```

Possibly a warning of "currently there is no support automatically restore split backups/dumps" might be a good idea, but again, I don't see that as a "regression", just missing functionality.

Does that make sense?

carbonin: Ah okay, I was missing the distinction between "everything is going to have splitting" and "splitting is opt-in".

Like I said, as long as this doesn't change the behavior for existing targets it's fine with me. :+1:

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,243 @@
...
+  end
+
+  def self.fetch_interface_class(opts)
+    return nil unless opts[:uri]
+
+    require 'uri'
+    scheme, _ = URI.split(URI::DEFAULT_PARSER.escape(opts[:uri]))
+    klass = storage_interface_classes[scheme]
+
+    raise ArgumentError, "Not a valid uri!" if klass.nil?
```
carbonin: I think we can be more descriptive here. Maybe something like:

```ruby
raise ArgumentError, "#{scheme} is not a valid MiqFileStorage uri scheme. Accepted schemes are #{storage_interface_classes.keys}" if klass.nil?
```

lib/gems/pending/util/object_storage/miq_s3_storage.rb
carbonin: Okay, but if we're keeping it, then we can probably drop the comment. Waiting on your input here @jerryk55

lib/gems/pending/util/extensions/aws-sdk/s3_upload_stream_patch.rb
Fryguy: Switching to v3 of the amazon API is not in the near future...  cc @bronaghs @agrare

lib/gems/pending/util/mount/miq_local_mount_session.rb
NickLaMuro: Actually, after doing some investigating, I am going to try deleting this code and use a slightly different approach.  I will report back, but that might be the easier route.

lib/gems/pending/util/miq_file_storage.rb
```diff
@@ -0,0 +1,254 @@
...
+      mkdir(File.dirname(@remote_file_path))
+      thread = handle_io_block(&block)
+      result = if byte_count
+                 upload_splits
+               else
+                 upload_single(@remote_file_path)
+               end
+      # `.join` will raise any errors from the thread, so we want to do that
+      # here (if a thread exists of course).
+      thread.join if block_given?
```
carbonin: I think this would be more clear if we used the thread's existence as the condition rather than if the block was provided.

It's a bit confusing that I have to look into `#handle_io_block` to figure out that a thread is only created if a block is provided.

NickLaMuro: Sure, makes sense to me.  I think I was just using that as the check elsewhere, but I think that is a reasonable change.

lib/gems/pending/util/extensions/aws-sdk/s3_upload_stream_patch.rb
carbonin: Looks like the far future has arrived? https://github.com/ManageIQ/manageiq-gems-pending/pull/374

lib/gems/pending/util/object_storage/miq_s3_storage.rb
```diff
@@ -0,0 +1,101 @@
...
+        bucket.create
+      end
+    end
+  end
+
+  private
+
+  def s3
+    require 'aws-sdk'
+    require 'util/extensions/aws-sdk/s3_upload_stream_patch'
```
NickLaMuro: Bah... this means that this didn't get loaded in the specs... which would have made things easier...

Oh well I guess, we caught the PR updating this, so I guess all is good.

PR #364, opened 2018-08-15
Use --xlog-method=stream for pg_basebackup

carbonin:
Previously we were using the "fetch" method which would fail if
we did not set wal_keep_segments to a high enough value. We were
never setting a value for wal_keep_segments so this would (infrequently)
fail.

https://bugzilla.redhat.com/show_bug.cgi?id=1616022


carbonin:
See also https://github.com/ManageIQ/container-postgresql/pull/12 for the container version of this PR

PR #365, opened 2018-08-22
Use STDOUT for pg_basebackup

NickLaMuro:
Extracted from https://github.com/ManageIQ/manageiq-gems-pending/pull/361

Background
----------

The `--pgdata` option for `pg_basebackup` only will allow either dumping the resulting tar to a directory or piping it to stdout, but not specifying the dir that the tar is written to (even though `--pgdata -` writes a single tar).

This change, while pretty dense, does simplify things a bit by not requiring a tmpdir and moving files around after the backup has run.  This means it can be streamed from this process as well without requiring branching logic.

This does a few things to accomplish this:

- Pushes output to `STDOUT` to the command always
- Switches out `AwesomeSpawn.run` for vanilla `Kernel.spawn`
- Redirects the output to a `IO.pipe`
- Uses IO.copy_stream to push the redirected `STDOUT` to the final file

To start, switching out `AwesomeSpawn` was necessary since there isn't the ability to use IO redirection with it's API.  We are still using it to build the options, and refactored a few methods to share code from `run_cmd` with what we are doing here.

The thread is necessary so we are piping our input at the same time we are copying it via `IO.copy_stream`.  `.copy_stream` is used since it is a lightweight, C based method that can work with IO objects and file paths seamlessly, which we do here.

What this means
---------------

AKA, what sucks about it...

We are feeding basically **ALL** of our data through Ruby now from `pg_basebackup`.  From [my testing](https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6), I didn't notice much of a difference, but I was only testing against 10MB backups and 3MB dumps, so there wasn't much to judge by.

That said, we are making use of `IO.copy_stream`, which should be about as good as we can get with it in ruby, and most of the streaming should be avoiding Ruby's GC and other slow parts of the language.

But, it is also dense and there is a lot of IO pipes and threads being tossed all over the place.  When using this paired with `EvmDatabaseOps` to a NFS storage, for example, we are:

- Shelling out to `pg_basebackup` from the main process
- Piping the STDOUT into 'write' end of a `IO.pipe`in another `Thread`
- From the main thread, taking the 'read' end of the pipe and feeding it into another IO (via `IO.copy_stream`)
- If we are using from `EvmDatabaseOps`, that `IO` will be a `fifo`, so another `IO` and "pipe-like".
- That is then sent to another `IO.copy_stream` to the final destination file
  - if we are splitting the file, it has multiple of these
- And since we are on NFS in this example, that data is being streamed over the wire to the actual attached storage

So yeah, a bit to wrap your head around, and probably not easy for someone to just walk in on and understand.

Links
-----

* Used by:
  - https://github.com/ManageIQ/manageiq-gems-pending/pull/361
  - https://github.com/ManageIQ/manageiq/pull/17894

### DIFF
--- a/lib/gems/pending/util/postgres_admin.rb
+++ b/lib/gems/pending/util/postgres_admin.rb
@@ -136,12 +136,26 @@ class PostgresAdmin
     opts.delete(:dbname)
 
     path = Pathname.new(opts.delete(:local_file))
-
     FileUtils.mkdir_p(path.dirname)
-    Dir.mktmpdir("vmdb_backup", path.dirname) do |dir|
-      runcmd("pg_basebackup", opts, :z => nil, :format => "t", :xlog_method => "fetch", :pgdata => dir)
-      FileUtils.mv(File.join(dir, "base.tar.gz"), path)
-    end
+
+    # Build commandline from AwesomeSpawn
+    args = {:z => nil, :format => "t", :xlog_method => "fetch", :pgdata => "-"}
+    cmd  = AwesomeSpawn.build_command_line("pg_basebackup", combine_command_args(opts, args))
+    $log.info("MIQ(#{name}.#{__method__}) Running command... #{cmd}")
+
+    # Run command in a separate thread
+    read, write    = IO.pipe
+    error_path     = Dir::Tmpname.create("") { |tmpname| tmpname }
+    process_thread = Process.detach(Kernel.spawn(pg_env(opts), cmd, :out => write, :err => error_path))
+    stream_reader  = Thread.new { IO.copy_stream(read, path) } # Copy output to path
+    write.close
+
+    # Wait for them to finish
+    process_status = process_thread.value
+    stream_reader.join
+    read.close
+
+    handle_error(cmd, process_status.exitstatus, error_path)
     path.to_s
   end
 
@@ -216,9 +230,7 @@ class PostgresAdmin
 
   def self.runcmd_with_logging(cmd_str, opts, params = {})
     $log.info("MIQ(#{name}.#{__method__}) Running command... #{AwesomeSpawn.build_command_line(cmd_str, params)}")
-    AwesomeSpawn.run!(cmd_str, :params => params, :env => {
-                        "PGUSER"     => opts[:username],
-                        "PGPASSWORD" => opts[:password]}).output
+    AwesomeSpawn.run!(cmd_str, :params => params, :env => pg_env(opts)).output
   end
 
   private_class_method def self.file_type(file)
@@ -233,6 +245,12 @@ class PostgresAdmin
     default_args.merge(args)
   end
 
+  private_class_method def self.pg_env(opts)
+    {
+      "PGUSER"     => opts[:username],
+      "PGPASSWORD" => opts[:password]
+    }
+  end
   # rubocop:disable Style/SymbolArray
   PG_DUMP_MULTI_VALUE_ARGS = [
     :t, :table,  :T, :exclude_table,  :"exclude-table", :exclude_table_data, :"exclude-table-data",
@@ -252,5 +270,17 @@ class PostgresAdmin
       end
     end
     args
+  end
+
+  private_class_method def self.handle_error(cmd, exit_status, error_path)
+    if exit_status != 0
+      result = AwesomeSpawn::CommandResult.new(cmd, "", File.read(error_path), exit_status)
+      message = AwesomeSpawn::CommandResultError.default_message(cmd, exit_status)
+      $log.error("AwesomeSpawn: #{message}")
+      $log.error("AwesomeSpawn: #{result.error}")
+      raise AwesomeSpawn::CommandResultError.new(message, command_result)
+    end
+  ensure
+    File.delete(error_path) if File.exist?(error_path)
   end
 end


NickLaMuro:
@carbonin I assume you will have some pretty major opinions on this, but that really relies on the contextual knowledge of what I am working on with the `MiqFileStorage` PRs (linked above).  Let me know if I need to explain this in more detail (then I already have), it was a bit confusing for me to wrap my head around as well.

Though, maybe I am just a bit dense...

cc @Fryguy @kbrock @jerryk55

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,24 @@ def self.backup_pg_compress(opts)
...
 
     path = Pathname.new(opts.delete(:local_file))
-
     FileUtils.mkdir_p(path.dirname)
-    Dir.mktmpdir("vmdb_backup", path.dirname) do |dir|
-      runcmd("pg_basebackup", opts, :z => nil, :format => "t", :xlog_method => "fetch", :pgdata => dir)
-      FileUtils.mv(File.join(dir, "base.tar.gz"), path)
+
+    # Build commandline from AwesomeSpawn
+    args = {:z => nil, :format => "t", :xlog_method => "stream", :pgdata => "-"}
```
carbonin: This needs to change back to `:xlog_method =>  "fetch"` right?

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,24 @@ def self.backup_pg_compress(opts)
...
+    cmd  = AwesomeSpawn.build_command_line("pg_basebackup", combine_command_args(opts, args))
+
+    # Run command in a separate thread
+    read, write = IO.pipe
+    cmd_thread = Thread.new do
+      pid = Kernel.spawn(pg_env(opts), cmd, :out => write)
```
carbonin: This looks good, but I think it's a bit cleaner to use the thread for reading the output and let the "main" thread do all the process and thread waiting. How does this look?

```ruby
# spawn a subprocess to do the backup and a thread to read the output
backup_pid = Kernel.spawn(pg_env(opts), cmd, :out => write)
write.close
stream_reader = Thread.new { IO.copy_stream(read, path) }

# wait for them both to finish
Process.wait(backup_pid)
stream_reader.join
read.close

Comment (carbonin) [reply lib/gems/pending/util/postgres_admin.rb]: Then you can also avoid the perfectly valid, but rather odd spawn from within a thread thing here.

File: lib/gems/pending/util/postgres_admin.rb [hunk 3]
```diff
@@ -136,12 +136,24 @@ def self.backup_pg_compress(opts)
...
+      Process.wait(pid)
```
carbonin: We should probably check this return code, right?

carbonin: (I think the old logic would raise if the backup failed)

NickLaMuro: Yeah, I might have to check what AwesomeSpawn.run! is doing for logging, and emulate that here. But I agree, we should handle here as well.

lib/gems/pending/util/postgres_admin.rb
NickLaMuro: @carbonin hrmm.... now that I look at this closely, not sure this would work, at least in it's current form.

Part of what I was doing with the previous code was waiting until `Process.wait` was finished before calling `write.close`.  Here you just close it, which will break the pipe as you are trying to write to it from `pg_basebackup`, unless the backup is super quick (not likely).

I think moving `write.close` to under the `Process.wait(backup_pid)` will still allow this to work, and then the `stream_reader` should be able to `.join` shortly after.  I will test that out.

carbonin: Nope, pretty sure this works as that only closes the write end in the current process. Maybe if `spawn` doesn't actually create a separate process this would be a problem, but because this is essentially a `fork .. exec` this should work.

Actually I wrote a test program to be sure:
```ruby
r, w = IO.pipe
puts "forking writer"
pid = fork do
  begin
    10.times do
      sleep 1
      w.puts "things"
    end
  rescue => e
    puts "writer error #{e.message}"
  end
  puts "writer done"
end
w.close

puts "creating reader thread"
reader = Thread.new { IO.copy_stream(r, STDOUT) }

puts "waiting on writer"
Process.wait(pid)

puts "wiating on reader"
reader.join

r.close
```

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,25 @@ def self.backup_pg_compress(opts)
...
-      FileUtils.mv(File.join(dir, "base.tar.gz"), path)
-    end
+
+    # Build commandline from AwesomeSpawn
+    args = {:z => nil, :format => "t", :xlog_method => "fetch", :pgdata => "-"}
+    cmd  = AwesomeSpawn.build_command_line("pg_basebackup", combine_command_args(opts, args))
+
+    # Run command in a separate thread
+    read, write   = IO.pipe
+    error         = Tempfile.new.tap(&:close).path
```
carbonin: Are you sure this will work?

Ruby will GC a tempfile if it goes out of scope so you need to keep a handle to the file until you're actually done with it or get a path without creating the file. I would be afraid of a race condition between you reading the file and ruby GCing (and removing whatever is at that path).

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,25 @@ def self.backup_pg_compress(opts)
     opts.delete(:dbname)
 
     path = Pathname.new(opts.delete(:local_file))
```
carbonin: Now, this thing won't always be a path right? Eventually you'll be passing in some other IO object here so you can do splitting, right?

I guess you'll have to come in and make changes here at that point anyway ...

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,25 @@ def self.backup_pg_compress(opts)
...
+    pid           = Process.detach(Kernel.spawn(pg_env(opts), cmd, :out => write, :err => error))
+    stream_reader = Thread.new { IO.copy_stream(read, path) } # Copy output to path
+    write.close
+
+    # Wait for them to finish
+    pid.join
+    stream_reader.join
+    read.close
+
+    handle_error(cmd, pid.exit_status)
```
carbonin: Think you're missing a parameter here

NickLaMuro: Yup, good catch.

lib/gems/pending/util/postgres_admin.rb
NickLaMuro: No, everything is actually a path so far in https://github.com/ManageIQ/manageiq-gems-pending/pull/361

That made things that way since they were easier to pass to commands in the way they already were being used (i.e. the `:local_file` option in this class).

Unfortunately, I didn't forsee having to do make the change in this PR, so IO objects may have been a more graceful way to handle things in the long run, but oh well...

NickLaMuro: I think I did a poor job of explaining that... let me try that again...

Since we were using `AwesomeSpawn` before, we had zero access to the `:in` and `:out` streams of the command we were passing, so this lib always passed file paths to `pg_*` executables we were shelling out to.

While I did switch to `Kernel.spawn` here so I had access to those streams to make this PR function (do to the limitations of `pg_basebackup` and `--pgdata`), I kept the interface the same as it was before, which assumed file paths, which is how I designed things in #361 as well (so file splitting works fine, and we just read from the `fifo`).

carbonin: Okay, so this will just be a named pipe which still qualifies as a file path :ok_hand: Got it

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,26 @@ def self.backup_pg_compress(opts)
...
+
+    # Build commandline from AwesomeSpawn
+    args = {:z => nil, :format => "t", :xlog_method => "fetch", :pgdata => "-"}
+    cmd  = AwesomeSpawn.build_command_line("pg_basebackup", combine_command_args(opts, args))
+    $log.info("MIQ(#{name}.#{__method__}) Running command... #{cmd}")
+
+    # Run command in a separate thread
+    read, write   = IO.pipe
+    error_path    = Dir::Tmpname.create("") { |tmpname| tmpname }
+    process       = Process.detach(Kernel.spawn(pg_env(opts), cmd, :out => write, :err => error_path))
```
carbonin: This is interesting. Why the change from `Process.wait` to `Process.detach`?

I thought we wouldn't be able to get the exit status if we detach.

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -136,12 +136,26 @@ def self.backup_pg_compress(opts)
...
+    stream_reader = Thread.new { IO.copy_stream(read, path) } # Copy output to path
+    write.close
+
+    # Wait for them to finish
+    pid = process.value
```
carbonin: Is this behavior documented anywhere? At this point `process` is a thread object and `#value` essentially does `#join` and returns the ["value" of the thread](https://ruby-doc.org/core-2.4.1/Thread.html#method-i-value) in this case are we sure that the thread that is created by `detach` will return something that responds to `#exitstatus`? I don't see that documented anywhere.

carbonin: I guess I'm still just confused why we can't `Process.wait` in the main thread.

lib/gems/pending/util/postgres_admin.rb
NickLaMuro: `Process.detach` returns a `Process::Waiter` (which is what is also what `Open3.popen3` for example will return instead, for example).

A `Process::Waiter` just inherits from `Thread`, that basically watches a pid until it finishes, and returns a `Process::Status` once it is finished.  The reason I went with this method is I would be wrapping this with `Process::Status` anyway to get the `.exitstatus`, so this is just a easier way of dealing with it.  I am less bothered by the extra thread anyway since I am pretty sure it is well optimized in C anyway, and will mostly deferring to the other IO heavy threads.

lib/gems/pending/util/postgres_admin.rb
NickLaMuro: > I guess I'm still just confused why we can't Process.wait in the main thread.

I think I covered most of this here (or at least the "why" part of your question):

https://github.com/ManageIQ/manageiq-gems-pending/pull/365/files#r213669910

Basically, to use vanilla `Process.wait`, I would be doing:

```ruby
Process.wait(pid)
status = Process::Status.new(pid).exitstatus
```

> Is this behavior documented anywhere?

Actually, I am having trouble finding it, but I swear I found it somewhere directly...

Anyway, where I did find it's usage of `Process.detach` is in `Open3`, in both the [`#popen_run`](https://github.com/ruby/ruby/blob/ac543abe/lib/open3.rb#L200) and [`#pipeline_run`](https://github.com/ruby/ruby/blob/ac543abe/lib/open3.rb#L659) methods, and it's recommendation to use `.value` in basically all of it's documentation:

https://ruby-doc.org/stdlib-1.9.3/libdoc/open3/rdoc/Open3.html#method-c-popen3

So I am not pulling this out of no where.

lib/gems/pending/util/postgres_admin.rb
carbonin: Okay, that's fine. I just couldn't find the behavior documented anywhere. Can we just use better names for the variables though? Could this one be `wait_thread` or something?

lib/gems/pending/util/postgres_admin.rb
carbonin: :+1: Can we just use a better variable name than `pid` for this? Makes it seem like it's an integer then it looks confusing when we call `#exitstatus` on it.

NickLaMuro: Right on, yeah, I am fine with that.

For the above, I was thinking of `status`, and for your other comment `process_thread`.  So:

```
pid     -> status
process -> process_thread
```

Sound reasonable?

PR #367, opened 2018-08-24
Revert to using --xlog-method fetch for postgres backup

### LINKED ISSUE #366: `--xlog-method=stream` causes `rake evm:db:backup` to break
Currently causes the following error:

```console
$ pg_basebackup --no-password -z --format t --xlog-method stream --pgdata backup_dir
pg_basebackup exit code: 1
pg_basebackup: WAL streaming can only be used in plain mode
```

Tested on `manageiq/gaprindashvili` 7.4.0 Vagrant VM.  Steps to reproduce:

```console
$ cd my_vagrant_testing_dir
$ vagrant init -m --output - manageiq/gaprindashvili | awk -v c='  config.vm.provider "virtualbox"' "{ print } /^Vagrant.configure/ { print c }" > Vagrantfile
$ vagrant box update --box manageiq/gaprindashvili
$ vagrant up
$ vagrant ssh
vagrant@appliance $ vmdb
vagrant@appliance $ sudo --shell
root@appliance $ curl https://raw.githubusercontent.com/ManageIQ/manageiq-gems-pending/master/lib/gems/pending/util/postgres_admin.rb > $(ls -d /usr/local/lib/ruby/gems/2.3.0/bundler/gems/manageiq-gems-pending-* | tail -n 1)/lib/gems/pending/util/postgres_admin.rb
root@appliance $ bin/rails evm:db:backup:local -- --local-file my-backup.tar.gz
root@appliance $ grep -A1 AwesomeSpawn log/evm.log
```

cc @carbonin

carbonin:
This is because we cannot use the tar option with stream.
We need the backup to be written in a single stream for the work
regarding file splitting and backing up to cloud-based targets to
move forward.

Fixes #366
Reverts #364 

cc @NickLaMuro

### DIFF
--- a/lib/gems/pending/util/postgres_admin.rb
+++ b/lib/gems/pending/util/postgres_admin.rb
@@ -139,7 +139,7 @@ class PostgresAdmin
 
     FileUtils.mkdir_p(path.dirname)
     Dir.mktmpdir("vmdb_backup", path.dirname) do |dir|
-      runcmd("pg_basebackup", opts, :z => nil, :format => "t", :xlog_method => "stream", :pgdata => dir)
+      runcmd("pg_basebackup", opts, :z => nil, :format => "t", :xlog_method => "fetch", :pgdata => dir)
       FileUtils.mv(File.join(dir, "base.tar.gz"), path)
     end
     path.to_s


bdunne:
Should we be setting wal_keep_segments if we revert https://github.com/ManageIQ/manageiq-gems-pending/pull/364#issue-208568723 ? Or is it too installation specific?
And do we need to revert for container-postgres as well? Or is that okay because of the way we run backups in podified?

NickLaMuro:
@bdunne in case you missed it, here is the conversation we had on it:

https://gitter.im/ManageIQ/manageiq/core?at=5b8024d8e5b40332abf1c289

> Should we be setting wal_keep_segments if we revert #364 (comment) ? Or is it too installation specific?

Based on my experience, not setting `wal_keep_segments` will still allow things to work, and it is (I think) when we get to a large enough size for the backups where this then becomes an issue, since it (I think) attempts to fetch the segments after cloning the directory(s) and fails because it is no longer in memory (again, this is a speculation on to the internals of `pg` from my end).

On the otherhand, what is currently in place is flat out broken unless we make some major changes, since `-z --format t` is completely incompatible with `--xlog-method="stream"` (short version:  you can't stream to a `tar` file).

My two cents would be to revert now and determine what to do about the `wal_keep_segments` later when we can better determine some sensible defaults, and fix things on a case by case basis for the time being where the issue from #364 is relevant.

> And do we need to revert for container-postgres as well? Or is that okay because of the way we run backups in podified?

I am unfortunately unfamiliar with that and unable to comment.  I can take a look tomorrow if someone else hasn't chimed in.

PR #368, opened 2018-08-29
Add MiqFtpStorage

NickLaMuro:
Built off of https://github.com/ManageIQ/manageiq-gems-pending/pull/360 (only the last two commits are added as a part of this pull request).

This adds the `FTP` protocol as a `ObjectStorage`, and allows it to work with the `MiqFileStorage::Interface`.

~~**Note:**  To any "very eager reviewer", I have a few commits in here at the moment that should live over in the https://github.com/ManageIQ/manageiq-gems-pending/pull/360 PR, but didn't realize I wanted them until I started this effort.  Will probably rebase them out into that other branch tomorrow.~~ Done.

Links
-----
* Built off of (needs to be merged first):
  - [x] https://github.com/ManageIQ/manageiq-gems-pending/pull/365
  - [x] https://github.com/ManageIQ/manageiq-gems-pending/pull/361
  - [x] https://github.com/ManageIQ/manageiq-gems-pending/pull/360

### DIFF
--- a/lib/gems/pending/util/object_storage/miq_ftp_storage.rb
+++ b/lib/gems/pending/util/object_storage/miq_ftp_storage.rb
@@ -1,0 +1,86 @@
+require 'util/miq_ftp_lib'
+require 'util/miq_object_storage'
+require 'logger'
+
+class MiqFtpStorage < MiqObjectStorage
+  include MiqFtpLib
+
+  attr_reader :uri, :username, :password
+
+  def self.uri_scheme
+    "ftp".freeze
+  end
+
+  def self.new_with_opts(opts)
+    new(opts.slice(:uri, :username, :password))
+  end
+
+  def initialize(settings)
+    super
+    @uri      = @settings[:uri]
+    @username = @settings[:username]
+    @password = @settings[:password]
+  end
+
+  # Override for connection handling
+  def add(*upload_args)
+    with_connection { super }
+  end
+
+  # Override for connection handling
+  def download(*download_args)
+    with_connection { super }
+  end
+
+  # Specific version of Net::FTP#storbinary that doesn't use an existing local
+  # file, and only uploads a specific size (byte_count) from the input_file
+  def upload_single(dest_uri)
+    ftp.synchronize do
+      ftp.send(:with_binary, true) do
+        conn = ftp.send(:transfercmd, "STOR #{uri_to_relative(dest_uri)}")
+        IO.copy_stream(source_input, conn, byte_count)
+        conn.close
+        ftp.send(:voidresp)
+      end
+    end
+    dest_uri
+  rescue Errno::EPIPE
+    # EPIPE, in this case, means that the data connection was unexpectedly
+    # terminated.  Rather than just raising EPIPE to the caller, check the
+    # response on the control connection.  If getresp doesn't raise a more
+    # appropriate exception, re-raise the original exception.
+    ftp.send(:getresp)
+    raise
+  end
+
+  def download_single(source, destination)
+    ftp.getbinaryfile(uri_to_relative(source), destination)
+  end
+
+  def mkdir(dir)
+    create_directory_structure(uri_to_relative(dir))
+  end
+
+  private
+
+  def login_credentials
+    [username, password].compact
+  end
+
+  # Currently assumes you have just connected and are at the root logged in
+  # dir.  Net::FTP (or ftp in general) doesn't seem to have a concept of a
+  # "root dir" based on your login, so this should be used right after
+  # `.connect`, or shortly there after.
+  #
+  # Or, you should be returning to the directory you came from prior to using
+  # this method again, or not using `ftp.chdir` at all.
+  def uri_to_relative(filepath)
+    result = URI.split(filepath)[5]
+    result = result[1..-1] if result[0] == "/".freeze
+    result
+  end
+
+  def _log
+    logger
+  end
+end
--- a/spec/util/object_storage/miq_ftp_storage_spec.rb
+++ b/spec/util/object_storage/miq_ftp_storage_spec.rb
@@ -1,0 +1,119 @@
+require 'util/object_storage/miq_ftp_storage.rb'
+
+describe MiqFtpStorage, :with_ftp_server do
+  subject         { described_class.new(ftp_creds.merge(:uri => "ftp://localhost")) }
+  let(:ftp_creds) { { :username => "ftpuser", :password => "ftppass" } }
+
+  describe "#add" do
+    include_context "generated tmp files"
+
+    shared_examples "adding files" do |dest_path|
+      let(:dest_path) { dest_path }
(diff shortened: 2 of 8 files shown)


NickLaMuro:
Ugh... looks like I didn't update one of the dependent branches before rebasing them into this one... I will fix tomorrow...

NickLaMuro:
(re-added the `[WIP]` label since I am waiting for the pre-reqs to be merged before this can be merged as well)

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
...
+        ftp.send(:voidresp)
+      end
+    end
+    dest_uri
+  rescue Errno::EPIPE
+    # EPIPE, in this case, means that the data connection was unexpectedly
+    # terminated.  Rather than just raising EPIPE to the caller, check the
+    # response on the control connection.  If getresp doesn't raise a more
+    # appropriate exception, re-raise the original exception.
+    getresp
```
carbonin: Where is this defined?

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
...
+  end
+
+  def initialize(settings)
+    super
+    @uri      = @settings[:uri]
+    @username = @settings[:username]
+    @password = @settings[:password]
+  end
+
+  # Override for connection handling
```
carbonin: The connection handling in question here is making sure you don't re-open for each file fragment I assume or is there more to it than that?

To be clear I'm fine with the implementation. I'm just curious.

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
...
+  def add(*upload_args)
+    with_connection { super }
+  end
+
+  # Override for connection handling
+  def download(*upload_args)
```
carbonin: Probably should be called `download_args`

spec/util/object_storage/miq_ftp_storage_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    it "can take input from a command" do
+      source_data = nil
+      subject.download(nil, source_path) do |input_writer|
+        source_data = `#{Gem.ruby} -e "print File.read('#{input_writer}')"`
+      end
+
+      # Nothing written, just printed the streamed file in the above command
+      expect(File.exist?(dest_path)).to be false
+      expect(source_path).to exist_on_ftp_server
```
carbonin: Shouldn't this be making some expectation on `source_data` (which I think is the data downloaded from the server) rather than `source_path` which definitely shouldn't have changed?

I may be missing some nuance here though.

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
+require 'util/miq_ftp_lib'
+require 'util/miq_object_storage'
+require 'logger'
+
+class MiqFtpStorage < MiqObjectStorage
+  include MiqFtpLib
+
+  FTP_CHUNKSIZE = ::Net::FTP::DEFAULT_BLOCKSIZE
```
carbonin: Where is this used?

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
+require 'util/miq_ftp_lib'
+require 'util/miq_object_storage'
+require 'logger'
+
+class MiqFtpStorage < MiqObjectStorage
```
carbonin: Can we move the existing implementation of `.new_with_opts` from `MiqObjectStorage` to this class? It's a bit strange right now that it is defined in the base class and we have two usable sub-classes, one of which overrides it. I feel like it would be more clear if each class defined it itself unless you know of a third which would share the implementation with `MiqFtpStorage`?

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
NickLaMuro: Yeah, basically the same reason that we overrode `MiqGenericMountSession#add` as well.  If we aren't splitting, technically it would have only done it once, but with splits, it would do it with each file fragment (as you have mentioned).  `.upload_single` is used in both cases, so overriding there doesn't make sense.

But, possibly it wouldn't hurt to have some follow up documentation (in a different PR) around `MiqFileStorage#add`/`MiqFileStorage#download` stating that those methods can be overwritten for cases like connection handling, but `super` must be called to conform to the spec.

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
NickLaMuro: (－‸ლ)

I'm a derp... (will fix)

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
NickLaMuro: Oops... yeah, this is part of `Net::FTP`, and it really should be `ftp.send(:getresp)`... pretty sure.

Yeah, screwed that one up, but don't have a good test around it currently.  Will see if I can work on up, but for now, I will confirm that `ftp.send(:getresp)` what makes sense since the code is a update to the internal `Net::FTP#storbinary`.

spec/util/object_storage/miq_ftp_storage_spec.rb
NickLaMuro: Sorry, this is a bit of a weird test.

I was mostly using these three tests as confirmation that the file existed in the first place, and that it was of the size we expected similar to the "sanity checks" in the above spec.

What I probably could do is label these three expects as `# sanity checks` and then have a final spec that takes `source_data`, and make sure it's `.size` is `10.megabytes`.  That make sense?

Unless I think of a more clever way of handling the `source_file` data (instead of it being `10.megabytes` of 0's), there isn't too much worth in validating it since it is so easy to generate.  Open to suggestions if we want to find a way to generate random data in a different, but some what speedy, way.  Considering some ideas ideas for that though, so I will report back if I do make a change.

lib/gems/pending/util/object_storage/miq_ftp_storage.rb
```diff
@@ -0,0 +1,92 @@
...
+  include MiqFtpLib
+
+  FTP_CHUNKSIZE = ::Net::FTP::DEFAULT_BLOCKSIZE
+
+  attr_reader :uri, :username, :password
+
+  class << self
```
NickLaMuro: Going to remove this whole block as we have `MiqObjectStorage#logger`, so I will just make use of that.

PR #378, opened 2018-10-04
Don't set empty ENV values for database dumps

kbrock:
Currently, `PGUSER` and `PGPASSWORD` are set in the environment.
We want these values to pass through to the subshells

### Before

`$PGUSER` was overridden by `""` if `--dbuser` was not passed into the rake task

### After

The existing `$PGUSER` value is used if there is no `--dbuser` passed in, but the new value is used if one is passed in.
(same goes for `$PGPASSWORD`)

related to:
- https://bugzilla.redhat.com/show_bug.cgi?id=1632433
- ManageIQ/manageiq-appliance_console#65

### DIFF
--- a/lib/gems/pending/util/postgres_admin.rb
+++ b/lib/gems/pending/util/postgres_admin.rb
@@ -249,7 +249,7 @@ class PostgresAdmin
     {
       "PGUSER"     => opts[:username],
       "PGPASSWORD" => opts[:password]
-    }
+    }.delete_blanks
   end
   # rubocop:disable Style/SymbolArray
   PG_DUMP_MULTI_VALUE_ARGS = [


NickLaMuro:
@kbrock I am fine with this PR, but I am questioning the `bug` label.

I have backups/dumps working on every provider we have fully implemented in an automated test suite without this change, so I am curious what is different with your setup that is making this necessary for things to work.

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -249,7 +249,7 @@ def self.runcmd_with_logging(cmd_str, opts, params = {})
     {
       "PGUSER"     => opts[:username],
       "PGPASSWORD" => opts[:password]
-    }
+    }.delete_if { |n, v| v.blank? }
```
carbonin: Do we have `delete_blanks` here? I can't remember if that's activesupport or morecore ...

Also, can you fix the code climate/bot comment about the unused argument?

kbrock:
@NickLaMuro I always thought you put BUG when you had a BZ. even if the BZ is an RFE. I can change

lib/gems/pending/util/postgres_admin.rb
kbrock: `more_core_extensions` is in the gemspec, so it should be good. (I always avoid morecore but in this case, it is good. thnx)

simaishi:
Hammer backport details:

```
$ git log -1
commit 9a5d8a05a209b25531fa042db96dc03dda8be568
Date:   Mon Oct 8 09:07:03 2018 -0400

    Merge pull request #378 from kbrock/env_fixup
    
    Don't set empty ENV values for database dumps
    
    (cherry picked from commit 3f6a759e18cab9e162e88792bbe3977d64180596)
```

PR #379, opened 2018-10-05
Add support to send data on standard input when executing a command

fdupont-redhat:
When running a command via SSH, one may need to send data on the standard input of the command. This PR adds a `stdin` argument to `exec` and `suexec` methods, with a default value of `nil`. The data is sent only if this is not nil, so current calls to these methods should not be impacted.

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -47,7 +47,7 @@ class MiqSshUtil
     end
   end
 
-  def exec(cmd, doneStr = nil)
+  def exec(cmd, doneStr = nil, stdin = nil)
     errBuf = ""
     outBuf = ""
     status = nil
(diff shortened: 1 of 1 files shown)


carbonin:
Can you reference a WIP patch where this would be used? It's hard to review without seeing the use case.

fdupont-redhat:
@carbonin it would be used like this https://github.com/ManageIQ/manageiq/blob/7de90d65d5069ad4a1cb2a9eb3695b739ad00f3d/app/models/conversion_host.rb#L47.

In Automate, I had implemented a method similar to `exec` and `suexec`, and allowing passing data on stdin: https://github.com/ManageIQ/manageiq-content/blob/master/content/automate/ManageIQ/Transformation/TransformationHosts/ovirt_host.class/__methods__/utils.rb#L11-L42.

carbonin:
@fdupont-redhat Does this need to go back to the Hammer release?

fdupont-redhat:
@carbonin Yes, it does. https://github.com/ManageIQ/manageiq/pull/18064 expected to be backported to Hammer, so this one should too.

simaishi:
Hammer backport details:

```
$ git log -1
commit 4bdf685aa5d00c518fc0a249b23cd7bba6b6cc16
Date:   Mon Oct 8 09:18:40 2018 -0400

    Merge pull request #379 from fdupont-redhat/add_support_for_stdin_to_ssh_exec
    
    Add support to send data on standard input when executing a command
    
    (cherry picked from commit 7cfbbb04a4d6c6b6e4c2b18cd1553d33d689b9b7)
```

PR #381, opened 2018-10-10
Leverage run_session for SFTP

fdupont-redhat:
This PR leverages the `run_session` method to get a Net::SSH::Connection::Session object that can be used to access the SFTP subsystem. This standardize connection via SSH.

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -39,9 +39,9 @@ class MiqSshUtil
   end # def initialize
 
   def get_file(from, to)
-    Net::SFTP.start(@host, @user, @options) do |sftp|
+    run_session do |ssh|
       $log.debug "MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}." if $log
-      data = sftp.download!(from, to)
+      data = ssh.sftp.download!(from, to)
       $log.debug "MiqSshUtil::get_file - Copying of #{@host}:#{from} to #{to}, complete." if $log
       return data
     end


lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -39,9 +39,9 @@ def initialize(host, user, password, options = {})
   end # def initialize
 
   def get_file(from, to)
-    Net::SFTP.start(@host, @user, @options) do |sftp|
+    run_session do |ssh|
       $log.debug "MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}." if $log
-      data = sftp.download!(from, to)
+      data = ssh.sftp.download!(from, to)
```
carbonin: This seems like an odd change, and I can't find documentation for getting to an sftp connection using the ssh session object and also this:

```ruby
[7] pry(main)> Net::SSH::Connection::Session.instance_methods.include?(:sftp)
=> false
[8] pry(main)> Net::SSH::Connection::Session.methods.include?(:sftp)
=> false
```

I trust that you've tested this, but I'm not comfortable making this change which doesn't really bring any upside when I can't find this behavior documented anywhere.

carbonin: That said, if you can find the documentation for this, then this looks good to me.

fdupont-redhat: I found it here https://net-ssh.github.io/net-sftp/classes/Net/SSH/Connection/Session.html.

carbonin: Ah okay, I needed to also `require 'net/sftp'` because it patches.

```ruby
[2] pry(main)> require 'net/ssh'
=> true
[3] pry(main)> Net::SSH::Connection::Session.instance_methods.include?(:sftp)
=> false
[4] pry(main)> require 'net/sftp'
=> true
[5] pry(main)> Net::SSH::Connection::Session.instance_methods.include?(:sftp)
=> true
```

Makes sense to me then :+1:

carbonin:
@fdupont-redhat is there anything that needs this change for the Hammer branch? I'm not sure that this actually adds any functionality, so I'm tempted to label `hammer/no`.

fdupont-redhat:
@carbonin AFAIK, there's no code depending on that change in Hammer. So ok for `hammer/no`.

PR #384, opened 2018-10-16
Be more lenient for locked down ftp servers

kbrock:
When uploading to an anonymous ftp server, sometimes ls (aka `nlst`) is disabled.

This logs that case and continues on

this is part of anonymous ftp feature
https://bugzilla.redhat.com/show_bug.cgi?id=1632433

@miq-bot add_label enhancement

### DIFF
--- a/lib/gems/pending/util/miq_ftp_lib.rb
+++ b/lib/gems/pending/util/miq_ftp_lib.rb
@@ -56,7 +56,11 @@ module MiqFtpLib
       end
       ftp.chdir(directory)
     end
-    ftp.chdir(pwd)
+  rescue Net::FTPPermError
+    raise unless @username.nil? && @password.nil?
+    _log.info("introspection of directories disabled. skipping create_directory_structure")
+  ensure
+    ftp.chdir(pwd) if pwd
   end
 
   def with_connection(cred_hash = nil)


lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -56,7 +56,10 @@ def create_directory_structure(directory_path)
       end
       ftp.chdir(directory)
     end
-    ftp.chdir(pwd)
+  rescue Net::FTPPermError => e
+    _log.info("introspection of directories disabled. skipping create_directory_structure")
```
carbonin: If we skip this where will the dump end up on the server? Is it really better to put it somewhere the user didn't expect rather than fail here?

lib/gems/pending/util/miq_ftp_lib.rb
```diff
@@ -56,7 +56,10 @@ def create_directory_structure(directory_path)
...
+  ensure
+    ftp.chdir(pwd) if pwd
```
NickLaMuro: Can we remove the `if pwd` here?  The `pwd` should always be set, so want to always make sure we are returning to that directory.

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: So I had a alternative way of doing this:

```ruby
raise e unless @username.nil? && @password.nil?
```

Which would only attempt to do this on anonymous FTP logins.

* * *

Also, notice my comment below.  The assumption was always to add the file from the logged in directory, and this code still returns to that directory once it is done adding directories.

kbrock: the dump ends up on the server - I've double checked with support that it was going through

lib/gems/pending/util/miq_ftp_lib.rb
kbrock: I feel more comfortable having the `if pwd`

It shouldn't be nil. but there is a chance that it gets nil if the first line throws an exception.

with the `if` then this line will be ignored, but the original exception will continue through.
if we remove the `if` here, then this will throw an unrelated (and meaningless(?) exception) - hiding the original exception

NickLaMuro: I am fine with that explanation.  Not standard practice I would guess, but makes sense to me why you are doing this now, so keep it as is.  👍

lib/gems/pending/util/miq_ftp_lib.rb
NickLaMuro: So I am good with this update.

@carbonin is this clear to you?  Basically, we always return to the `root` dir after creating the directory structure, then pass the full path for the file when doing the upload, which will make sure the file is in the proper place.

I am testing that this is the case in "my test" suite as well (though, maybe I should disable `nlst` on there as well...)

simaishi:
Hammer backport details:

```
$ git log -1
commit 59fc6803591f731a1b5a98fe2f2e8f66b8856063
Date:   Thu Oct 18 15:29:29 2018 -0400

    Merge pull request #384 from kbrock/lenient_ftp
    
    Be more lenient for locked down ftp servers
    
    (cherry picked from commit 21e666fa58c2b6121c76056bddef811c27bbec31)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1632433
```

PR #385, opened 2018-10-18
Adds full specs for restoring pg_dump and pg_basebackup backups

NickLaMuro:
This lays the ground work for some further refactoring changes to the restore tasks in `PostgresAdmin` by adding specs to confirm existing functionality.

Architecture/Design
-------------------

This PR is targeting two different environments:

- Local developer machines
- Travis CI

Each has their own assumptions for their setups, and thus, require different configuration to work properly.

_**NOTE:**_  It has already been discussed that we are probably changing this approach, but this is just a documentation of what is done here.  Other later issues/pull requests might provide the current implementation.

### ----- On Local Developer machines -----

For the local developer, it is assumed that:

- `postgres` is installed on the host machine with client libs
- `pg_ctl` is available on the user's `PATH`
- An existing `postgres` database instance in which should be kept intact and not disturb other databases
  - (A restore of a `pg_basebackup` will destroy ALL databases, not just one being restored)

With this, we then approach this environment by creating a new database instance targeted to the test run, and clean shut it down and clean it up when we are finished.

#### `PostgresRunner`

To control this, `PostgresRunner` (`spec/postgres_runner_helper.rb`) is in charge of:

- Running `initdb` for this new DB
- Initializing any database `ROLES` to match an appliance environment
- Fix an configuration problems that don't work locally prior to start
- Create/Start/Stop/Destroy the `postgres` instance

`PostgresRunner` was written similar to an init script, and it's class was written to be a drop in replacement for a `LinuxAdmin::Service` instance (it is a `Singleton`, so the `PostgresRunner` class equals a `LinuxAdmin::Service` instance).  It will be loaded as part of the `spec:dev` task only, or if specifically required when running `bundle exec rspec`.  It is also configured that these new database specs won't run unless the `postgres_runner_helper.rb` has been loaded, and running `rake spec` won't do that by default.

#### Stubbed configs

For testing database restores, the configuration file is whatever was on the appliance when the backup was taken, so we need to make some slight edits to it to make sure the process can reboot.  We do this by removing any reference to `/etc/manageiq/postgresql.conf.d` in the `postgresql.conf`.

We also need to borrow one configuration setting from that appliance config, which is the `ssl = on` config so the `pg_hba.conf` works with a restore.

Things are also configured so that the specs for running both the "dump restore" and the "basebackup restore" can happen in either order, so the `PgEnvironmentUpdater` is used to do just that.  That said, this code is meant to work on both the local and Travis.

### ----- On Travis -----

With Travis, there are some different assumptions being made:

- The database can be blown away without issue
- We are using the baked in Travis `postgres` addon
- We are using the startup tools configured by Travis
  - Not compatible with `LinuxAdmin` currently

This does allow us to treat some things more like an appliance, such as we create a stub `etc/manageiq/postgresql.conf.d` directory instead of editing the configs (though, this ends up still being tweaked for other reasons).

But because some indirect assumptions that the user has root permissions are being made when running `PostgresAdmin.restore` with `basebackups`, as well as some weirdness with Travis' `postgres` configuration, this means a few tricks were also required to make things work properly.

#### Triggering a `.restore` (with `sudo`)

To start with, when we doing stubbing for the specs on a dev environment, we mostly just change what the user/group will be, where the pg directory is, and stub in the `PostgresRunner` for the `LinuxAdmin::Service`.  This works fine since the user running the specs created the pg directory in the first place.

But for Travis, that isn't the case, and the directories are owned by a proper `postgres` user and the specs are run by the `travis` user.  This user does have `sudo` permissions, but `PostgresAdmin` doesn't have that privileged from the spec process.  To handle that, we end up stubbing `PostgresAdmin.restore` as part of the `spec/support/contexts/db_restore.rb` and triggering a `sudo` subprocess that ends up calling the same method.

This sub process loads the barebones ruby process with just the necessary libs required to run `PostgresAdmin.restore`.  This includes handling stubbing `LinuxAdmin::Service` (via a monkey patch), which is done by loading `spec/ci_helper.rb`, which has a lightweight `CiPostgresRunner`, which simply handles starting and stopping `postgres` during the `.restore` method.

#### Dealing with the split configs/data

Part of what Travis does with it's configuration of `postgres` is that it uses `RAMFS` data directory, which I can only assume allows for faster... something?

That said, basically this is configured on the `init` of the `postgres` process if the directory doesn't exist, which just copies the `/var/lib/postgresql` to the `/var/ramfs/postgresql` directory.  There still seems to be some connection between the `/var/ramfs` dir and the one in `/var/lib`, so restores still assume that is the directory, so we `rm -rf` the directory prior triggering a start, allowing the init script to regenerate it based off the `/var/lib` dir (so with our data).

Also, our configs (`postgresql.conf` and the `pg_hba.conf`) are located in the data directory.  For Travis, they live in the `/etc/postgresql/` directory.  Since `PostgresAdmin` doesn't take this into account, we just move these files from our data dir over prior to starting the process again.

#### Config file editing

So because don't need to worry about blowing away data being used for development, we can use the default `postgres` instance and also just create a `etc/manageiq/postgresql.conf.d` directly on the system without having to worry about cleaning it up afterwards.

We do with `create_stub_manageiq_configs_on_ci` in the `PgEnvironmentUpdater` module prior to running the test suite.  We also add a small config file that has `ssl = on`, which is the only real config we need to set to get things working from our appliance configs.

That said, while it would have been convenient to set `data_directory = '...'` in the file as well, this doesn't work with Travis' setup.  It is setup to use multiple versions of `postgres` and using `pg_wrapper`, it can enable one via the yaml file (which we do).  But it seems that `pg_wrapper` won't respect `include_dir` configurations, so setting `data_directory = '...'` from within the stubbed config would cause the instance to not boot up after a restore.

The above is addressed by appending a `data_directory = '...'` to the end of our `postgresql.conf` that was pulled from the `basebackup`.  It isn't completely true to the restore, but it was the best way to work with the existing system.


NickLaMuro:
@miq-bot assign @carbonin 

@carbonin I still need to test things against travis, and probably flesh things out in the PR description on how this is intended to be run, but hopefully this makes some sense as to what is being done here.

I realize there is a lot of code, and possibly some of it could be trimmed down, but wanted to make this in a way that could be run on both a developer's machine and travis while still using backups that where generated on an appliance, which wasn't terribly easy to figure out and accomplish.

Let me know what needs further clarification, because I am sure there is at least a few things.

spec/support/mixins/pg_root_role_creator.rb
```diff
@@ -0,0 +1,37 @@
...
+module PgRootRoleCreator
+  def create_root_role
+    with_connection do |conn|
+      pass = conn.escape_string("smartvm")
+      conn.exec("CREATE ROLE root WITH LOGIN CREATEDB SUPERUSER PASSWORD '#{pass}'")
+    end
+  end
+  module_function :create_root_role
+
+  def create_postgres_role
```
carbonin: Is this not always there by default?

NickLaMuro: I don't think so.  I would have to research to give you a definitive answer, but my guess is why this was there in most cases was because the rpm for postgresql that we use probably also creates a postgres user, and thus, as part of it's `initdb` step, automatically creates the `postgres` role.

Locally, in my case, I end up with a `nicklamuro` role by default, since that is the user that I am running the specs as, so when the dump file gets to these lines:

```SQL
REVOKE ALL ON SCHEMA public FROM PUBLIC;
REVOKE ALL ON SCHEMA public FROM postgres;
GRANT ALL ON SCHEMA public TO postgres;
GRANT ALL ON SCHEMA public TO PUBLIC;
```

It falls over and dies.

NickLaMuro:
> You're still working on running these using the built-in postgres on Travis, right?

Yup, that was something that I was going to attempt "first try", but expected I would need a few attempts to get it working right.

NickLaMuro:
Kicking travis by closing/opening...

NickLaMuro:
_@NickLaMuro realizes that it probably was the GitHub outage that is causing the lack of Travis jobs..._

NickLaMuro:
@carbonin Okay, I think this is finally in a place where I am ready for a proper review.  🍅 🍅

spec/postgres_runner_helper.rb
```diff
@@ -0,0 +1,142 @@
...
+
+  def pid_file
+    File.join(db_dir, "postmaster.pid")
+  end
+
+  def initdb
+    dirname = Dir::Tmpname.make_tmpname("postgres_admin", "data")
+    @db_dir = File.expand_path(dirname, __dir__)
+
+    system("#{pg_ctl} init -D #{db_dir} -o '-A trust'", :out => File::NULL)
```
carbonin: Would specifying `-U` to `initdb` here simplify some things around setting up users? If we always initialize the database with a consistent superuser then maybe we don't have to work around how things will work differently locally vs on travis.

spec/postgres_runner_helper.rb
```diff
@@ -0,0 +1,142 @@
...
+  end
+
+  # TODO:  NickLaMuro: Maybe just don't get a backup from the appliance?
+  #
+  # Unsure if it is better to have a slightly "different from prod" backup, or
```
carbonin: I would prefer keeping this generic rather than having to work around strangeness in the appliance backups.

spec/postgres_runner_helper.rb
```diff
@@ -0,0 +1,142 @@
...
+  def set_env
+    @user = self.class.user # used by PgEnvironmentUpdater
+    ENV["APPLIANCE_PG_DATA"]    = db_dir
+    ENV["APPLIANCE_PG_SERVICE"] = "local_pg_instance"
+  end
+
+  private
+
+  def set_port
+    @port = TCPServer.open("127.0.0.1", 0) { |sock| sock.addr[1] }
```
carbonin: Do we really have to worry about ports? I would figure that we would be connecting to the local socket rather than over the network.

spec/support/contexts/db_restore.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+end
+
+# Required that you define `db_name` yourself in your specs
+shared_context "Database Restore Validation Helpers" do
+  let(:active_record_connection_info) do
+    {
+      "adapter"  => "postgresql",
+      "encoding" => "utf8",
+      "host"     => "localhost",
```
carbonin: Again, I feel like connecting using the socket would simplify this. Or does that not work if multiple servers are running on the same system?

spec/support/contexts/db_restore.rb
```diff
@@ -0,0 +1,96 @@
+require "active_record"
```
carbonin: Were we previously using active record in this gem? I wonder if we could get away with using PG instead?

Rakefile
```diff
@@ -7,6 +7,15 @@ RSpec::Core::RakeTask.new(:spec) do |t|
   t.rspec_opts = "--options #{File.expand_path(".rspec_ci", __dir__)}" if ENV['CI']
 end
 
+desc "Run RSpec code examples (with local postgres dependencies)"
+RSpec::Core::RakeTask.new("spec:dev") do |t|
```
carbonin: Why `spec:dev` for the name here? Maybe `spec:full` or something that indicates that we are including additional tests?

spec/util/postgres_admin_spec.rb
```diff
@@ -1,6 +1,40 @@
...
+        restore_opts[:dbname] = "pg_dump_restore_of_simple_db"
+        PostgresAdmin.restore(restore_opts)
+
+        expect(author_count).to eq(2)
+        expect(book_count).to   eq(3)
+      end
+    end
+
+    context "with a pg_basebackup file" do
+      include_context "Database Restore Validation Helpers"
```
carbonin: Could we move this to the `describe` level?

spec/ci_helper.rb
```diff
@@ -0,0 +1,57 @@
...
+  PGVER       = "9.5".freeze
+  PGDATADIR   = "/var/lib/postgresql/#{PGVER}/main".freeze
+  PGCONFIGDIR = "/etc/postgresql/#{PGVER}/main".freeze
+
+  def self.start
+    # Make sure we have our postgresql.conf in the right spot on Travis
+    system("cp #{PGDATADIR}/postgresql.conf #{PGCONFIGDIR}/postgresql.conf")
+
+    # Move our configured `pg_hba.conf` to the config dir as well
+    system("cp #{PGDATADIR}/pg_hba.conf #{PGCONFIGDIR}/pg_hba.conf")
```
carbonin: Why do we need to do these? I would think the built-in configuration would work for this.

spec/support/contexts/db_restore.rb
```diff
@@ -86,11 +86,40 @@ def set_spec_env_for_postgres_admin_basebackup_restore
...
+      }
 
-      allow(LinuxAdmin::Service).to receive(:new).with(PostgresAdmin.service_name)
-                                                 .and_return(CiPostgresRunner)
+      dir_opts     = "-I #{RestoreHelper::SPEC_DIR} -I #{RestoreHelper::LIB_DIR}"
+      require_opts = "-r linux_admin -r ci_helper -r gems/pending/util/postgres_admin"
+      ruby_eval    = "-e 'PostgresAdmin.restore(#{RestoreHelper.restore_backup_opts.inspect})'"
+      cmd          = "sudo #{Gem.ruby} #{dir_opts} #{require_opts} #{ruby_eval}"
+
+      allow(PostgresAdmin).to receive(:restore).and_wrap_original do |_, *_args|
```
carbonin: Why use `and_wrap_original` here rather than `and_return` if you aren't using the yielded params?

spec/support/mixins/pg_environment_updater.rb
```diff
@@ -22,6 +22,33 @@ def create_postgres_role
...
+  # This is fine to do on CI as it is an ephemeral VM, so it isn't going to
+  # mess with state at all of a persisted machine (like creating this on
+  # someone's dev box would).
+  #
+  # This does not add in all of the configs that you would find on an
+  # appliance, however, and just defines enough to get things working (so
+  # basically just 'ssl = on').
+  #
+  # See PostgresRunner#fix_pg_conf for how we handle this locally.
+  def create_stub_manageiq_configs_on_ci
```
carbonin: Why are we doing this? Does the restore for the base backup fail if this doesn't exist? Or is this just to be as close to prod as possible?

spec/support/contexts/db_restore.rb
NickLaMuro: Barely, but yes:

https://github.com/ManageIQ/manageiq-gems-pending/blob/86396bfb/lib/gems/pending/util/extensions/miq-array.rb#L26

I am fine switching this to just using `PG` for these specs, in this section specifically, I knew the `ActiveRecord` interface better.

Rakefile
NickLaMuro: Because these are specific changes that are for a developers machine, where `spec:ci` is the CI specific changes.

Now, in retrospect, seeing that it wasn't super straight forward to get this working on Travis like I was hoping, I almost think it makes more sense to have local and CI doing the same process with `pg_ctl` instead.  But unsure if that is worth the effort at this point since there are a bunch of other PRs dependent on this one I have in the pipeline.

spec/util/postgres_admin_spec.rb
NickLaMuro: Probably, I will look into that.

spec/ci_helper.rb
NickLaMuro: This is something I really needed document better in the PR description... and probably in the code.

Travis uses a Debian image, and they use a filesystem (feature?) called RAMFS, which I assume is to improve performance on boot or some other performance reason.  From what I found, it still ends up pointing back to the data directory in some fashion, but these files need to exist in the `/var/ramfs/...` dir on first boot for the `postgres` process to read that, instead of the one that existed previously (which wouldn't be a great example of a restore if we didn't).

Again, possibly using the `pg_ctl` method would have been wiser for CI...

spec/support/contexts/db_restore.rb
NickLaMuro: Few things with this (and I should probably add a comment around this:

First off, it is assumed with `PostgresAdmin` the we have `elevated` permissions to do most of the tasks in a restore.  I am even pretty sure that the `appliance_console` alias on an appliance is just a wrapper that includes calling `sudo /path/to/manageiq-appliance_console/bin/...` as well.  If I don't do this with `sudo`, basically all of the `FileUtils` commands will fail.

The commands do actually still run `PostgresAdmin.restore` with the same (see the `ruby_eval =` line above), but I just decided to pull the opts from the `RestoreHelper` instead of from the `*_args`.  Could have gone either way.

The other reason we use `and_wrap_original` is I wanted to do a `yield`, and not a static return that happened on execution of this line.  If I used return, it would have run this prior to it being called later in the spec.  This is probably pedantic, but it was mostly to do the best to simulate actually calling the method by instead call it through a elevated sub process with the limited number of files necessary to load `PostgresAdmin` and run `.restore`.

spec/support/mixins/pg_environment_updater.rb
NickLaMuro: > Does the restore for the base backup fail if this doesn't exist? Or is this just to be as close to prod as possible?

Both.  `postgres` won't even start up if this directory doesn't exist, and ideally I was trying to have this match prod as much as possible (but again, this seems like it was a very futile effort).

Rakefile
NickLaMuro: > Why spec:dev for the name here?

Because this was meant to specify that this specifically designed to run on a development machine, where it would be creating a separate instance of postgres as part of the test run to do it.  I feel like the `:default` task is the `spec:full` task as you are describing it.

spec/ci_helper.rb
NickLaMuro: Because our appliance configuration actually sets `ssl = on`, and I figured that would be disingenuous if we didn't try reconnecting with SSL through the port connection.

Rakefile
carbonin: Yeah, that makes sense in context of the ci task. I was reviewing this commit-by-commit so at the time this was created the naming seemed odd.

spec/postgres_runner_helper.rb
NickLaMuro: Well, again, on Travis (currently), we are trying to use the existing database instance, where locally we are trying to not blow away the developers data when doing a base backup restore.  If we wanted to keep this the same, basically both Travis and local would have to use the `pg_ctl` method, which is actually probably fine.  I tried not going that route originally because I did want to try working with an existing database when doing a restore when possible.

That said, this would probably avoid all of the `sudo ruby -r postgres_admin` stuff I do, so this might not be a bad idea.  The `-U` would be unnecessary, though, as we would just use the user that is running the specs be the owner of the DB and leave it at that.

spec/postgres_runner_helper.rb
NickLaMuro: Does this make sense doing that knowing that `PostgresAdmin` makes a bunch of assumptions about it being on an appliance then?  This was kind of what I was struggling with when putting this together.

For example, when restoring a `pg_basebackup`, we just inherit whatever users created for the backed up system.  So to connect to it properly in an idempotent way, I preload them as part of the mixin, and always connect with those same users (so basically, it doesn't matter if the specs run the `pg_dump` restore first, or the `pg_basebackup` restore first).

spec/postgres_runner_helper.rb
NickLaMuro: On dev, there would be a port conflict if we left it to the defaults (I am pretty sure), and `postgres` wouldn't boot.

spec/ci_helper.rb
carbonin: Oh I see, this is not "our" configuration it's the config provided by the built-in PG on travis.

I misunderstood. I thought you were somehow grabbing the conf files that we install on appliances and trying to use those.

spec/postgres_runner_helper.rb
carbonin: I feel like the assumptions we're making are around environment rather than anything specific to our configuration though. Like we expect to run as root and expect that certain environment variables exist, but I don't think we need to have the include directory or run with ssl.

I think it's fine to test restore from a base backup taken from a clean cluster rather than deal with all the weirdness in our config files. And I'm hoping that that would simplify some things here.

spec/support/contexts/db_restore.rb
carbonin: Resolving this as it was answered in a different thread.

spec/support/contexts/db_restore.rb
carbonin: Yeah, I think I would prefer PG here. Then if we extract this from gems pending we don't need to bring AR along.

spec/ci_helper.rb
NickLaMuro: > Oh I see, this is not "our" configuration it's the config provided by the built-in PG on travis.

No, you were right with your first assumption (and clearly I shouldn't be responding to code review in the morning...).  Let me see if I can do a better job explaining...

In "our" setup, we have the database directory and the config directory in the same spot, so when we do a database restore by just unziping the tar, everything is ready to go.  On Travis, because there is a split between what is configured as the data directory and what is the config directory, this `cp` of files is necessary, otherwise we are booting `postgres` with old configs, which really isn't how we assume our restores work.

spec/support/contexts/db_restore.rb
carbonin: Makes sense, but I think we should use the yielded args because otherwise what we pass into `.restore` in the test isn't actually being validated, right? I know it's just the same method result now, but I want to make sure this still works if that ends up changing for some reason.

spec/support/mixins/pg_environment_updater.rb
carbonin: Yeah, I think we should resolve this by not using a backup from an appliance to test restore.

spec/ci_helper.rb
carbonin: Oh so this is happening after the restore. That makes more sense. This seems like something else that could be solved by starting up a server using pg_ctl in both cases as well. So this can go into the list of things to tackle in that follow-up refactoring.

spec/support/contexts/db_restore.rb
NickLaMuro: >  but I think we should use the yielded args...

No prob, and I was on the fence about this as well.  I will go ahead and change this.

NickLaMuro: Turns out that there is a better way to do this as well:

https://relishapp.com/rspec/rspec-mocks/v/3-8/docs/configuring-responses/block-implementation

So I will probably switch to using that as well.

NickLaMuro:
@carbonin addressed your most recent comment and added a large change to the PR description describing the architecture of the PR.  Hope it makes sense.

PR #386, opened 2018-10-23
bump fog-openstack to v0.3

gildub:
This is needed because other dependencies on fog-core.
For more details please see https://github.com/ManageIQ/manageiq/issues/17997

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -27,7 +27,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "aws-sdk",                 "~> 2.9.7"
   s.add_runtime_dependency "binary_struct",           "~> 2.1"
   s.add_runtime_dependency "bundler",                 ">= 1.8.4" # rails-assets requires bundler >= 1.8.4, see: https://rails-assets.org/
-  s.add_runtime_dependency "fog-openstack",           "~> 0.1.22"
+  s.add_runtime_dependency "fog-openstack",           "~> 0.3"
   s.add_runtime_dependency "linux_admin",             "~> 1.0"
   s.add_runtime_dependency "mime-types",              "~> 3.0"
   s.add_runtime_dependency "minitar",                 "~> 0.6"


carbonin:
@jerryk55 this won't cause any issues with the backup work, right?

carbonin:
@gildub why are we doing this? Is there a particular bug or feature what you're adding? Should this be backported to the Hammer release?

gildub:
@miq-bot add_label hammer/yes

gildub:
@carbonin, although `fog-openstack` has been added recently to `manageiq-gems-pending` it is also heavily used by `manageiq-providers-openstack`. V0.3.x contains massive refactor and we're targeting  next sub release.

I've added the flag for hammer backport.

carbonin:
>next sub release

What do you mean by that? Hammer is not GA yet, so this would be in the GA release, not a "sub release"

carbonin:
Okay, then this shouldn't be `hammer/yes` at this point that means that it would go into GA

gildub:
@carbonin, yeap, thanks.

gildub:
@carbonin, Well actually we still have time for Hammer GA because that's would be really a plus.
Besides hammer, https://github.com/ManageIQ/manageiq-providers-openstack/pull/334 is waiting on this one, it's a  blocker. Unless if you see any reason not to do you mind merging?

carbonin:
@agrare what do you think about this (and by extension https://github.com/ManageIQ/manageiq-providers-openstack/pull/334) going back to hammer?

I would rather not backport this version bump if we aren't going to end up using it.

carbonin:
@gildub I'll merge this to master for now and continue the conversation about backport.

agrare:
@carbonin we had planned on the new version of fog only going into ivanchuk and not hammer.

gildub:
@agrare, sorry I missed your last comment! The VCRs were ready or on their way, timing was the essence. I believe that having several repos depending on fog-openstack should be avoided and more likely gem-pending is the temporary lasting too long.

agrare:
@gildub Ah okay the last few comments sounded like Sam was going to record them for you

gildub:
@agrare, @mansam, the bump for hammer (for the record) is https://github.com/ManageIQ/manageiq-providers-openstack/pull/387. Thanks

PR #390, opened 2018-10-25
Handle pipes for pg_restore

NickLaMuro:
This is built off of #385

Better diff can be found here:

https://github.com/NickLaMuro/manageiq-gems-pending/compare/postgres_admin_restore_and_dump_specs...handle-piping-for-postgres-admin-restore-pg-dump

* * *

For whatever the reason, `pg_dump` doesn't like reading from a fifo when it isn't passed in as `STDIN`.  This assigns the same treatment to it as was done to `pg_dump`/`pg_restore`, which is passes it through `STDIN`.

Since we don't really have good access to STDIN via `AwesomeSpawn`, we use vanilla `Kernel.spawn` to handle the actually command execution.  We call out to `AwesomeSpawn` for the command builder, as well as emulate the errors that it would throw so there is limited change in how this library behaves to the user.

### DIFF
--- a/lib/gems/pending/util/postgres_admin.rb
+++ b/lib/gems/pending/util/postgres_admin.rb
@@ -173,8 +173,23 @@ class PostgresAdmin
 
   def self.restore_pg_dump(opts)
     recreate_db(opts)
+    args = { :verbose => nil, :exit_on_error => nil }
 
-    runcmd("pg_restore", opts, :verbose => nil, :exit_on_error => nil, nil => opts[:local_file])
+    if File.pipe?(opts[:local_file])
+      cmd_args   = combine_command_args(opts, args)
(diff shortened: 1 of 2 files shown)


NickLaMuro:
@miq-bot assign @carbonin 

While #385 needs to be merged for this to work, this could use a look as well.  Might add a test as well.

simaishi:
Hammer backport details:

```
$ git log -1
commit bd650c42eac0bacfcd33357744dad73f867255c6
Date:   Mon Nov 5 16:02:46 2018 -0500

    Merge pull request #390 from NickLaMuro/handle-piping-for-postgres-admin-restore-pg-dump
    
    Handle pipes for pg_restore
    
    (cherry picked from commit a4b1ae0be59297f9610b87d562425c2ac2b16682)
```

PR #396, opened 2018-11-02
[MiqFtpStorage] Support @byte_count in #download_single

NickLaMuro:
Since there is no way in the spec of `Net::FTP#getbinaryfile` to only download a specified number of bytes, it was required to rework a form of `Net::FTP#retrbinary` (a private method of that class) to allow only reading from the socket for the specified number of bytes, then moving on.

The goal was to remain consistent with `#add` and continue to use `IO.copy_stream`, a slight tweak that was necessary to get that working.  The `conn` variable is a `BufferedSocket`, which basically has IO like, but wraps an existing `@io` with some methods and tries to maintain a IO-like class interface.

Unfortunately, it's `#read` implementation doesn't have the same number of args as `IO#read`:

```
(method definitions)

IO#read( [ length, [ output ] ] )

BufferedSocket#read( [ len ] )
```

When `IO.copy_stream` calls `#read` under the covers, it does pass in the second argument of `output` to the result, which in turn causes an error if we just pass `conn` as the first argument to `copy_stream`.

The underlying `@io` attribute for `conn` does actually adhere to the argument interface of `copy_stream`, and since we don't miss anything by working around it, we just used that instead for our use here.  This should, however, probably be fixed in ruby's stdlib for `Net::FTP`.

### DIFF
--- a/lib/gems/pending/util/object_storage/miq_ftp_storage.rb
+++ b/lib/gems/pending/util/object_storage/miq_ftp_storage.rb
@@ -54,7 +54,18 @@ class MiqFtpStorage < MiqObjectStorage
   end
 
   def download_single(source, destination)
-    ftp.getbinaryfile(uri_to_relative(source), destination)
+    ftp.synchronize do
+      ftp.send(:with_binary, true) do
+        begin
+          conn = ftp.send(:transfercmd, "RETR #{uri_to_relative(source)}")
+          IO.copy_stream(conn.io, destination, byte_count)
+          conn.shutdown(Socket::SHUT_WR)
+          conn.read_timeout = 1
+        ensure
+          conn.close if conn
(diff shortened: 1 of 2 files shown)


NickLaMuro:
@miq-bot assign @carbonin

simaishi:
Hammer backport details:

```
$ git log -1
commit ce36ff1cb768985ebcfa20ac8f0735d225eeb57d
Date:   Mon Nov 5 12:16:17 2018 -0500

    Merge pull request #396 from NickLaMuro/support-byte-count-for-ftp-object-stores
    
    [MiqFtpStorage] Support @byte_count in #download_single
    
    (cherry picked from commit 9fc9564cfed3e8f6d29108519c567dcd899de06e)
```

PR #398, opened 2018-11-02
Fixes and cleanup for MiqSwiftStorage

NickLaMuro:
- Fixes a bug with `@container_name` generation (didn't properly handle nested keys
  - "Simplifies" the code a bit too (depends on your affinity for regular expressions I guess...)
- Aligns comments to a standard 80 character width, with a few tweaks.

### DIFF
--- a/lib/gems/pending/util/object_storage/miq_swift_storage.rb
+++ b/lib/gems/pending/util/object_storage/miq_swift_storage.rb
@@ -16,12 +16,15 @@ class MiqSwiftStorage < MiqObjectStorage
     @bucket_name = URI(@settings[:uri]).host
 
     raise "username and password are required values!" if @settings[:username].nil? || @settings[:password].nil?
-    _scheme, _userinfo, @host, @port, _registry, @mount_path, _opaque, query, _fragment = URI.split(URI.encode(@settings[:uri]))
+    _scheme, _userinfo, @host, @port, _registry, path, _opaque, query, _fragment = URI.split(URI.encode(@settings[:uri]))
     query_params(query) if query
     @swift          = nil
     @username       = @settings[:username]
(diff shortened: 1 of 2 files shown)


lib/gems/pending/util/object_storage/miq_swift_storage.rb
```diff
@@ -16,12 +16,15 @@ def initialize(settings)
     @bucket_name = URI(@settings[:uri]).host
 
     raise "username and password are required values!" if @settings[:username].nil? || @settings[:password].nil?
-    _scheme, _userinfo, @host, @port, _registry, @mount_path, _opaque, query, _fragment = URI.split(URI.encode(@settings[:uri]))
```
carbonin: Was the `@mount_path` variable unused before? Seems odd that it was being assigned here if we didn't need it. @jerryk55 ?

carbonin: To clarify, I see that it was being used to set `@container_name`, but because it's an instance variable (rather than a local) I'm being cautious about it being used elsewhere.

carbonin: Just took a closer look. Looks like this is only used in mount sessions (rather than object stores).

NickLaMuro: Yeah, we should probably remove the `@mount_path` and just make it a local variable so it can be garbage collected, but I decided to leave it for now.  Can do a follow up later to fix.

NickLaMuro: Oh, I just looked at the code I wrote, and I did do that... ignore me...

PR #399, opened 2018-11-02
[MiqS3Storage] Support @byte_count in #download_single

NickLaMuro:
Built off of https://github.com/ManageIQ/manageiq-gems-pending/pull/397 (this branch will most likely need to be rebased after that is merged)

Simply supports byte_count by passing the `:range` attribute to `#get_object`, as defined here:

https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#get_object-instance_method

Also, will require support `StringIO` in the future, so that change was made here as well.

Links
-----

* Source branch:  https://github.com/ManageIQ/manageiq-gems-pending/pull/397
* `aws-sdk` docs:  https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Aws/S3/Client.html#get_object-instance_method

PR #397, opened 2018-11-02
Fixes some bugs with MiqS3Storage#download_single

NickLaMuro:
The following have been addressed:

- Usage of `remote_file` (wrong variable name) switched to `source` (correct)
- Removes unassigned variable `local_file` from being the return value (mistakenly copied from #upload_single`).
  - The return value isn't currently used for this method, so it has not been replaced.
- Force encoding of the streamed chunks to match the encoding of the destination IO

PR #400, opened 2018-11-03
Add MiqSwiftStorage#download_single

NickLaMuro:
Currently built off of multiple dependent PRs.  See "Links" section below.

This does the following:

- Moves code in `MiqS3Storage#download_single` for handling chunk writing into `MiqObjectStorage#write_chunk_proc`
- Implements `MiqSwiftStorage#download_single` in a similar fashion to `#add`
- Updates the implementation to also support `byte_count`

Worth noting is that `#write_chunk_proc` uses a `Proc.new` instead of a `lambda` to work around the arg checking of `lambda`.  Both `aws-sdk` and `excon` (used by `fog`) handle response blocks where it's first argument is `chunk`, but `Excon` adds an additional two, which are include some extra information about the amount of remaining bytes and total bytes, which we are not concerned about.  Since we can reuse this, for now, we are.

Implementation for support `byte_count` also pretty much mirrors that of `S3`, it just requires setting the header value directly, versus a support key as done in `aws-sdk`.

Links
-----

- Built off of:
  - https://github.com/ManageIQ/manageiq-gems-pending/pull/397
  - https://github.com/ManageIQ/manageiq-gems-pending/pull/398
  - https://github.com/ManageIQ/manageiq-gems-pending/pull/399
- Openshift Swift API docs:  https://developer.openstack.org/api-ref/object-store/?expanded=get-object-content-and-metadata-detail


NickLaMuro:
@miq-bot assign @carbonin 
@miq-bot remove_label unmergable

NickLaMuro:
... 

😩 

@miq-bot remove_label unmergeable

lib/gems/pending/util/miq_object_storage.rb
```diff
@@ -45,6 +45,17 @@ def write_single_split_file_for(file_io)
     clear_split_vars
   end
 
+  def write_chunk_proc(destination)
+    # We use a `proc` here instead of `lambda` because we are only concerned
+    # about the first argument, and other arguments (additional ones added by
+    # Excon's response_call signature, for example) are unneeded.
+    #
+    # `lambda do` will do argument checking, while `proc do` won't.
+    proc do |chunk|
```
carbonin: What are `chunk` and `destination` here? Is this some kind of common object?

lib/gems/pending/util/object_storage/miq_swift_storage.rb
```diff
@@ -78,6 +78,26 @@ def upload_single(dest_uri)
     end
   end
 
+  def download_single(source, destination)
+    #
+    # Get the remote path, and parse out the bucket name.
```
carbonin: I don't think we need a comment here

lib/gems/pending/util/miq_object_storage.rb
NickLaMuro: `destination` is the variable passed in from the method arg...

lib/gems/pending/util/object_storage/miq_swift_storage.rb
NickLaMuro: Fair enough.  I think I copied this from `s3`, so I was clearly using the wrong terminology as a result.  Will remove.

lib/gems/pending/util/miq_object_storage.rb
NickLaMuro: `chunk` on the other hand is just the common var name used by both `aws-sdk` and `fog`:

- https://docs.aws.amazon.com/sdkforruby/api/Aws/S3/Client.html#get_object-instance_method
  - (see the "Streaming data to a block" section)
- https://github.com/excon/excon#streaming-responses
  - (`Excon` is the HTTP lib for `fog`, by the same author)

So just reused that terminology.

carbonin: I understand where the variable names come from, I'm just surprised that the objects we're working with quack the same for both s3 and swift so I was curious what kinds of objects these were in those cases.

carbonin: Or are they just byte strings or something?

carbonin: Ah got it. `chunk` is a string and `destination` is an `IO`. Could have just looked it up a bit, sorry.

NickLaMuro: @carbonin yeah, sorry, the usage of this as a `&block` is probably also not making this easier to grok...

simaishi:
Hammer backport details:

```
$ git log -1
commit d7d8c163449bf41fbeb46911d92ea40c96cc27cf
Date:   Thu Nov 8 13:27:12 2018 -0500

    Merge pull request #400 from NickLaMuro/add-download-single-to-swift-object-store
    
    Add MiqSwiftStorage#download_single
    
    (cherry picked from commit 1525c2c5ad512efb860caa136bc8a52de853257a)
```

PR #402, opened 2018-11-05
Add :backup_type option to PostgresAdmin.restore

NickLaMuro:
Allows pre-fetching a backup_type when calling `PostgresAdmin.restore`, so no extra file type checking is done on the `opt[:local_file]` (which could be a streamed IO where that would break things down the line).

This behaves as it had if this option isn't passed.

PR #405, opened 2018-11-07
[PostgresAdmin] Fix backup_type being ignored

NickLaMuro:
This is a FIXUP to https://github.com/ManageIQ/manageiq-gems-pending/pull/402

With the previous implementation, the code would always attempt to find the magic number of the file for a `pg_dump` before checking if the `backup_type` value was instead a `:basebackup`.  When streaming a file, this would cause it to fail farther down the line since it would have strip some of the content off from the beginning of the file, invalidating it as a proper backup.

This change fixes by forcing the check for the backup type first, and then by inspecting the file contents.


NickLaMuro:
Going to add a test to this real quick, since this took me WAY to long to debug yesterday... 😩

NickLaMuro:
@miq-bot assign @carbonin
@miq-bot add_label bug, hammer/yes

PR #406, opened 2018-11-07
Use Gem::Package for tar pg restore unpacking via download streaming

NickLaMuro:
This requires https://github.com/ManageIQ/manageiq-gems-pending/pull/405 to be merged to function properly with files streams.

This change has a few benefits:

- It removes the dependency for an external gem, `minitar` (minor, but nice)
- It supports streaming files (important)

The first was simply a side affect, though one I admit I am happy with the result about, but unfortunately was a required change to allow me to do this without too much custom code.

`Archive::Tar::Minitar.unpack` does not support IO objects that don't use `rewind`, since it uses `Archive::Tar::Minitar::Input` under the hood:

- https://github.com/halostatue/minitar/blob/41c3ca65/lib/archive/tar/minitar.rb#L249
- https://github.com/halostatue/minitar/blob/41c3ca65/lib/archive/tar/minitar/input.rb#L6-L7

The main code that we wanted from the codebase, however, was the `extract_entry` method from `Input`:

- https://github.com/halostatue/minitar/blob/41c3ca65/lib/archive/tar/minitar/input.rb#L117

Which again, requires an `Input` model, and the check for if the `@io` is "seekable" is done on instantiation:

- https://github.com/halostatue/minitar/blob/41c3ca65/lib/archive/tar/minitar/input.rb#L76-L78

`Gem::Package.extract_tar_gz` method, however, only requires instantiating a instance with a "gem name", which doesn't even have to be a real gem, and then we have access to the method without worry, and can pass in our own `IO` object as we see fit.

That said, it isn't ideal, but the alternative would be re-writing one of these two methods:

- https://github.com/halostatue/minitar/blob/41c3ca65/lib/archive/tar/minitar/input.rb#L117-L194
- https://github.com/rubygems/rubygems/blob/master/lib/rubygems/package.rb#L381-L418

In which LJ probably wouldn't approve of all of the additions...

* * *

Of note, and a minor "fun fact":  The code for `minitar` is based on the same person's code, Mauricio Julio Fernández Pradier, as is found in `rubygems`:

- https://github.com/halostatue/minitar/blob/41c3ca65/README.rdoc#L37
- https://github.com/halostatue/minitar/blob/41c3ca65/Licence.md#L7
- https://github.com/rubygems/rubygems/blob/d43e4d01/lib/rubygems/package/tar_reader.rb#L4

### DIFF
--- a/lib/gems/pending/util/postgres_admin.rb
+++ b/lib/gems/pending/util/postgres_admin.rb
@@ -116,11 +116,17 @@ class PostgresAdmin
     pg_service.stop
     prep_data_directory
 
-    require 'zlib'
-    require 'archive/tar/minitar'
+    require 'rubygems/package'
 
-    tgz = Zlib::GzipReader.new(File.open(file, 'rb'))
-    Archive::Tar::Minitar.unpack(tgz, data_directory.to_s)
+    # Using a Gem::Package instance for the #extract_tar_gz method, so we don't
+    # have to re-write all of that logic.  Mostly making use of
+    # `Gem::Package::TarReader` + `Zlib::GzipReader` that is already part of
+    # rubygems/stdlib and integrated there.
+    unpacker = Gem::Package.new("obviously_not_a_gem")
+    File.open(file, IO::RDONLY | IO::NONBLOCK) do |backup_file|
+      unpacker.extract_tar_gz(backup_file, data_directory.to_s)
+    end
+
     FileUtils.chown_R(PostgresAdmin.user, PostgresAdmin.group, PostgresAdmin.data_directory)
 
     pg_service.start
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -30,7 +30,6 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "fog-openstack",           "~> 0.3"
   s.add_runtime_dependency "linux_admin",             "~> 1.0"
   s.add_runtime_dependency "mime-types",              "~> 3.0"
-  s.add_runtime_dependency "minitar",                 "~> 0.6"
   s.add_runtime_dependency "more_core_extensions",    "~> 3.4"
   s.add_runtime_dependency "net-scp",                 "~> 1.2.1"
   s.add_runtime_dependency "net-sftp",                "~> 2.1.2"


NickLaMuro:
@carbonin I want to double check it works with my test suite, which has been quite a pain to get working, just to make sure there isn't other issues that need to be addressed.  I will report back when it is.

NickLaMuro:
@gmcculloug @carbonin So I had a thought today when thinking about merging this one:

First off, I probably should remove `minitar` from the gemspec:

https://github.com/ManageIQ/manageiq-gems-pending/blob/b3f99e12/manageiq-gems-pending.gemspec#L33

but I was unsure if this was something that would be used by some automate scripts, and should stay in for the time being.  From what I can tell, we aren't using it anywhere else:

https://github.com/search?l=Ruby&q=org%3AManageIQ+%22minitar%22&type=Code
https://github.com/search?l=Ruby&q=org%3AManageIQ+%22Archive%22&type=Code

But unsure either of you have heard of this being used in `automate` for any reason.

NickLaMuro:

NickLaMuro:
@carbonin I think I am good with this now.  I got things working with my test suite (still have to push the commits for that though...), and I feel pretty confident with this.

Let me know if [any of the above](https://github.com/ManageIQ/manageiq-gems-pending/pull/406#issuecomment-437530459) should be addressed however.

carbonin:
@NickLaMuro I think we should probably remove this from the spec.

I added it fairly recently when we moved to basebackup-based backups so I don't feel like anyone should be relying on it being present.

gmcculloug:
@NickLaMuro To answer the automate question, we do not guarantee that any gems used by the application will remain on the system.  The automate methods run in a [Bundler.with_clean_env](https://github.com/ManageIQ/manageiq-automation_engine/blob/master/lib/miq_automation_engine/engine/miq_ae_engine/miq_ae_method.rb#L140) block which gives them access to the gems, but they are not locked into the same gems or versions, they can supply their own as needed.

simaishi:
Hammer backport details:

```
$ git log -1
commit e60c60cbc03c4b5ab0ac8102c933f4f1a037159d
Date:   Mon Nov 12 15:13:07 2018 -0500

    Merge pull request #406 from NickLaMuro/use-gem-package-to-stream-unpacking-database-backups
    
    Use Gem::Package for tar pg restore unpacking via download streaming
    
    (cherry picked from commit e145282690d50a31b33ba1e4285eb18e15fb04a4)
```

PR #410, opened 2018-11-09
Fixes to swift storage

NickLaMuro:
Adds a few fixes to the `MiqSwiftStorage`:

* Adds missing commit that was meant to be a part of #400 (eb533f49)
  - `#download_single` expected `#with_standard_swift_error_handling` to be there, and bombs without it
* Adds fix for `@container_name` regexp (242bf6ff)
  - Without this, this would cause all `MiqSwiftStorage` instances that weren't instantiated with a multi-level nested URI (so a URI with a path of something like `/path/file.rb` instead of `/path`) to break, since they would not properly remove the leading slash based on the regexp (previous version expected a trailing slash to some degree in `path`, so `/path/other/things` or `/path/`).
  - The capture group in the regexp, `([^\/]+)` already greedily grabs all of the non-slash characters it can, so there is no need to have this extra slash check following it, since we would delete it anyway.

Links
-----

* Previous PRs
  - Previous fixes to `MiqSwiftStorage`: https://github.com/ManageIQ/manageiq-gems-pending/pull/398
  - Adding `MiqSwiftStorage#download_single`: https://github.com/ManageIQ/manageiq-gems-pending/pull/400

### DIFF
--- a/lib/gems/pending/util/object_storage/miq_swift_storage.rb
+++ b/lib/gems/pending/util/object_storage/miq_swift_storage.rb
@@ -43,7 +43,8 @@ class MiqSwiftStorage < MiqObjectStorage
     # write dump file to swift
     #
     logger.debug("Writing [#{source_input}] to => Bucket [#{container_name}] using object file name [#{object_file}]")
-    begin
+
+    with_standard_swift_error_handling("uploading") do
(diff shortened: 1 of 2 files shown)


NickLaMuro:
@miq-bot assign @carbonin

PR #417, opened 2019-01-08
Remove miq-password and use ManageIQ::Password instead

carbonin:
This allows us to keep all the code in one place while not
also having to change all of the existing gems relying on
"require 'util/miq-password'" to still work.


carbonin:
Leaving this as WIP right now until we resolve https://github.com/ManageIQ/manageiq-password/pull/4 vs https://github.com/ManageIQ/manageiq-password/issues/7

Right now, the released version of manageiq-password uses `PasswordError`, but the master branch uses `MiqPasswordError`, but there is an issue open to continue using `PasswordError`

Using `PasswordError` would be a breaking change for all of the users of this gem and would require a bunch of PRs to those places before we merge this.

cc @agrare @Fryguy @bdunne

spec/support/custom_matchers/be_encrypted.rb
```diff
@@ -1,3 +1,4 @@
+
```
Fryguy: Bonus line?

carbonin:
Tested this branch in manageiq:

```ruby
11:12:32:~/Source/manageiq (master *)$ bin/rails c
** Using session_store: ActionDispatch::Session::MemCacheStore
** ManageIQ master, codename: Ivanchuk
MiqLoading development environment (Rails 5.0.7.1)
irb(main):001:0> MiqPassword
=> ManageIQ::Password
irb(main):002:0> MiqPassword::MiqPasswordError
=> ManageIQ::Password::PasswordError

```

simaishi:
Travis was failing for hammer branch due to manageiq-password gem change:

``` 
 1) MiqProvisionRequestTemplate exists after source template is deleted
     Failure/Error: val = MiqPassword.try_encrypt(val) unless val.blank?
     NoMethodError:
       undefined method `encrypt64' for nil:NilClass
```

As per discussion with @Fryguy (and approval from @dmetzger57), backported this PR to Hammer branch.

```
commit c36522d71ab4027a0949f4e27b3dfe65049c196c
Author: Brandon Dunne <brandondunne@hotmail.com>
Date:   Tue Jan 8 11:25:06 2019 -0500

    Merge pull request #417 from carbonin/remove_miq_password
    
    Remove miq-password and use ManageIQ::Password instead
    
    (cherry picked from commit fd6f30e82da5cd7d32ce87172529f9bdc28f8a13)
```

PR #419, opened 2019-01-25
[WIP] Extract Benchmark.realtime_store/block to more_core_extensions

Fryguy:
Depends on ManageIQ/more_core_extensions#65

@jrafanie @bdunne Please review


Fryguy:
This dependent PR was released in 3.7.0, but changed the way that exception logging occurs, so this PR is not valid

PR #422, opened 2019-02-19
Change pg_basebackup xlog-method option to wal-method

carbonin:
This option changed names in PG 10


carbonin:
@NickLaMuro I can't seem to get this spec to pass with PG 10. Can you take a look at the last commit and see if there's anything I missed updating?

carbonin:
Ah it seems they moved to systemd:
```
$ travis_setup_postgresql 10
Starting PostgreSQL v10
sudo systemctl start postgresql@10-main
```

May need some help working out what needs to change ...

NickLaMuro:
@carbonin let me know if you are done trying things.  I can dig around if you want and figure out what changed, but I don't want to step on your toes.

carbonin:
@NickLaMuro this seems to work. What do you think? I ended up just restoring directly to the `/var/ramfs/postgres/...` directory because I didn't see any copying behavior in my debugging session.

spec/ci_helper.rb
```diff
@@ -16,17 +16,14 @@ def self.start
 
     # Make sure directly in the postgresql.conf, the data_directory is set
     # (requirement for pg_wrapper I think...)
-    system("echo \"data_directory = '/var/ramfs/postgresql/9.5/main'\" >> #{PGCONFIGDIR}/postgresql.conf")
-
-    # Clear out the existing ramfs dir so it will be re-generated on boot
-    system("sudo rm -rf /var/ramfs/postgresql/9.5")
+    system("echo \"data_directory = '/var/ramfs/postgresql/10/main'\" >> #{PGCONFIGDIR}/postgresql.conf")
```
NickLaMuro: This is my fault, but I am wondering why I didn't use `PGDATADIR` here...

Want to change it and pretend that I did it correctly the first time?

carbonin: Well the first time around they weren't the same directory so definitely not your fault. I'll DRY it up.

NickLaMuro:
Cool, looks good!

I assume we are leaving the [WIP] for now as we get the rest of the PG10 stuff up to date, right?

* * *

Also, something I just thought of:  Does it make sense keeping a `9.5` dump around so that we can test the upgrade path of `9.5` to `10`?

carbonin:
> Also, something I just thought of: Does it make sense keeping a `9.5` dump around so that we can test the upgrade path of `9.5` to `10`?

Not unless we're going to implement an upgrade integration test in this repo, which to be clear, I don't think belongs here.

PR #426, opened 2019-03-20
Add put_file method

fdupont-redhat:
This pull request adds the ability to either create/override a file on the remote host, either from a file on local filesystem or from a variable. This will for example be used by V2V to create the throttling limits file on the conversion host.

Associated RHBZ: https://bugzilla.redhat.com/show_bug.cgi?id=1690851

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -50,7 +50,7 @@ class MiqSshUtil
   def put_file(to, content = nil, path = nil)
     raise "Need to provide either content or path" if content.nil? && path.nil?
     run_session do |ssh|
-      content ||= File.open(path, 'rb')(&:read)
+      content ||= File.open(path, 'rb') { |f| f.read }
       $log&.debug("MiqSshUtil::put_file - Copying file to #{@host}:#{to}.")
       ssh.sftp.file.open(to, 'wb') { |f| f.write(content) }
       $log&.debug("MiqSshUtil::get_file - Copying of file to #{@host}:#{to}, complete.")
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -40,9 +40,9 @@ class MiqSshUtil
 
   def get_file(from, to)
     run_session do |ssh|
-      $log.debug "MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}." if $log
+      $log&.debug("MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}.")
       data = ssh.sftp.download!(from, to)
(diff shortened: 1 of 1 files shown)


fdupont-redhat:
@miq-bot add_label transformation, enhancement, hammer/yes
@miq-bot add_reviewer @agrare 
@miq-bot add_reviewer @carbonin

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -40,13 +40,23 @@ def initialize(host, user, password, options = {})
...
-      $log.debug "MiqSshUtil::get_file - Copying of #{@host}:#{from} to #{to}, complete." if $log
+      $log&.debug("MiqSshUtil::get_file - Copying of #{@host}:#{from} to #{to}, complete.")
       return data
     end
   end
 
+  def put_file(to, content = nil, path = nil)
+    raise "Need to provide either content or path" if content.nil? && path.nil?
+    run_session do |ssh|
+      content ||= File.open(path, 'rb') { |f| f.read }
```
kbrock: How is this different from `IO.read(path)` ?

Just a little curious.

(sorry, drive by PR comment)

fdupont-redhat: Good question. It's a solution from stackoverflow... So, I'm ok with whatever is the preferred way.

fdupont-redhat: Reading further, it seems that `IO.binread` has been introduced in Ruby 1.9.1, and should be preferred. So, I updated my code. Thanks for pointing this out.

carbonin:
Removing the hammer label as the BZ is marked as an RFE and not labeled with a backport flag.

carbonin:
@simaishi @dmetzger57 are we backporting RFEs?

simaishi:
@carbonin We can backport RFEs if approved.

fdupont-redhat:
It's a requirement for IMS 1.2, which is targeted on 5.10.3, so backporting would be great. Otherwise, I'll have to implement it in the ConversionHost class...

carbonin:
Okay, I think that needs to be indicated on the BZ then, right?

fdupont-redhat:
For other IMS 1.2 BZs, we simply set the right Target Release on the BZ and hammer/yes label on the PR. Though, I'm ok to comply with whatever you prefer. What key/value should I put in the BZ ?

carbonin:
Okay, we should probably take this offline as it's a Red Hat specific process issue. I'll send an email.

simaishi:
Hammer backport details:

```
$ git log -1
commit 1834a57bbc5e2b16bd047e7f2d3d4846546e179d
Date:   Thu Mar 21 09:32:46 2019 -0400

    Merge pull request #426 from fdupont-redhat/add_put_file_method
    
    Add put_file method
    
    (cherry picked from commit 942c1243c081ff0b0165ae26b567126c1ff21e60)
    
    https://bugzilla.redhat.com/show_bug.cgi?id=1702085
```

PR #428, opened 2019-03-25
Add documentation to the MiqSshUtil class

djberg96:
As part of my understanding through documentation process, and in preparation for moving this bit of code into its own gem, I have added documentation to the `MiqSshUtil` class. This code was previously completely undocumented.

There are also a couple very minor code changes. One was to give each `attr_reader` its own line so that they can be documented individually. The other change was to `require 'tempfile'` at the top of the file. Frankly, we were getting lucky we didn't hit a `LoadError` up to this point because it's used within the `temp_cmd_file` method without being required first. So, some other code must have required it earlier.

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -306,7 +306,7 @@ class MiqSshUtil
   #
   # This method is functionally identical to the following code, except that it
   # yields itself (and nil) and re-raises certain Net::SSH exceptions as
-  # MiqException objects.
+  # ManageIQ exceptions.
   #
   #   MiqSshUtil.new(host, remote_user, remote_password, {:su_user => su_user, :su_password => su_password})
   #
@@ -320,6 +320,17 @@ class MiqSshUtil
     raise MiqException::MiqSshUtilHostKeyMismatch
   end
 
+  # Executes the provided +cmd+ using the exec or suexec method, depending on
+  # whether or not the :su_user option is set. The +doneStr+ and +stdin+
+  # arguments are passed along to the appropriate method as well.
+  #
+  # In the case of suexec, escape characters are automatically removed from
+  # the final output.
+  #
+  #--
+  # The _shell argument appears to be an artifact that has been retained
+  # over time for reasons that aren't immediately apparent.
+  #
   def shell_exec(cmd, doneStr = nil, _shell = nil, stdin = nil)
     return exec(cmd, doneStr, stdin) if @su_user.nil?
     ret = suexec(cmd, doneStr, stdin)
@@ -328,6 +339,12 @@ class MiqSshUtil
     ret
   end
 
+  # Copies the remote +file_path+ to a local temporary file, and then
+  # yields or returns a filehandle to the local temporary file.
+  #--
+  # Presumably this method was meant for use with the SCVMM provider
+  # given the hardcoded name of the temporary file.
+  #
   def fileOpen(file_path, perm = 'r')
     if block_given?
       Tempfile.open('miqscvmm') do |tf|
@@ -344,6 +361,8 @@ class MiqSshUtil
     end
   end
 
+  # Returns whether or not the remote +filename+ exists.
+  #
   def fileExists?(filename)
     shell_exec("test -f #{filename}") rescue return false
     true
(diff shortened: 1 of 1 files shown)


Fryguy:
@gmcculloug Please review

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -223,6 +288,18 @@ def temp_cmd_file(cmd)
     end
   end
 
+  # Shortcut method that creates and yields an MiqSshUtil object, with the +host+,
+  # +remote_user+ and +remote_password+ options passed in as the first three
+  # params to the constructor, while the +su_user+ and +su_password+ parameters are
+  # automatically set the :su_user and :su_password options. Remaining options are
```
Fryguy: automatically se**n**t?  There's a typo or grammatical error here.

djberg96:
Damn, need to rebase and fix.

lib/gems/pending/util/MiqSshUtil.rb
djberg96: I think the "are" is supposed to be removed. I reworded this sentence a couple times and it got a little goofed up.

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -47,6 +81,12 @@ def get_file(from, to)
     end
   end
 
+  # Upload the contents of local file +to+ to remote location +path+. You may
+  # use the specified +content+ instead of the content of the local file.
+  #
+  # At least one of the +content+ or +path+ parameters must be specified or
+  # and error is raised.
```
gmcculloug: *an error is raised.

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -57,6 +97,20 @@ def put_file(to, content = nil, path = nil)
...
 
+  # Execute the remote +cmd+ via ssh. This is automatically handled via
+  # channels on the ssh session so that various states can be checked,
+  # stored and logged independently and asynchronously.
+  #
+  # If the :passwordless_sudo option was set to true in the constructor
+  # then the +cmd+ will automatically be prepended with "sudo".
+  #
+  # If specified, the data collection will stop the first time a +doneStr+
+  # argument is encountered at the end of a line. If practice you would
```
gmcculloug: *In practice

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -233,6 +304,18 @@ def temp_cmd_file(cmd)
     end
   end
 
+  # Shortcut method that creates and yields an MiqSshUtil object, with the +host+,
+  # +remote_user+ and +remote_password+ options passed in as the first three
+  # params to the constructor, while the +su_user+ and +su_password+ parameters
+  # automatically set the corresponding :su_user and :su_password options. The
+  # remaining options are are passed normally.
```
gmcculloug: *remaining options are passed normally.

djberg96: Are are you sure? ;)

djberg96: Fixed.

lib/gems/pending/util/MiqSshUtil.rb
djberg96: Fixed.

lib/gems/pending/util/MiqSshUtil.rb
djberg96: Fixed.

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -38,6 +69,9 @@ def initialize(host, user, password, options = {})
     end
   end # def initialize
 
+  # Download the contents of the remote +from+ file to the local +to+ file. Some
+  # messages will be written to the global ManageIQ log in debug mode.
+  #
   def get_file(from, to)
     run_session do |ssh|
       $log&.debug("MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}.")
```
gmcculloug: I realize this is a documentation PR, but it would be nice to update the calls to `$log.debug` to be consistent.  Some use this safe-navigation approach but many still use a `if $log` check.

In additional, the calls can use a block so the string is not interpolated unless the logging level is active:
`$log&.debug { "MiqSshUtil::get_file - Copying file #{@host}:#{from} to #{to}." }`

Just some thoughts for a followup PR.

djberg96: Yeah, there's some oddities with the logging, and I'm not entirely sure why the log isn't set on the ssh object itself. That part was commented out for some reason, though I don't know why.

PR #431, opened 2019-04-29
Fix uninitialized warning in vmdb-logger.rb

djberg96:
Currently the core specs are flooded with this warning if rspec warnings are enabled:

`lib/gems/pending/util/vmdb-logger.rb:171: warning: global variable '$_miq_worker_current_msg' not initialized`

This global is set (eventually) by app/models/miq_queue_worker_base/runner.rb (where it is also uninitialized, but nevermind that for now):

https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_queue_worker_base/runner.rb#L101

### DIFF
--- a/spec/util/vmdb-logger_spec.rb
+++ b/spec/util/vmdb-logger_spec.rb
@@ -1,6 +1,6 @@
 require 'util/vmdb-logger'
 
-describe VMDBLogger do
+RSpec.describe VMDBLogger do
   describe "#log_hashes" do
     let(:buffer) { StringIO.new }
     let(:logger) { described_class.new(buffer) }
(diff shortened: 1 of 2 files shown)


Fryguy:
We are planning to move this miq_queue stuff out entirely.  @carbonin Thoughts here?

carbonin:
Yeah we are planning to extract this, but haven't had the cycles to finish it up. For now can we push this initialization to the `prefix_task_id` method rather than `initialize`? Considering that's the only method that uses the global and it's also the only one that's manageiq specific I think that would make sense for now.

spec/util/vmdb-logger_spec.rb
```diff
@@ -95,7 +95,7 @@
   end
 
   context "long messages" do
-    let(:logger) { VMDBLogger.new(@log) }
+    let(:logger) { VMDBLogger.new(File.expand_path(File.join(File.dirname(__FILE__), 'data/long_messages.log'))) }
```
carbonin: This is odd. Can we test an instance of the formatter directly since that's all this one spec seems to be doing? Then we can drop this `let` altogether, right?

carbonin: I think line 102 would become:
```ruby
 _, message = VMDBLogger::Formatter.new.call(:error, Time.now.utc, "", msg).split("-- : ")
```

djberg96: Done!

PR #435, opened 2019-05-30
Introduce FirmwareUpdateError

miha-plesko:
We're implementing a new process, the physical server firmware update, with state machine and everything. And it's best to raise a dedicated error when things go wrong.

@miq-bot add_label enhancement
@miq-bot assign @carbonin

### DIFF
--- a/lib/gems/pending/util/miq-exception.rb
+++ b/lib/gems/pending/util/miq-exception.rb
@@ -46,6 +46,9 @@ module MiqException
   # Exceptions during Provisioning
   class MiqProvisionError < Error; end
 
+  # Exceptions during Firmware Update
+  class MiqFirmwareUpdateError < Error; end
+
 
   class MaintenanceBundleInvalid < Error; end
   class MiqDeploymentError < Error; end


miha-plesko:
@carbonin or @agrare here we don't really need spec right?

carbonin:
Where is this being used?

agrare:
@carbonin e.g. here https://github.com/ManageIQ/manageiq-providers-redfish/pull/69/files#diff-eaea9756c62eee1b9caf9e487f1f7fe1R10

miha-plesko:
It's part of a series of the PRs related to Physical Server firmware updating, see the cover PR here https://github.com/ManageIQ/manageiq-providers-redfish/pull/69 (with links to all 8 PRs).

Generally we raise it in internal state machine for any error during firmware update process.

carbonin:
Okay, no problem. I just didn't know we collected all the exception here. I feel like this doesn't really belong in gems pending and the users in this repo should have their own exception types.... But obviously not something that could or should be addressed in this patch.

PR #437, opened 2019-06-24
MiqSshUtil updates

djberg96:
This PR fixes a bug in the `MiqSshUtil#exec` method, and makes a few tweaks to the constructor and exec method in general.

Fixes:

Error handling is currently semi-busted. The `on_extended_data` hook is missing a yield param. The net result is that there is never any data in the error output. In addition, the `status` check could break since it defaults to nil, with the result that it will try to call `.zero?` on a nil object, which will raise an exception. So, I changed the default from nil to 0.

Updates:

* Some unused code that was commented out was removed, i.e. logging.
* Some workaround code for older versions of net-ssh that is no longer relevant was removed.
* You can now configure the `use_agent` option. It still defaults to false for backwards compatibility.
* The error handling was updated slightly. It now no longer relies on activesupport methods, and checks the error buffer, too.

You can see these issues in practice with this simple script:

```
# Assuming you're in the manageiq-gems-pending repo
$:.unshift('lib/gems/pending/util')
require 'MiqSshUtil'
require 'etc'

# Adjust as desired
hostname = 'localhost'
username = Etc.getlogin
password = 'xxxxxxx'
command  = 'bogus'

pubkey = IO.read(File.join(Dir.home, '.ssh', 'id_rsa.pub'))

# You will have to edit MiqSshUtil.rb if you're on master and need 'use_agent'.
opts = {
  :key_data => pubkey,
  :use_agent => true
}

ssh = MiqSshUtil.new(hostname, username, password)

begin
  result = ssh.exec(command)
rescue => err
  puts "OOPS: #{err.message}"
else
  puts "RESULT: #{result}"
end
```

BEFORE:

`RESULT: ` <---- Blank! And not registered as an error!

AFTER:

`OOPS: MiqSshUtil::exec - Command 'bogus' failed: bash: bogus: command not found, status: 127`

Partially addresses https://bugzilla.redhat.com/show_bug.cgi?id=1719689

### DIFF
--- a/lib/gems/pending/util/MiqSshUtil.rb
+++ b/lib/gems/pending/util/MiqSshUtil.rb
@@ -46,10 +46,8 @@ class MiqSshUtil
       :remember_host   => false,
       :verbose         => :warn,
       :non_interactive => true,
+      :use_agent       => false
     }.merge(options)
-
-    # Seems like in 2.9.2, there needs to be blank :keys, when we are passing private key as string
-    @options[:keys] = [] if options[:key_data]
 
     # Pull the 'remember_host' key out of the hash because the SSH initializer will complain
     @remember_host     = @options.delete(:remember_host)
@@ -59,14 +57,6 @@ class MiqSshUtil
 
     # Obsolete, delete if passed in
     @options.delete(:authentication_prompt_delay)
-
-    # don't use the ssh-agent
-    @options[:use_agent] = false
-
-    # Set logging to use our default handle if it exists and one was not passed in
-    unless @options.key?(:logger)
-      #        @options[:logger] = $log if $log
-    end
   end # def initialize
 
   # Download the contents of the remote +from+ file to the local +to+ file. Some
@@ -128,7 +118,7 @@ class MiqSshUtil
           data.each_line { |l| return outBuf if doneStr == l.chomp } unless doneStr.nil?
         end
 
-        channel.on_extended_data do |_channel, data|
+        channel.on_extended_data do |_channel, _type, data|
           $log.debug "MiqSshUtil::exec - STDERR: #{data}" if $log
           errBuf << data
         end
@@ -149,10 +139,10 @@ class MiqSshUtil
 
         channel.on_close do |_channel|
           $log.debug "MiqSshUtil::exec - Command: #{cmd}, exit status: #{status}" if $log
-          unless signal.nil? || status.zero?
-            raise "MiqSshUtil::exec - Command #{cmd}, exited with signal #{signal}" unless signal.nil?
-            raise "MiqSshUtil::exec - Command #{cmd}, exited with status #{status}" if errBuf.empty?
-            raise "MiqSshUtil::exec - Command #{cmd} failed: #{errBuf}, status: #{status}"
+          if !signal.nil? || status.nonzero? || !errBuf.empty?
+            raise "MiqSshUtil::exec - Command '#{cmd}', exited with signal #{signal}" unless signal.nil?
+            raise "MiqSshUtil::exec - Command '#{cmd}', exited with status #{status}" if errBuf.empty?
+            raise "MiqSshUtil::exec - Command '#{cmd}' failed: #{errBuf.chomp}, status: #{status}"
           end
           return outBuf
         end
@@ -160,7 +150,7 @@ class MiqSshUtil
         $log.debug "MiqSshUtil::exec - Command: #{cmd} started." if $log
         channel.exec(cmd) do |chan, success|
           raise "MiqSshUtil::exec - Could not execute command #{cmd}" unless success
-          if stdin.present?
+          unless stdin.nil?
             chan.send_data(stdin)
             chan.eof!
           end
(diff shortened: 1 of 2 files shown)


djberg96:
@roliveri How's it look?

djberg96:
@miq-bot add_reviewer @bdunne

djberg96:
@miq-bot add_label bug, changelog/yes

djberg96:
@miq-bot add_label hammer/yes

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -149,18 +139,18 @@ def exec(cmd, doneStr = nil, stdin = nil)
 
         channel.on_close do |_channel|
           $log.debug "MiqSshUtil::exec - Command: #{cmd}, exit status: #{status}" if $log
-          unless signal.nil? || status.zero?
-            raise "MiqSshUtil::exec - Command #{cmd}, exited with signal #{signal}" unless signal.nil?
-            raise "MiqSshUtil::exec - Command #{cmd}, exited with status #{status}" if errBuf.empty?
-            raise "MiqSshUtil::exec - Command #{cmd} failed: #{errBuf}, status: #{status}"
+          if !signal.nil? || status.nonzero? || !errBuf.empty?
```
carbonin: I think the logic here is being changed.

Previously we would enter the block only `if signal != nil && status != 0`

Now we will enter the block if only `signal != nil`

I think the conditional you want here is:
```ruby
if !errBuf.empty? || (!signal.nil? && status.nonzero?)
```

Alternatively you could change a bit less and go with:
```ruby
if !(signal.nil? || status.zero?) || !errBuf.empty?
```

As a side note it seems like the content of the block is checking against the possibility that `signal` could still be nil even though it has to be non-nil in the original implementation. So if the conditional was wrong to begin with then this looks good. 

I'm just not sure which direction you were trying to go with this.

-----------------------------------------

FWIW this is why I don't like `unless` with more than one condition.

djberg96: @carbonin Not sure what you mean:

`if !signal.nil? || status.nonzero? || !errBuf.empty?`

We'll enter the block if any of those conditions is true.

Fryguy: @djberg96 You applied the unless wrong, I think...

Previously...

```ruby
unless signal.nil? || status.zero?
# means
if !(signal.nil? || status.zero?)
# =>
if !signal.nil? && !status.zero?
# =>
if !signal.nil? && status.nonzero?
```

However you have (ignoring the new extra thing)

```ruby
if !signal.nil? || status.nonzero?
```

lib/gems/pending/util/MiqSshUtil.rb
```diff
@@ -149,18 +139,18 @@ def exec(cmd, doneStr = nil, stdin = nil)
...
+            raise "MiqSshUtil::exec - Command '#{cmd}' failed: #{errBuf.chomp}, status: #{status}"
           end
           return outBuf
         end
 
         $log.debug "MiqSshUtil::exec - Command: #{cmd} started." if $log
         channel.exec(cmd) do |chan, success|
           raise "MiqSshUtil::exec - Could not execute command #{cmd}" unless success
-          if stdin.present?
+          unless stdin.nil?
```
Fryguy: Why change this?  The present? was clearer than the new double negative, IMO.

Fryguy: If you're dropping activesupport, then perhaps, just `if stdin`?

lib/gems/pending/util/MiqSshUtil.rb
djberg96: @Fryguy right, I'm saying the logic I have now is the logic I want. If any of those conditions is met then something went wrong.

lib/gems/pending/util/MiqSshUtil.rb
djberg96: Alright, I'll have to add an explicit require line for require 'active_support/core_ext/object/blank', though.

djberg96: @Fryguy Updated. :)

lib/gems/pending/util/MiqSshUtil.rb
carbonin: @djberg96 I was just pointing out that the logic changed. You didn't call out this change in the description or commit message so I didn't know if it was intentional.

I know specs for this are probably tough to write, but having those would clarify intent quite nicely.

djberg96:
@carbonin Alright, I've added some specs that should demonstrate that it behaves as expected. I didn't want to go down this road originally because I knew it would mean making a couple other refactorings.

Namely, the `Channel#exec` block had to be moved up (so that channel failure could be checked before command failure), and I had to strip the signal check result because it can (apparently) return a blank string of more than 1 character.

simaishi:
Hammer backport details:

```
$ git log -1
commit 7300bf53e17784a7c12dc1133239bc8b76a61f8a
Author: Jason Frey <jfrey@redhat.com>
Date:   Tue Jul 2 16:16:08 2019 -0400

    Merge pull request #437 from djberg96/miq_ssh_util_updates
    
    MiqSshUtil updates
    
    (cherry picked from commit 02c9fb67797331640e2840682d8234f2f4a5415c)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1726439
```

PR #446, opened 2019-09-05
[IVANCHUK] Run Travis for ruby 2.5.x only

simaishi:
Not sure why this repo got missed in the last blaster...  cc @d-m-u


d-m-u:
Oh, cause ```if `grep -E 'git clone.+manageiq.git' bin/setup`.include?('manageiq.git')``` won't work here, sorry, yeah.

PR #447, opened 2019-09-06
[MiqGenericMountSession] Fix .source_for_log

NickLaMuro:
Does better fallback handling and nil checking for this method.

This can fail as part of `MiqGenericMountSession#add`'s `rescue` statement when the `@source_input` is not defined, and that will mask the actual error that was raised.

This fixes that bug so the proper error is shown in the logs.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1763890

### DIFF
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -490,6 +490,12 @@ class MiqGenericMountSession < MiqFileStorage::Interface
   end
 
   def source_for_log
-    @input_writer ? "<STREAMED_FROM_CMD>" : @source_input.path
+    if @input_writer
+      "<STREAMED_FROM_CMD>"
+    elsif @source_input
+      @source_input.path
+    else
+      "<NOT_READY>"
+    end
   end
 end


lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -490,6 +490,10 @@ def settings_mount_point
   end
 
   def source_for_log
-    @input_writer ? "<STREAMED_FROM_CMD>" : @source_input.path
+    case
```
NickLaMuro: I think this looks better than an `if` would, but clearly `rubocop` disagrees.  Up to the reviewer if they want me to change it, but not going to go out of my way to do so.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -490,6 +490,10 @@ def settings_mount_point
...
+    when @input_writer then "<STREAMED_FROM_CMD>"
+    when @source_input then @source_input.path
+    else "<NOT_READY>"
```
djberg96: At the very minimum I would argue that we should be using the instance methods rather than referencing instance variables that don't even exist in this file.

I also feel like there should be an inheritable method that does this instead of defining it in a subclass.

I also prefer `'if` here, as case without an argument just looks odd to me.

NickLaMuro: > At the very minimum I would argue that we should be using the instance methods rather than referencing instance variables that don't even exist in this file.

Fair enough, I think that is a reasonable request.  Granted, these were meant to be private variables, which is why I didn't create methods for them but I think your argument makes sense.

> I also prefer 'if here...

Fine, you twisted my arm... I will do it...

😄

NickLaMuro: @djberg96 So I changed the if, but didn't do the other request.

It isn't that I don't disagree with it, but I want to mostly fix the bug in this PR, and address the major refactoring that this code desperately needs at another time.  This code I know will work, where as I might be introducing other bugs if I try and fix the inheritance and instance variable usage further.

Hope that makes sense.

NickLaMuro:
@miq-bot add_label hammer/yes ivanchuk/yes

NickLaMuro:
F...

@miq-bot add_label hammer/yes, ivanchuk/yes

simaishi:
Ivanchuk backport details:

```
$ git log -1
commit d09786dc6191764fb0a38caf5ea2210d6cfddca6
Date:   Wed Oct 23 11:51:28 2019 -0400

    Merge pull request #447 from NickLaMuro/fix_source_for_log_errors
    
    [MiqGenericMountSession] Fix .source_for_log
    
    (cherry picked from commit bb0b1d483ea9a27a919b829c11ef0421ac6630ba)
    
    Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1767895
```

PR #449, opened 2019-10-17
update aws-sdk gem to v3

AlexanderZagaynov:
This is a part of global update `aws-sdk` gem to `v3`

Together with:
- https://github.com/ManageIQ/manageiq-providers-amazon/pull/488
- https://github.com/ManageIQ/manageiq/pull/19411
- https://github.com/ManageIQ/amazon_ssa_support/pull/44

It should fix https://github.com/ManageIQ/manageiq-providers-amazon/issues/465


lib/gems/pending/util/object_storage/miq_s3_storage.rb
```diff
@@ -84,8 +84,8 @@ def bucket
   private
 
   def s3
-    require 'aws-sdk'
-    require 'util/extensions/aws-sdk/s3_upload_stream_patch'
+    require 'aws-sdk-s3'
```
djberg96: Do you want to use `require "aws-sdk-#{service.to_s.downcase}"` here as well? Or is that overkill?

AlexanderZagaynov: where I can take `service` here? :)

djberg96: True enough!

djberg96:
Huh, 2.5 passed, but 2.4 failed. Maybe restart Travis on this one, see what happens.

AlexanderZagaynov:
yeah, some timeout issue, however I have no control to restart it...

djberg96:
@AlexanderZagaynov Neither can I. You can do the old "close and reopen" trick, I think that will kick it.

djberg96:
@NickLaMuro What specifically is at stake? Anything we can help with?

agrare:
Provider tests are green with this applied

NickLaMuro:
@djberg96 I think it is fine that this is merged without me testing (sorry... computer swapping all day), but this is your question:

> @NickLaMuro What specifically is at stake? Anything we can help with?

It is currently being used for here:

https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/object_storage/miq_s3_storage.rb

Which, as you probably can imagine, has zero unit tests around the database backup code for this.  Also, we are streaming the data from the backup directly to S3, which makes it complicated.  That said, what was here previously was a patch of a direct port of what existed in v3 that should work just fine.

That said, I haven't had a chance to run through my test suite of this yet, since I honestly have missed placed where I have configured it, but the test suite is basically here:


https://gist.github.com/NickLaMuro/8438015363d90a2d2456be650a2b9bc6

That said, getting that repo ("test suite") up and running is no easy task, so I just need to find some time to update things and make sure this new code works with the s3 tests still.  On my TODO list...

PR #458, opened 2020-01-25
Update sys-proctable gem to 1.2.x

djberg96:
Update the sys-proctable gem to 1.2.x, which eliminates the platform-specific gems, and thus avoids the bundler issues that plagued it.

It also includes some efficiency updates for Mac from our very own @NickLaMuro. :)

The 1.2.x release switched to use keyword arguments, so a couple minor updates are needed to go along with it.

Related PR: https://github.com/ManageIQ/manageiq/pull/19767

Cross repo tests: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/64

### DIFF
--- a/lib/gems/pending/util/miq-process.rb
+++ b/lib/gems/pending/util/miq-process.rb
@@ -111,7 +111,7 @@ class MiqProcess
       percent_cpu                    = (1.0 * result[:cpu_time]) / cpu_total
       result[:percent_cpu]           = round_to(percent_cpu * 100.0, 2)
 
-      smaps = Sys::ProcTable.ps(pid).smaps
+      smaps = Sys::ProcTable.ps(:pid => pid).smaps
       result[:proportional_set_size] = smaps.pss
       result[:unique_set_size]       = smaps.uss
     when :macosx
@@ -130,7 +130,7 @@ class MiqProcess
   def self.command_line(pid)
     # Already exited pids, or permission errors cause ps or ps.cmdline to be nil,
     # so the best we can do is return an empty string.
-    Sys::ProcTable.ps(pid).try(:cmdline) || ""
+    Sys::ProcTable.ps(:pid => pid).try(:cmdline) || ""
   end
 
   def self.alive?(pid)
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -38,7 +38,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "net-ssh",                 "~> 4.2.0"
   s.add_runtime_dependency "nokogiri",                "~> 1.8.1"
   s.add_runtime_dependency "rake",                    ">= 11.0"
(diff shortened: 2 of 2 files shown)


djberg96:
@miq-bot add_reviewer @Fryguy, @NickLaMuro

Fryguy:
Do these also need to be changed in core?

- https://github.com/ManageIQ/manageiq/blob/08891f6da51707e08b72b1fa88ac60f4e8a3f4e0/app/models/miq_cockpit_ws_worker.rb#L32
- https://github.com/ManageIQ/manageiq/blob/08891f6da51707e08b72b1fa88ac60f4e8a3f4e0/app/models/miq_cockpit_ws_worker/runner.rb#L6
- https://github.com/ManageIQ/manageiq/blob/08891f6da51707e08b72b1fa88ac60f4e8a3f4e0/app/models/miq_cockpit_ws_worker/runner.rb#L129

Fryguy:
On that note, can you run cross repo tests?

djberg96:
@Fryguy only the last one, since it takes a pid. Without arguments, there's no change.

lib/gems/pending/util/miq-process.rb
```diff
@@ -111,7 +111,7 @@ def self.processInfo(pid = nil)
       percent_cpu                    = (1.0 * result[:cpu_time]) / cpu_total
       result[:percent_cpu]           = round_to(percent_cpu * 100.0, 2)
 
-      smaps = Sys::ProcTable.ps(pid).smaps
+      smaps = Sys::ProcTable.ps(:pid => pid).smaps
```
NickLaMuro: @djberg96 So I was thinking about this when you made this PR, but what are your thoughts about replacing all of the functionality found here (in `MiqProcess`) with current (or new) functionality in `Sys::ProcTable`?

There might be a few changes we want to consider to do that, since we probably don't need to load *everything* that `.ps` loads for each of the methods in this file, but I think it just makes sense as another thing we can remove from this repo.

djberg96: Looking at `MiqProcess`, it seems *most* of it could be replaced, but I would have to do some more research.

NickLaMuro: Yeah, I guess it isn't a perfect 1-to-1 port, but it would be nice to remove a lot of it if nothing else.

jrafanie:
Agreed, this seems reasonable but would like cross-repo results.  That way we can line up all of the needed PRs and get them merged the moment everything is ready to go.

djberg96:
@jrafanie https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/64

djberg96:
@jrafanie, @Fryguy, @NickLaMuro cross-repo tests are green.

NickLaMuro:
Sorry for my delay, but added a comment to the cross repo PR.

That said, I don't expect much of a change since it seems like we aren't failing currently.  😩

djberg96:
@NickLaMuro Updated as per your suggestion, still green.

Issue #459, opened 2020-01-27
Can MiqProcess be dropped completely in favor of sys-proctable?


djberg96:
@NickLaMuro asked the question over at https://github.com/ManageIQ/manageiq-gems-pending/pull/458#discussion_r371317219 as to whether or not we can drop `MiqProcess` completely in favor of sys-proctable.

I've created this issue to see if I can do a method by method comparison of how it would be done. I'm pretty sure most of it can.

djberg96:
`MiqProcess.get_active_process_by_name(process_name)` (source: `https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-process.rb#L9`)

The sys-proctable equivalent:

`Sys::ProcTable.ps.find_all{ |process| process.name == name }`

Comments: The Windows code looks broken to me. It's collecting handles for some reason, which are not pids, but are the Windows equivalent of a file descriptor.

djberg96:
`MiqProcess.linux_process_stat(pid = nil)`

Source: https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-process.rb#L24

The sys-proctable gem already collects information out of `/proc/<pid>/stat`, as well as several other files, so all of those fields are available as methods on the resulting proctable struct, e.g.` process.pid`, `process.ppid`, and so on. So, nothing special needs to be done here.

djberg96:
`MiqProcess.processInfo(pid = nil)`

Source: https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-process.rb#L84

All of the information that method collects is available within the data returned by sys-proctable. There are some calculations happening here, so those would have to be emulated.

djberg96:
`MiqProcess.command_line(pid)`

Source: https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-process.rb#L130

The sys-proctable equivalent:

`Sys::ProcTable.ps(pid).try(:cmdline) || ""`

Comments: The underlying implementation is the equivalent code I pasted above.

djberg96:
`MiqProcess.alive?(pid)`

Source: https://github.com/ManageIQ/manageiq-gems-pending/blob/master/lib/gems/pending/util/miq-process.rb#L136

You don't need sys-proctable for his, the current implementation is sufficient.

[...]

PR #461, opened 2020-01-29
Refactor MiqProcess

djberg96:
Some initial refactorings based on https://github.com/ManageIQ/manageiq-gems-pending/issues/459. Specifically:

* Simplified the `get_active_process_by_name` method to a ~~one~~ two liner using the sys-proctable gem.
~~* Updated the `alive?` method, the code is already cross-platform.~~
~~* Updated the `suspend_process` and `resume_process` methods. Removed the dead Windows code, added an implementation for unixy platforms.~~
* Remove the `round_to` method, just use `Float#round`.

I also added and updated some comments.

Update: The `pause`, `resume` and `alive?` methods were moved into the `Process` module in the `more_core_extensions` library.

### DIFF
--- a/lib/gems/pending/util/miq-process.rb
+++ b/lib/gems/pending/util/miq-process.rb
@@ -130,15 +130,6 @@ class MiqProcess
     Sys::ProcTable.ps(:pid => pid).try(:cmdline) || ""
   end
 
-  # Return whether or not the given +pid+ is running.
-  #
-  def self.alive?(pid)
-    Process.kill(0, pid)
-    true
-  rescue Errno::ESRCH
-    false
-  end
-
   def self.is_worker?(pid)
     command_line = self.command_line(pid)
     command_line.include?(MiqWorker::PROCESS_TITLE_PREFIX)
@@ -230,29 +221,5 @@ class MiqProcess
     # Convert format 00:00:00 to seconds
     t = time_str.split(':')
     (t[0].to_i * 3600) + (t[1].to_i * 60) + t[2].to_i
-  end
-
-  # Suspend the process +pid+ using a SIGSTOP. If the process isn't running
-  # then this is no-op.
-  #
-  def self.suspend_process(pid)
-    case Sys::Platform::OS
-    when :unix
-      Process.kill('STOP', pid) if alive?(pid)
-    else
-      raise "Method MiqProcess.suspend_process not implemented on this platform [#{Sys::Platform::IMPL}]"
-    end
-  end
-
-  # Resume the process +pid+ using a SIGCONT. If the process isn't running
-  # then this is a no-op.
-  #
-  def self.resume_process(pid)
-    case Sys::Platform::OS
-    when :unix
-      Process.kill('CONT', pid) if alive?(pid)
-    else
-      raise "Method MiqProcess.resume_process not implemented on this platform [#{Sys::Platform::IMPL}]"
-    end
   end
 end
--- a/lib/gems/pending/util/miq-process.rb
+++ b/lib/gems/pending/util/miq-process.rb
@@ -232,23 +232,25 @@ class MiqProcess
     (t[0].to_i * 3600) + (t[1].to_i * 60) + t[2].to_i
   end
 
-  # Suspend the process +pid+ using a SIGSTOP.
+  # Suspend the process +pid+ using a SIGSTOP. If the process isn't running
+  # then this is no-op.
   #
   def self.suspend_process(pid)
     case Sys::Platform::OS
     when :unix
-      Process.kill('STOP', pid)
+      Process.kill('STOP', pid) if alive?(pid)
     else
       raise "Method MiqProcess.suspend_process not implemented on this platform [#{Sys::Platform::IMPL}]"
     end
   end
 
-  # Resume the process +pid+ using a SIGCONT.
+  # Resume the process +pid+ using a SIGCONT. If the process isn't running
+  # then this is a no-op.
   #
   def self.resume_process(pid)
     case Sys::Platform::OS
     when :unix
-      Process.kill('CONT', pid)
+      Process.kill('CONT', pid) if alive?(pid)
     else
       raise "Method MiqProcess.resume_process not implemented on this platform [#{Sys::Platform::IMPL}]"
     end
--- a/lib/gems/pending/util/miq-process.rb
+++ b/lib/gems/pending/util/miq-process.rb
@@ -119,21 +119,23 @@ class MiqProcess
     result
   end
 
+  # Return the command line string for the given +pid+. If the pid has already
+  # exited, or there is some sort of permissions issue that causes it to be set
+  # to nil, then return an empty string instead.
+  #
   def self.command_line(pid)
     # Already exited pids, or permission errors cause ps or ps.cmdline to be nil,
     # so the best we can do is return an empty string.
     Sys::ProcTable.ps(:pid => pid).try(:cmdline) || ""
   end
 
+  # Return whether or not the given +pid+ is running.
+  #
   def self.alive?(pid)
-    raise NotImplementedError, "Method MiqProcess.alive? not implemented on this platform [#{Sys::Platform::IMPL}]" unless Sys::Platform::OS == :unix
-
-    begin
-      Process.kill(0, pid)
-      true
-    rescue Errno::ESRCH
-      false
-    end
+    Process.kill(0, pid)
+    true
+  rescue Errno::ESRCH
+    false
   end
 
   def self.is_worker?(pid)
--- a/lib/gems/pending/util/miq-process.rb
+++ b/lib/gems/pending/util/miq-process.rb
@@ -6,19 +6,11 @@ require 'util/runcmd'
 require 'util/miq-system'
 
 class MiqProcess
+  # Collect and return a list of PID's for all processes that match
+  # +process_name+ or +process_name.exe+.
+  #
   def self.get_active_process_by_name(process_name)
-    pids = []
-
-    case Sys::Platform::IMPL
-    when :mswin, :mingw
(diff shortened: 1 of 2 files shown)


djberg96:
@miq-bot add_reviewer @NickLaMuro

lib/gems/pending/util/miq-process.rb
```diff
@@ -258,10 +256,3 @@ def self.round_to(number, precision)
     (number * mult).round.to_f / mult
   end
 end
-
-# Examples:
```
NickLaMuro: Do you think these might have a purpose at the top of the file?  Or maybe as documentation for usage of each method?

lib/gems/pending/util/miq-process.rb
```diff
@@ -237,17 +229,23 @@ def self.str_time_to_sec(time_str)
...
     end
   end
 
+  # Resume the process +pid+ using a SIGCONT.
+  #
   def self.resume_process(pid)
     case Sys::Platform::OS
-    when :windows then Process.process_thread_list[pid].each { |tid| Process.suspend_resume_thread(tid, true) }
+    when :unix
+      Process.kill('CONT', pid)
```
NickLaMuro: Yeah, this seems much simpler.

However, does there maybe need to be a rescue here as a result?  I think with the `.each`, it would only find the pids if they existed, and no-op'd if they didn't.  That might be what was desired and wanted here.

lib/gems/pending/util/miq-process.rb
```diff
@@ -6,19 +6,11 @@
...
-    when :mswin, :mingw
-      require 'util/win32/miq-wmi'
-      WMIHelper.connectServer.run_query("select Handle,Name from Win32_Process where Name = '#{process_name}.exe'") { |p| pids << p.Handle.to_i }
-    when :linux, :macosx
-      pids = `ps -e | grep #{process_name} | grep -v grep `.split("\n").collect(&:to_i)
-    else
-      raise "Method MiqProcess.get_active_process_by_name not implemented on this platform [#{Sys::Platform::IMPL}]"
-    end
-    pids
+    Sys::ProcTable.ps.select{ |process| [process_name, "#{process_name}.exe"].include?(process.name) }.map(&:pid)
```
NickLaMuro: I would be curious to know if `.ps` is now slower than `ps -e | grep ... | grep -v ...`, since I know under the hood there is a lot of file reading required for `Sys::ProcTable.ps` when no pid file is required.  Especially since most of the read data wouldn't be used (mem info, CPU usage, etc.).

Might be interesting to determine if lazy loading would be possible for `.ps` for some of the extra data there, but that is just the performance side of my brain being triggered.

djberg96:
Hm, what did I break?

```
1) Benchmark Timeout raising within .realtime_block
     Failure/Error: raise("Completed without Timeout!?  Tip: may result from unclosed Tempfile object.")
     
     RuntimeError:
       Completed without Timeout!?  Tip: may result from unclosed Tempfile object.
     # ./spec/util/extensions/miq-benchmark_spec.rb:124:in `block (3 levels) in <top (required)>'
     # ./spec/util/extensions/miq-benchmark_spec.rb:112:in `times'
     # ./spec/util/extensions/miq-benchmark_spec.rb:112:in `block (2 levels) in <top (required)>'
```

NickLaMuro:
> Hm, what did I break?

@djberg96 sorry, just submitted my review just after your comment.  I think that spec fails every so often, so you probably can re-trigger the build and it most likely will go away.

lib/gems/pending/util/miq-process.rb
djberg96: Ah, true, we can disable at least some of the reads, e.g. `:smaps => false`.

lib/gems/pending/util/miq-process.rb
djberg96: They didn't look very useful to me.

lib/gems/pending/util/miq-process.rb
djberg96: Ok, updated.

lib/gems/pending/util/miq-process.rb
djberg96: Updated.

djberg96:
@miq-bot add_reviewer @carbonin

djberg96:
@miq-bot remove_label unmergeable

djberg96:
Should be mergeable after the latest push, but the label stuck around. Attempting to remove it....

djberg96:
@NickLaMuro You ok with this now? @carbonin What do you think?

lib/gems/pending/util/miq-process.rb
```diff
@@ -127,21 +120,23 @@ def self.processInfo(pid = nil)
...
   def self.alive?(pid)
-    raise NotImplementedError, "Method MiqProcess.alive? not implemented on this platform [#{Sys::Platform::IMPL}]" unless Sys::Platform::OS == :unix
-
-    begin
-      Process.kill(0, pid)
-      true
-    rescue Errno::ESRCH
-      false
-    end
+    Process.kill(0, pid)
```
bdunne: Any reason not to use https://github.com/ManageIQ/more_core_extensions/pull/73/files#diff-b07041a64a859612407de7d132fe52bdR47 ?

lib/gems/pending/util/miq-process.rb
```diff
@@ -6,19 +6,12 @@
...
-    when :mswin, :mingw
-      require 'util/win32/miq-wmi'
-      WMIHelper.connectServer.run_query("select Handle,Name from Win32_Process where Name = '#{process_name}.exe'") { |p| pids << p.Handle.to_i }
-    when :linux, :macosx
-      pids = `ps -e | grep #{process_name} | grep -v grep `.split("\n").collect(&:to_i)
-    else
-      raise "Method MiqProcess.get_active_process_by_name not implemented on this platform [#{Sys::Platform::IMPL}]"
-    end
-    pids
+    procs = Sys::ProcTable.ps(:smaps => false, :cgroup => false)
```
NickLaMuro: I forgot that you did add the kwargs stuff, so this is great.

Looking at the `sys-proctable` code, does seem like there are a few extra files being read and "Dir's glob'd" that aren't needed for this use case, but that is something to improved on in the future.

lib/gems/pending/util/miq-process.rb
```diff
@@ -237,31 +232,27 @@ def self.str_time_to_sec(time_str)
     (t[0].to_i * 3600) + (t[1].to_i * 60) + t[2].to_i
   end
 
+  # Suspend the process +pid+ using a SIGSTOP. If the process isn't running
+  # then this is no-op.
+  #
   def self.suspend_process(pid)
```
NickLaMuro: Wasn't this something you were going to handle with `more-core-extensions`, or are you waiting to do that later?

lib/gems/pending/util/miq-process.rb
djberg96: @bdunne You want me to just yank it completely? I couldn't find anything using it currently. The miq_web_server_worker_mixin.rb was using it, but it seems the relevant code got deleted last year.

@NickLaMuro Am I missing any calls elsewhere?

Same for pause and resume probably, since the current code was windows-only and was broken.

lib/gems/pending/util/miq-process.rb
```diff
@@ -6,19 +6,12 @@
...
-      require 'util/win32/miq-wmi'
-      WMIHelper.connectServer.run_query("select Handle,Name from Win32_Process where Name = '#{process_name}.exe'") { |p| pids << p.Handle.to_i }
-    when :linux, :macosx
-      pids = `ps -e | grep #{process_name} | grep -v grep `.split("\n").collect(&:to_i)
-    else
-      raise "Method MiqProcess.get_active_process_by_name not implemented on this platform [#{Sys::Platform::IMPL}]"
-    end
-    pids
+    procs = Sys::ProcTable.ps(:smaps => false, :cgroup => false)
+    procs.select { |process| [process_name, "#{process_name}.exe"].include?(process.name) }.map(&:pid)
```
carbonin: Can we add some specs for this? It's hard to see if this behavior is the same before and after this change without them.

lib/gems/pending/util/miq-process.rb
carbonin: Those sound like two very different questions (use a gem or remove). I think @bdunne is suggesting an alias for short term compatibility rather than copying the method contents.

lib/gems/pending/util/miq-process.rb
djberg96: @NickLaMuro removed, since it's now in more_core_extensions.

lib/gems/pending/util/miq-process.rb
djberg96: @carbonin pretty sure there's no way the current code is being used. It's only written for Windows, and that code appears to be invalid.

@bdunne, @carbonin, @NickLaMuro I've removed it. Anything that needs it can use `more_core_extensions`.

lib/gems/pending/util/miq-process.rb
djberg96: @NickLaMuro Updated, removed.

djberg96:
> Seems like we need to have a gem update as part of this PR, unless I am missing something.

I just needed to rebase I think. We updated the gem to 1.2.2 a few weeks ago, so all good.

lib/gems/pending/util/miq-process.rb
djberg96: @carbonin Ok, added some specs.

lib/gems/pending/util/miq-process.rb
djberg96: @NickLaMuro Yeah, it's a bit less efficient, but not terribly so with the options being passed.

PR #474, opened 2020-05-18
maintain consistent versioning of fog openstack across repos

d-m-u:
I feel like we should be consistent as well here: 

```
 manageiq-providers-openstack was resolved to 0.1.0, which depends on
      fog-openstack (>= 0.3.10)
```


d-m-u:
@miq-bot add_label dependencies

Fryguy:
If fog follows semver, I think it's better to leave it as 0.3.  @agrare Thoughts?

agrare:
:+1: fog-openstack is good about versioning (fog-google not so much)  `~> 0.3` is good with me

PR #475, opened 2020-05-28
Use Etc for num_cpus

djberg96:
For the `MiqSystem#num_cpus` method, just use the` Etc.nprocessors` method since:

* It's in the stdlib (and this method has been available since Ruby 2.2)
* It's cross-platform
* It's less code

### DIFF
--- a/lib/gems/pending/util/miq-system.rb
+++ b/lib/gems/pending/util/miq-system.rb
@@ -45,9 +45,12 @@ class MiqSystem
     nil
   end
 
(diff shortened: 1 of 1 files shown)


Fryguy:
@bdunne Please review.  I'm not sure who the callers are and so I'm not sure if there are any issues with this change or not

I'm curious what this will do 
- when we eventually have non-root on the appliance
- in a container

Note that the current code may have the same limitations...these are just general observations.

PR #477, opened 2020-06-15
Remove circular require from xml_hash.rb

djberg96:
Currently there's a circular require caused by xml_hash.rb. The circle is:

* xml_utils.rb -> `require 'util/miq-xml'`
* miq-xml.rb -> `require 'util/xml/miq_rexml'`
* miq_rexml.rb -> `require 'util/xml/xml_hash'`
* xml_hash.rb -> `require 'util/xml/xml_utils'`

The xml_utils.rb file provides 4 classes: `XmlFind`, `XmlHelpers`, `Xml2tags`, and `Xml2Array`. The xml_hash.rb file is only using one method from one of these classes: `XmlHelpers.stringify_keys`. Since activesupport has an extension that does this, I've replaced that code with the activesupport code (and activesupport is already a dependency).

The circular dependency manifests itself in the manageiq-smartstate specs, which I've posted about at https://github.com/ManageIQ/manageiq-smartstate/issues/131.

### DIFF
--- a/lib/gems/pending/util/xml/xml_hash.rb
+++ b/lib/gems/pending/util/xml/xml_hash.rb
@@ -1,6 +1,5 @@
 require 'enumerator'
 require 'rexml/document'
-require 'util/xml/xml_utils'
 
 module XmlHash
   class Element < Hash
--- a/lib/gems/pending/util/xml/xml_hash.rb
+++ b/lib/gems/pending/util/xml/xml_hash.rb
@@ -1,5 +1,6 @@
 require 'enumerator'
 require 'rexml/document'
+require 'active_support/core_ext/hash/keys'
 
 module XmlHash
   class Element < Hash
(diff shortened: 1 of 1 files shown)


Fryguy:
Here's the definition of XmlHelpers.stringify_keys

https://github.com/ManageIQ/manageiq-gems-pending/blob/d5a92a93e24d823a64aebcfbfa75741981a37bc6/lib/gems/pending/util/xml/xml_utils.rb#L81-L84

So, this PR LGTM :+1:

Fryguy:
@djberg96 Can we remove `XmlHelpers.stringify_keys` after this (i.e. are all callers removed?)

djberg96:
@Fryguy manageiq-smartstate uses it:

https://github.com/ManageIQ/manageiq-smartstate/blob/master/lib/metadata/util/win32/Win32Services.rb#L72-L80

Issue #231, opened 2017-07-10
gems/pending/util general cleanup


Fryguy:
### Delete

- [x] diag/miqping.rb - MiqWsPing
- [x] MiqDumpObj.rb
- [x] MiqVmFsUtil.rb

### Dedicated gem/process/application

- [ ] duplicate_blocker/* + duplicate_blocker.rb

### Move

- [ ] miq_apache/*  -> ManageIQ (for evmserver purposes)
- [ ] SyncDebug -> vmware_web_service
- [ ] miq-metakit -> smartstate
- [ ] miq-syntax-checker -> automation_engine / automation_models (MiqAeMethod)
- [ ] miq-unicode -> smartstate
- [ ] miq_tempfile -> smartstate
- [ ] system/evm_watchdog.rb -> manageiq-appliance

### Other

- [ ] miq-encode.rb
  - [ ] base24Decode -> smartstate
  - [x] base64* -> delete
  - [ ] encode/decode -> Keep in core

### Keep in core

- [ ] mount/*
- [ ] xml/* (core - to delete)
- [ ] win32/* (delete or breakdown further - e.g. wimparser to gem)
  - MiqWinrm
- [ ] MiqSockUtil
- [ ] miq-exception.rb
- [ ] miq-hash_struct.rb
- [ ] miq-ipmi (core - to delete)
- [ ] miq-logger (core - to delete)
- [ ] miq-password
- [ ] miq-process
- [ ] miq-system
- [ ] miq-uuid (can we get this into vmware speciicaly, even though hardware class uses it?)
- [ ] miq-logger-processor
- [ ] postgres_admin
- [ ] require_with_logging
- [ ] runcmd (core - to delete)
- [ ] vmdb-logger
- [ ] MiqSshUtil (possibly delete depending on the following)
  - Host fleecing (is that even a thing, since the code says "not esxi"?)
  - Automate can ssh exec a script on a host
  - Host verify ssh creds (for what purpose, though?)

### core extensions

- move to more_core_extensions
  - [ ] miq-deep#deep_clone (maybe?)
  - [ ] miq-hash
  - [ ] miq-module
  - [ ] miq-range
  - [ ] miq-symbol
  - [ ] to_i_with_method
  - [ ] miq-yaml (monkeypatch)
  - [ ] miq-stats (extension to Math module)
- other
  - [ ] to_miq_a (can we remove this or switch to Array.wrap)
  - [ ] benchmark -> dedicated gem
[...]

PR #478, opened 2020-06-15
Remove miq-kernel.rb / require_relative

djberg96:
The original issue mentioned in http://bugs.ruby-lang.org/issues/4487 appears to be fixed:

```
>cat eval_me1.rb 
eval(File.read('eval_me2.rb'), binding, File.expand_path('./eval_me2.rb'))

>cat eval_me2.rb 
require_relative 'eval_me1.rb'

>ruby eval_me2.rb # nothing
>ruby eval_me1.rb # nothing
```

I also tried loading them in irb:

```
irb(main):001:0> load "eval_me1.rb"
=> true
irb(main):002:0> load "eval_me2.rb"
=> true
```

I tried with both 2.5.7 and 2.6.5.

Part of the cleanup from https://github.com/ManageIQ/manageiq-gems-pending/issues/231

### DIFF
--- a/lib/gems/pending/util/extensions/miq-kernel.rb
+++ b/lib/gems/pending/util/extensions/miq-kernel.rb
@@ -1,7 +1,0 @@
-module Kernel
-  # Replaces Kernel's require_relative to allow it to be used in irb and eval
-  # See: http://bugs.ruby-lang.org/issues/4487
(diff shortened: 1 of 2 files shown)


Fryguy:
I also verified locally with irb and eval directly in a clean irb and it all seems to work now.

PR #479, opened 2020-06-16
Remove circular from miq_rexml.rb

djberg96:
Followup to https://github.com/ManageIQ/manageiq-gems-pending/pull/477, this time the circle is:

* xml/xml_utils requires `miq-xml`
* miq-xml requires `xml/miq_rexml`
* xml/miq_rexml requires `xml/xml_utils` (see first bullet)

The miq_rexml.rb file is only using the `XmlHelpers.validate_attrs` method, which is internally two separate methods. Since I do not know of an activesupport method that quite does what these methods do, I simply copied them directly into the miq_rexml.rb file, and removed the circular require.

### DIFF
--- a/lib/gems/pending/util/xml/miq_rexml.rb
+++ b/lib/gems/pending/util/xml/miq_rexml.rb
@@ -278,7 +278,7 @@ module REXML
     end
 
     def validate_attrs(hash)
-      return nil if h.nil?
+      return nil if hash.nil?
       hash.each_with_object({}) { |(k, v), h| h[k.to_s] = remove_invalid_chars(v.to_s.encode('UTF-8', :undef => :replace, :invalid => :replace, :replace => '')) }
     end
   end
--- a/lib/gems/pending/util/xml/miq_rexml.rb
+++ b/lib/gems/pending/util/xml/miq_rexml.rb
@@ -2,7 +2,6 @@ require 'time'
 require 'rexml/document'
 require 'util/miq-encode'
 require 'util/xml/xml_hash'
-require 'util/xml/xml_utils'
 require 'util/xml/xml_diff'
 require 'util/xml/xml_patch'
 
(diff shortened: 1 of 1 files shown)


djberg96:
@jerryk55 or @roliveri Is this ok?

jrafanie:
> The miq_rexml.rb file is only using the `XmlHelpers.validate_attrs` method, which is internally two separate methods. Since I do not know of an activesupport method that quite does what these methods do, I simply copied them directly into the miq_rexml.rb file, and removed the circular require.

Why do we need to copy them?  Can they be moved to a central location to be used by everyone?

jrafanie:
There's no need to mix code style/linter stuff with logical changes like the require and the copied methods.  It's ok here because the change is small but it was confusing as it adds noise to a simple change.  In the future, you can open followup PRs to clean up style issues.

djberg96:
> Why do we need to copy them? Can they be moved to a central location to be used by everyone?

Isn't `xml/xml_utils` already a central location? If you have a better suggestion for untangling this circular require, I'm all ears.

djberg96:
Actually, I wonder if just replacing `require` with `require_relative` would fix this. Will check.

jrafanie:
> > Why do we need to copy them? Can they be moved to a central location to be used by everyone?
> 
> Isn't `xml/xml_utils` already a central location? If you have a better suggestion for untangling this circular require, I'm all ears.

I was commenting about copying it and not deleting the original.  I'm not understanding if we're copying the methods, why it needs to live in 2 locations?  Maybe I'm misunderstanding but I'd rather have in one place if we can.

djberg96:
@jrafanie I still don't know what you mean by a move. What is your proposed solution?

jrafanie:
@djberg96 depends... what are you trying to solve?  If it's to remove the circular dependency, why is utils requiring things outside?  Maybe things can be moved to utils so it's a standalone and has no outside dependencies, and then you can require utils from the others with no worry about a cycle.

Note, I never understood the layout of these files so it might not be easy to do.

PR #485, opened 2020-07-22
[WIP] Remove miq-hash_struct.rb

djberg96:
With https://github.com/ManageIQ/manageiq/pull/20374, I think we can remove this.

Edit: NM for now, several providers use it.


Fryguy:
This is used in a LOT of places - https://github.com/search?q=org%3AManageIQ+MiqHashStruct&type=Code

PR #487, opened 2020-08-04
Remove miq-uuid.rb and specs

djberg96:
`MiqUUID.clean_guid` was replaced with `Digest::UUID.clean` and updated in core so this PR removes it from gems-pending.

Followup to https://github.com/ManageIQ/more_core_extensions/pull/81 and https://github.com/ManageIQ/manageiq/pull/20403


Fryguy:
Seems you missed a few spots

djberg96:
@Fryguy Looks like there was a single spec file that required it, but wasn't actually using it. Not sure why it was there. Anyway, fixed.

kbrock:
I couldn't find in any Miq repo. LGTM

PR #506, opened 2021-01-22
Remove miq-metakit.rb

djberg96:
There is no mk4rb gem. This file doesn't work and hasn't for about 14 years.


Fryguy:
FYI, all of virtfs is archived now.

PR #508, opened 2021-01-24
don't define duplicate getter for logger

d-m-u:
the getter is on line 28

i don't think we should create duplicate named methods

### DIFF
--- a/lib/gems/pending/util/mount/miq_generic_mount_session.rb
+++ b/lib/gems/pending/util/mount/miq_generic_mount_session.rb
@@ -16,7 +16,8 @@ class MiqGenericMountSession < MiqFileStorage::Interface
 
   class NoSuchFileOrDirectory < RuntimeError; end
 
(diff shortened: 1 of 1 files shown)


lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -16,7 +16,8 @@ class MiqGenericMountSession < MiqFileStorage::Interface
 
   class NoSuchFileOrDirectory < RuntimeError; end
 
-  attr_accessor :settings, :mnt_point, :logger
+  attr_accessor :settings, :mnt_point
+  attr_reader :logger
```
Fryguy, line 20: Line 28 is the _getter_ not the setter.  Think you meant to make this `attr_writer`?

Fryguy (approved):

djberg96:
LGTM 👍🏻

PR #509, opened 2021-01-25
Remove miq-unicode.rb

djberg96:
Remove miq-unicode.rb from gems-pending as part of the cleanup effort.

The main user of this was manageiq-smartstate, and that was reimplemented (using refinements) at https://github.com/ManageIQ/manageiq-smartstate/pull/149. The other callers were virtfs-xxx repos, and those were archived.


Fryguy:
I see a caller in https://github.com/ManageIQ/manageiq-smartstate/blob/23a23fec3db89859e280d9cc449946605543b2e6/lib/metadata/util/win32/Win32Accounts.rb#L627-L645, but there's no require of miq_unicode in that file.  That method does not have any test coverage.

Fryguy:
Also seeing references in this repo directly:

- [ ] https://github.com/ManageIQ/manageiq-gems-pending/blob/c2adf1f657fb8162e60df7896314596f4fc344b7/lib/gems/pending/util/win32/miq-powershell.rb#L6 + https://github.com/ManageIQ/manageiq-gems-pending/blob/c2adf1f657fb8162e60df7896314596f4fc344b7/lib/gems/pending/util/win32/miq-powershell.rb#L97

djberg96:
Ah, dammit, missed those. Will update.

Fryguy:
For the one in this repo, it seems like only a single refernce, so I'd be ok with inlining the conversion

djberg96:
@Fryguy ok, updated the inline one. I'll take a look at the core code now.

lib/gems/pending/util/win32/miq-powershell.rb
```diff
@@ -94,7 +93,8 @@ def self.is_error_object?(xml)
   end
 
   def self.run_script(script)
-    encoded_command = MIQEncode.encode(script.AsciiToUtf8.Utf8ToUnicode, false)
+    string = script.dup.force_encoding("ISO-8859-1").force_encoding("UTF-8").encode("UTF-16LE")
```
Fryguy, line 96: The intermediate encoding seems pointless.  Let's do

```suggestion
    string = script.dup.force_encoding("ISO-8859-1").encode("UTF-16LE")
```

djberg96: I was sort of baffled by it, but didn't want to risk breaking something. Ok, will update.

Fryguy: It was probably to avoid a new `AsciiToUnicode` method since the other 2 were already available.

djberg96:
> I see a caller in https://github.com/ManageIQ/manageiq-smartstate/blob/23a23fec3db89859e280d9cc449946605543b2e6/lib/metadata/util/win32/Win32Accounts.rb#L627-L645, but there's no require of miq_unicode in that file.

It's picked up via `disk/modules/MSCommon.rb`

djberg96:
> https://github.com/ManageIQ/manageiq-smartstate/blob/23a23fec3db89859e280d9cc449946605543b2e6/lib/metadata/VmConfig/cfgConfig.rb#L8

Hah, well, how was this working before? I'll add it.

djberg96:
@Fryguy https://github.com/ManageIQ/manageiq-smartstate/pull/152

PR #510, opened 2021-01-27
Remove MiqUtil.runcmd in favor of AwesomeSpawn

Fryguy:
Part of #231 

Depends on:

- [x] https://github.com/ManageIQ/manageiq/pull/20987
- [x] https://github.com/ManageIQ/manageiq-smartstate/pull/153
  - [x] release of the gem
  - [x] updated Gemfile reference to the newest release - https://github.com/ManageIQ/manageiq/pull/20995

@chessbyte Please review


lib/gems/pending/util/mount/miq_generic_mount_session.rb
```diff
@@ -29,18 +29,18 @@ def logger
     @logger ||= $log.nil? ? :: Logger.new(STDOUT) : $log
   end
 
-  def runcmd(cmd_str, args = {})
-    self.class.runcmd(cmd_str, args)
+  def run_command(cmd_str, args = {})
```
Fryguy, line 32: I renamed this because it makes it easier to search for the callers when the callers aren't named the same.

Also, in a follow up PR I'm going to remove this anyway, because the only callers to run_command are internal, and _ALL_ of them only do `mount` or `unmount`.  IMO, we should then just create a private `mount` method and a private `unmount` method.

lib/gems/pending/util/postgres_admin.rb
```diff
@@ -245,24 +245,29 @@ def self.vacuum(opts)
...
   end
 
-  def self.runcmd(cmd_str, opts, args)
-    runcmd_with_logging(cmd_str, opts, combine_command_args(opts, args))
+  def self.run_command(cmd_str, opts, args)
+    run_command_with_logging(cmd_str, opts, combine_command_args(opts, args))
   end
 
-  def self.runcmd_with_logging(cmd_str, opts, params = {})
+  def self.run_command_with_logging(cmd_str, opts, params = {})
```
Fryguy, line 261: Similarly here I renamed it for clarity.  There are exactly 2 callers in core specs that stub runcmd_with_logging, which is why i need the alias below, but I'll remove that in a follow up.

lib/gems/pending/util/win32/miq-powershell.rb
```diff
@@ -28,7 +28,7 @@ def self.execute(ps_command)
     ps_exec = exepath
     command = "start /wait cmd /c #{ps_exec} #{ps_command}"
     $log.debug "PowerShell: Running command: [#{command}]" if $log
-    ret = MiqUtil.runcmd(command).chomp
+    ret = AwesomeSpawn.run!(command, :combined_output => true).output.chomp
```
Fryguy, line 31: This one (and the next one) are super-interesting, because there is no way for them to possibly work.  `start` and `taskkill` are Windows commands, and so can't be run by `AwesomeSpawn` nor could they have been run by `MiqUtil.runcmd`.  I converted it for now, but it's actually broken.  I _think_ the correct way to run it is with `WMIHelper.connectServer.runProcess(command, true)`, but I'm not 100% sure and I have no idea how to test it.

lib/gems/pending/util/miq-ipmi.rb
```diff
@@ -185,10 +185,11 @@ def run_command(ipmi_cmd, *args)
...
     begin
-      return MiqUtil.runcmd("ipmitool", :params => command_args)
-    rescue => err
-      return err.to_s if continue_on_error == true && $?.exitstatus == 1
-      raise "Command:<#{command_line}> exited with status:<#{$?.exitstatus}>\nCommand output:\n#{err}"
+      AwesomeSpawn.run!("ipmitool", :params => command_args, :combined_output => true).output
+    rescue AwesomeSpawn::CommandResultError => err
+      return err.to_s if continue_on_error == true && err.exit_status == 1
+
+      raise "Command:<#{err.command_line}> exited with status:<#{err.exit_status}>\nCommand output:\n#{err}"
```
Fryguy, line 192: Also note this whole error handling was a bunch of bugs beforehand. `command_line` standalone is an non-existent variable name, and `$?` was never set by `runcmd`

lib/gems/pending/util/win32/miq-powershell.rb
chessbyte: I believe this code was used a long time ago when we had agents on clients for various reasons.

Searching for MiqPowerShell yields [3 results](https://github.com/search?q=org%3AManageIQ+MiqPowerShell&type=code), only one of which is an [actual caller](https://github.com/ManageIQ/manageiq-providers-scvmm/blob/4165805bf7a9b72476349dcdbaff76eaf56dbbde/app/models/manageiq/providers/microsoft/infra_manager/powershell.rb#L39).  So, I believe we will be able to delete (in a future PR?) most of the guts of MiqPowerShell.  If we try really hard, we can probably delete the whole module as the only remaining use for it is to convert strings. ;-)

lib/gems/pending/util/miq-ipmi.rb
```diff
@@ -185,10 +185,12 @@ def run_command(ipmi_cmd, *args)
...
-      return MiqUtil.runcmd("ipmitool", :params => command_args)
-    rescue => err
-      return err.to_s if continue_on_error == true && $?.exitstatus == 1
-      raise "Command:<#{command_line}> exited with status:<#{$?.exitstatus}>\nCommand output:\n#{err}"
+      AwesomeSpawn.run!("ipmitool", :params => command_args, :combined_output => true).output
+    rescue AwesomeSpawn::CommandResultError => err
+      result = err.result
+      return err.to_s if continue_on_error == true && result.exit_status == 1
+
+      raise "Command:<#{result.command_line}> exited with status:<#{result.exit_status}>\nCommand output:\n#{err}"
```
Fryguy, line 193: Also note this whole error handling was a bunch of bugs beforehand. command_line standalone is an non-existent variable name, and $? was never set by runcmd

chessbyte (approved):

lib/gems/pending/util/win32/miq-powershell.rb
Fryguy: Yeah, I'm not certain we really need it anymore either...hoping we can kill it in a future PR.

lib/gems/pending/util/mount/miq_generic_mount_session.rb
Fryguy: https://github.com/ManageIQ/manageiq-gems-pending/pull/511 .  Note I built that PR indepedently, so one of these PRs will conflict with the other, but that's ok.

NickLaMuro:
😍 

@Fryguy I didn't even know this was being done!  100% approve with this change, and is a big tech debt win!  🎉

PR #516, opened 2021-03-05
Remove  PostgresAdmin

NickLaMuro:
Part of https://github.com/ManageIQ/manageiq-appliance_console/pull/153

Also checks a box in https://github.com/ManageIQ/manageiq-gems-pending/issues/231

Dependencies
------------

- [x] ManageIQ/manageiq-appliance_console#153
- [x] PR removing `evm_database_ops.rb` from core
  - ~~https://github.com/ManageIQ/manageiq/pull/21360~~
  - https://github.com/ManageIQ/manageiq/pull/21384


Rakefile
```diff
@@ -3,26 +3,6 @@ require 'rspec/core/rake_task'
...
-  # via the helper.
-  pg_runner = File.join("spec", "postgres_runner_helper.rb")
-  t.rspec_opts = "-r #{File.expand_path(pg_runner, __dir__)}"
-end
-
-task :default do
-  Rake::Task["spec#{':ci' if ENV['CI']}"].invoke
-  Rake::Task["test"].invoke
-end
+task :default => [:spec, :test]
```
NickLaMuro, line 8: Note:  I don't know if the `:test` tests have ever been running, but was not able to get them to run locally.  This should still work this way, but might be worth looking into to determine if these tests are still needed, or should just be converted to specs for simplicity.

NickLaMuro:
Note:  This one should have everything removed, but I would rather wait to merge this one until we merge https://github.com/ManageIQ/manageiq-appliance_console/pull/153 as well as merging code that will eventually remove the need for `lib/evm_database_ops.rb` and friends in Core (which rely on `PostgresAdmin`).

Fryguy:
> `+1 −1,219`

😍

Fryguy:
I'm seeing 2 more references to this on core.

- https://github.com/ManageIQ/manageiq/blob/master/app/models/miq_server/log_management.rb#L74-L76
- https://github.com/ManageIQ/manageiq/blob/master/lib/tasks/evm_dba.rake#L74-L100

PR #524, opened 2021-08-20
Loosen manageiq-password to < 2

Fryguy:
Must be merged before https://github.com/ManageIQ/manageiq/pull/21178

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "fog-openstack",           "~> 0.3"
   s.add_runtime_dependency "handsoap",                "=  0.2.5.5"
   s.add_runtime_dependency "linux_admin",             "~> 2.0"
-  s.add_runtime_dependency "manageiq-password",       "~> 0.3"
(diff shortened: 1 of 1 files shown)


manageiq-gems-pending.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "fog-openstack",           "~> 0.3"
   s.add_runtime_dependency "handsoap",                "=  0.2.5.5"
   s.add_runtime_dependency "linux_admin",             "~> 2.0"
-  s.add_runtime_dependency "manageiq-password",       "~> 0.3"
+  s.add_runtime_dependency "manageiq-password",       "~> 1.0"
```
bdunne, line 31: Can this be loosened to `< 2`?

Fryguy: Is that not effectively the same thing?

Fryguy: Oh i see - that would also allow the 0.3 in this particular repo - interesting idea.

bdunne: Yeah, this doesn't appear to depend on anything that broke in the major version change, so it would ease the upgrade process

Fryguy:
@bdunne Updated

PR #532, opened 2021-12-14
Fix simplecov handling for codeclimate via GitHub Actions

Fryguy:
When running the tests, we need to make sure the CC_TEST_REPORTER_ID
value is present. This allows SimpleCov to generate the JSON payload
that is needed for Code Climate, as SimpleCov has special handling when
that env var is present.

Also fix calling of minitest based tests

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,5 +1,5 @@
 require "simplecov"
-SimpleCov.start { command_name "spec" }
+SimpleCov.start
 
 $LOAD_PATH.unshift File.expand_path('../../lib', __FILE__)
 require 'manageiq-gems-pending'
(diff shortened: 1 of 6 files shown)


spec/spec_helper.rb
```diff
@@ -1,5 +1,5 @@
 require "simplecov"
-SimpleCov.start { command_name "spec" }
```
bdunne, line 2: I thought we needed these to prevent one test suite from overwriting all of the data from the previous one

Fryguy: Perhaps in the past, but the current version of SimpleCov automatically creates two sections named "RSpec" and "Unit tests" for rpsec and minitest respectively.  This keeps the defaults, and then SimpleCov automatically merges them together into the JSON payload when it sees `CC_TEST_REPORTER_ID` as an env var.

PR #533, opened 2021-12-14
Move minitest based tests to rspec

Fryguy:
@bdunne Please review.  This is to drop the extra dependency but also not make it complicated that there are multiple ways to run tests.


Fryguy:
Also fixed some rubocops

PR #536, opened 2021-12-15
Remove handsoap

Fryguy:
handsoap was removed from the gemspec, but accidentally left in the
Gemfile.  This removes it completely.

@agrare Please review.


agrare (approved):

kbrock:
Yay. removing hand soap is good clean fun!

PR #544, opened 2022-05-26
Bump nokogiri to >= 1.13.6

agrare:
https://github.com/ManageIQ/amazon_ssa_support/issues/77

Don't want to say this will resolve that yet, since amazon_ssa_support pulls in 1.13.6 already so whitesource seems out of date


jrafanie:
Should this be backported?  I think so, let's review in backport meeting.  Added najdorf/yes?

agrare:
Najdorf already has the latest version: https://github.com/ManageIQ/manageiq/blob/najdorf/Gemfile.lock.release#L870  But not against backporting it to make it "official"
I'm not sure why whitesource flagged this as 1.13.3 when we clearly were using the latest in the 1.13 branch

jrafanie:
> I'm not sure why whitesource flagged this as 1.13.3 when we clearly were using the latest in the 1.13 branch

I guess it depends how it's scanning it.  If the Gemfile/gemspec allows for vulnerable (old) versions vs. an optimistic that tries to install the latest "allowed" versions.

jrafanie:
I'm fine with not allowing known vulnerable versions via the Gemfile/gemspec.

PR #549, opened 2022-09-30
New exception type created for vm resize

alizapeikes:


Fryguy (approved):
Seems ok to me - @agrare ?

PR #551, opened 2022-12-10
 Add zeitwerk loader and teach it about this gem

jrafanie:
Depends on:
- [x] Core using classic loader (just this PR) - green cross repo: https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/800 

* specify load path directories
* inform of organizational directories that aren't namespaces
* specific files that must be required and not autoloaded
* custom inflectors where we deviate from convention
* remove requires for autoload-able files (caution must be taken when we need to monkey patch as we need to require or we may never get our monkey patch loaded)
* use require_relative for ones that must be manually required

TODO:

- [x] Extract commits that are completely backward compatible https://github.com/ManageIQ/manageiq-gems-pending/pull/559
- [x] Keep WIP the remaining parts that depend on having zeitwerk
- [x] Merge PRs removing dependency on lib/manageiq-gems-pending.rb file
  - [x] https://github.com/ManageIQ/manageiq/pull/22548
  - [x] https://github.com/ManageIQ/amazon_ssa_support/pull/96

### DIFF
--- a/lib/gems/pending/util/miq-xml.rb
+++ b/lib/gems/pending/util/miq-xml.rb
@@ -1,9 +1,5 @@
 require 'time'
-require 'util/xml/miq_rexml'
-require 'util/xml/xml_hash'
-require 'util/miq-encode'
-require 'util/xml/xml_diff'
-require 'util/xml/xml_patch'
+require 'util/xml/miq_rexml' # we must monkey patch rexml early
 
 class MiqXml
   MIQ_XML_VERSION = 2.1 # Refactor Nokogiri handling
--- a/lib/gems/pending/util/miq-xml.rb
+++ b/lib/gems/pending/util/miq-xml.rb
@@ -42,7 +42,7 @@ class MiqXml
       when :xmlhash
         XmlHash::Document
       when :nokogiri
-        require 'util/xml/miq_nokogiri'
+        require_relative 'xml/miq_nokogiri'
         @nokogiri = true
         Nokogiri::XML::Document
       else
--- a/lib/manageiq-gems-pending.rb
+++ b/lib/manageiq-gems-pending.rb
@@ -1,1 +1,52 @@
 require 'manageiq/gems/pending'
+
+require "zeitwerk"
+loader = Zeitwerk::Loader.for_gem(warn_on_extra_files: false)
+loader.ignore(__FILE__)
+
+# For straight requires from here
+loader.push_dir("#{__dir__}/gems/pending/util")
+
+# This tells the loader to not expect Mount or ObjectStorage namspaces...
+loader.collapse("#{__dir__}/gems/pending/util/mount")
+loader.collapse("#{__dir__}/gems/pending/util/object_storage")
+loader.collapse("#{__dir__}/gems/pending/util/win32")
+loader.collapse("#{__dir__}/gems/pending/util/xml")
+
+# These files skip zeitwerk and must be required manually
+##### TODO: check for requires outside of this gem to these files and util/ in requires
+loader.ignore("#{__dir__}/gems/pending/util/miq-extensions.rb")       # loader file, no class
+loader.ignore("#{__dir__}/gems/pending/util/require_with_logging.rb") # file has no class ;-)
+loader.ignore("#{__dir__}/gems/pending/util/xml/miq_nokogiri.rb")     # monkey patch, so no class
+loader.ignore("#{__dir__}/gems/pending/util/xml/xml_utils.rb")        # multiple classes in one file
+
+# These inflectors teach zeitwerk our naming convention here
+loader.inflector.inflect(
+  "manageiq"  => "ManageIQ",
+  "version"   => "VERSION",  # TODO: why is this needed, for_gem is supposed to do this
+)
+
+# These inflectors are teaching zeitwerk when we fail at conventions.
+# It expects downcased and underscored file names (not hyphens).
+# TODO: These could be fixed but require changing client code to use the new constant.
+loader.inflector.inflect(
+  "miq-encode"      => "MIQEncode",
+  "miq-exception"   => "MiqException",
+  "miq-hash_struct" => "MiqHashStruct",
+  "miq-ipmi"        => "MiqIPMI",
+  "miq-powershell"  => "MiqPowerShell",
+  "miq-process"     => "MiqProcess",
+  "miq-system"      => "MiqSystem",
+  "miq-wmi"         => "WMIHelper",     # WAT
+  "miq-wmi-linux"   => "WmiLinux",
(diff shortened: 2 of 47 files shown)


lib/manageiq-gems-pending.rb
```diff
@@ -1 +1,52 @@
...
+  "miq-xml"         => "MiqXml",
+  "miq_rexml"       => "MIQRexml",
+  "xml_diff"        => "MiqXmlDiff",
+  "xml_patch"       => "MiqXmlPatch",
+)
+
+loader.setup
+
+# TODO: consider turning on eager_load and only disable really slow/fat requires
+# loader.eager_load
```
jrafanie, line 52: This is what I enabled locally to get everything to fail so I could then fix or ignore.

manageiq-gems-pending.gemspec
```diff
@@ -30,8 +30,10 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "nokogiri",                "~> 1.13", ">= 1.13.6"
   s.add_runtime_dependency "sys-proctable",           "~> 1.2.5"
   s.add_runtime_dependency "sys-uname",               "~> 1.2.1"
+  s.add_runtime_dependency "win32ole",                "~> 1.8.8" # this gem was extracted in ruby 3 - required if we use wmi on windows
   s.add_runtime_dependency "winrm",                   "~> 2.1"
   s.add_runtime_dependency "winrm-elevated",          "~> 1.0.1"
+  s.add_runtime_dependency "zeitwerk"
```
jrafanie, line 36: probably want to lock this down to a minor version

lib/gems/pending/util/miq_winrm.rb
```diff
@@ -10,6 +7,10 @@ def initialize
     @shell           = nil
     @elevated_runner = nil
     require 'uri'
+
+    # These are slow/bulky (15 MB+) requires so delay them until they're needed
+    require 'winrm'
+    require 'winrm-elevated'
```
Fryguy, line 13: Can you upstream this change separately?

lib/manageiq-gems-pending.rb
```diff
@@ -1 +1,54 @@
...
+
+require "zeitwerk"
+loader = Zeitwerk::Loader.for_gem(warn_on_extra_files: false)
+loader.ignore(__FILE__)
+
+# For straight requires from here
+loader.push_dir("#{__dir__}/gems/pending/util")
+
+# The collapsed directories are not proper namespaces and only for organizing.
+# This tells Zeiwerk to not expect Mount or ObjectStorage namspaces...
```
Fryguy, line 11:
```suggestion
# This tells Zeitwerk to not expect Mount or ObjectStorage namspaces...
```

:trollface:

lib/gems/pending/util/miq_winrm.rb
jrafanie: yeah, I was playing with eager load and found/remembered this.

jrafanie: For future me, I hope this whole file can be deleted because of https://github.com/ManageIQ/manageiq/pull/22279

The only callers of `MiqWinRM` in my bundle are this file and it's test and two files in smartstate.

```
% rg -M 300 "MiqWinRM" ./ `gem env home`
/Users/joerafaniello/.gem/ruby/2.7.6/bundler/gems/manageiq-gems-pending-d4ced36efe52/spec/util/miq_winrm_spec.rb
3:describe MiqWinRM do
8:    @winrm    = MiqWinRM.new

/Users/joerafaniello/.gem/ruby/2.7.6/bundler/gems/manageiq-gems-pending-d4ced36efe52/lib/gems/pending/util/miq_winrm.rb
4:class MiqWinRM

/Users/joerafaniello/.gem/ruby/2.7.6/gems/manageiq-smartstate-0.8.1/lib/Scvmm/miq_hyperv_disk.rb
23:    @winrm     = MiqWinRM.new

/Users/joerafaniello/.gem/ruby/2.7.6/gems/manageiq-smartstate-0.8.1/lib/Scvmm/miq_scvmm_vm_ssa_info.rb
13:    @winrm       = MiqWinRM.new
```

cc @agrare   🙏

jrafanie: Ok, I removed the commit to delay load winrm as it's being deleted in https://github.com/ManageIQ/manageiq-gems-pending/pull/556

lib/gems/pending/util/miq-xml.rb
```diff
@@ -41,7 +41,7 @@ def self.xml_document(xmlClass)
       when :xmlhash
```
kbrock, line 41: tangent: do we have enough coverage to know if we can drop `:xmlhash` or `:rexml`?

kbrock: I had assumed we only used this in a few rare places, but besides smart state (and specs), looks like we use it in kubernetes, vmware, ui classic (only in 1-2 files each) along with a dozen core models.

kbrock:
We used to require `util/` for all of these. Happy that you are getting rid of these.

lib/gems/pending/util/miq_winrm.rb
jrafanie: Resolving

jrafanie:
@miq-bot cross-repo-test /all

lib/manageiq/gems/pending/zeitwerk.rb
```diff
@@ -0,0 +1,50 @@
...
+# This tells the loader to not expect Mount or ObjectStorage namspaces...
+loader.collapse("#{lib}/gems/pending/util/mount")
+loader.collapse("#{lib}/gems/pending/util/object_storage")
+loader.collapse("#{lib}/gems/pending/util/win32")
+loader.collapse("#{lib}/gems/pending/util/xml")
+
+# These files skip zeitwerk and must be required manually
+##### TODO: check for requires outside of this gem to these files and util/ in requires
+loader.ignore("#{lib}/gems/pending/util/miq-extensions.rb")       # loader file, no class
+loader.ignore("#{lib}/gems/pending/util/require_with_logging.rb") # file has no class ;-)
```
Fryguy, line 18: We should move this out into a diagnostics thing or maybe move it into core.

jrafanie:
ok, this is green other than the already existing constants: 
image


The cross repo is green other than ui service

jrafanie:
I think this is ready to go.  Once in and backported to quinteros, we can decide about about the core PR.  They can be independent upgrades.

agrare (approved):

Fryguy:
Skipping backport to `quinteros`, because it is already in the branch.

Issue #403, opened 2018-11-06
Move MiqFileStorage into core and provider specific gems


NickLaMuro:
Related to the large issue, https://github.com/ManageIQ/manageiq-gems-pending/issues/231 (the `mount/*` checkbox), though has grown in scope since that issue was created do to the addition of the following classes/libs being created/reworked:

- `lib/gems/pending/util/miq_file_storage.rb`
- `lib/gems/pending/util/miq_object_storage.rb`
- `lib/gems/pending/util/miq_ftp_lib.rb`
- `lib/gems/pending/util/object_storage/*.rb`
  - `miq_ftp_storage.rb`
  - `miq_s3_storage.rb`
  - `miq_swift_storage.rb`

Really, this is all `manageiq` related, and should just live there in `lib/` (or a core gem... if we ever go that route).  The provider specific ones should live in their respective provider repos, and get loaded in as part of the rake tasks to determine what functionality is available then (instead of the hard coding that is done currently).

Task items
----------

* [ ] Move into "core" ( https://github.com/ManageIQ/manageiq/pull/19547 , https://github.com/ManageIQ/manageiq-gems-pending/pull/454 )
* [ ] Create a plugin system for providers
* [ ] Extract provider code to their respective repos

NickLaMuro:
Updated full file list of files to be extracted (see PR description):

- `lib/miq_file_storage.rb`
- `lib/miq_ftp_lib.rb`
- `lib/miq_object_storage.rb`
- `lib/mount/*.rb`
	* `lib/mount/miq_generic_mount_session.rb`
	* `lib/mount/miq_glusterfs_session.rb`
	* `lib/mount/miq_local_mount_session.rb`
	* `lib/mount/miq_nfs_session.rb`
	* `lib/mount/miq_smb_session.rb`
- `lib/object_storage/*.rb`
	* `lib/object_storage/miq_ftp_storage.rb`
	* `lib/object_storage/miq_s3_storage.rb`
	* `lib/object_storage/miq_swift_storage.rb`
- `spec/lib/miq_file_storage_spec.rb`
- `spec/lib/miq_ftp_lib_spec.rb`
- `spec/lib/miq_object_storage_spec.rb`
- `spec/lib/mount/miq_generic_mount_session_spec.rb`
- `spec/lib/mount/miq_local_mount_session_spec.rb`
- `spec/lib/object_storage/miq_ftp_storage_spec.rb`
[...]

PR #558, opened 2023-05-02
Update fog-openstack to v1.0

agrare:
fog-openstack released v1.0 in Jan 2019 we should update to the latest version, this was also keeping us on an older version of fog-core

Co-depends:
* https://github.com/ManageIQ/manageiq-providers-openstack/pull/734

### DIFF
--- a/lib/gems/pending/util/object_storage/miq_swift_storage.rb
+++ b/lib/gems/pending/util/object_storage/miq_swift_storage.rb
@@ -95,7 +95,7 @@ class MiqSwiftStorage < MiqObjectStorage
   end
 
   #
-  # Some calls to Fog::Storage::OpenStack::Directories#get will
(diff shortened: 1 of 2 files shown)


agrare:
Actually do we still use `MiqFileStorage` for anything @Fryguy @jrafanie ?  It seems like it was removed from `evm_database_ops.rb` in https://github.com/ManageIQ/manageiq/pull/21384 and not used in the "new" implementation in appliance_console https://github.com/ManageIQ/manageiq-appliance_console/pull/160

kbrock:
we were able to successfully drop `MiqFileStorage` for database backups.
I thought there was a use for it in the ui.

may lead to more info:
- https://github.com/ManageIQ/manageiq-gems-pending/issues/403
- https://github.com/ManageIQ/manageiq/pull/19547

PR #559, opened 2023-05-03
Zeitwerk friendly changes

jrafanie:
Extracted from https://github.com/ManageIQ/manageiq-gems-pending/pull/551

Cross repo is 💚 https://github.com/ManageIQ/manageiq-cross_repo-tests/pull/789

### DIFF
--- a/lib/gems/pending/util/miq-extensions.rb
+++ b/lib/gems/pending/util/miq-extensions.rb
@@ -1,5 +1,3 @@
 require 'active_support/inflector'
 require 'manageiq-gems-pending'
 require 'more_core_extensions/all'
-
-Dir.glob(File.expand_path(File.join(File.dirname(__FILE__), "extensions", "*.rb"))) { |f| require_relative "extensions/#{File.basename(f, ".*")}" }
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "nokogiri",                "~> 1.14", ">= 1.14.3"
   s.add_runtime_dependency "sys-proctable",           "~> 1.2.5"
   s.add_runtime_dependency "sys-uname",               "~> 1.2.1"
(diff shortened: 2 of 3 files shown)


jrafanie:
@miq-bot cross-repo-test /core

manageiq-gems-pending.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "nokogiri",                "~> 1.14", ">= 1.14.3"
   s.add_runtime_dependency "sys-proctable",           "~> 1.2.5"
   s.add_runtime_dependency "sys-uname",               "~> 1.2.1"
+  s.add_runtime_dependency "win32ole",                "~> 1.8.8" # this gem was extracted in ruby 3 - required if we use wmi on windows
```
kbrock, line 32: I remember we used to use this for smartproxy on windows

Do we still need this?

kbrock: I can't fine `WmiMswin` used anywhere in our code base

kbrock: ok, `vm.rb` uses wmi gook.
think there is some :fire: opportunity here

jrafanie: yeah, I think so too... it's a definite squirrel though

lib/gems/pending/util/miq-extensions.rb
```diff
@@ -1,5 +1,3 @@
 require 'active_support/inflector'
 require 'manageiq-gems-pending'
 require 'more_core_extensions/all'
-
-Dir.glob(File.expand_path(File.join(File.dirname(__FILE__), "extensions", "*.rb"))) { |f| require_relative "extensions/#{File.basename(f, ".*")}" }
```
Fryguy, line 5: This concerns me as what will load the extensions?

Fryguy: make me wonder if we need this file _at all_

kbrock: I was also diving into understanding the `win32ole` gem and what files we actually need in here.

But I had to eject and remember that this gem is a utility drawer and we will get there when we get there.

PR #570, opened 2024-01-12
Allow rails 7 gems in gemspec

jrafanie:
Allow rails 7 gems in gemspec

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,13 @@ gemspec
 
 plugin "bundler-inject", "~> 1.1"
 require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundler-inject") rescue nil
+
+minimum_version =
+  case ENV['TEST_RAILS_VERSION']
+  when "7.0"
+    "~>7.0.8"
+  else
+    "~>6.1.4"
+  end
+
+gem "activesupport", minimum_version
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "> 2.4"
 
-  s.add_runtime_dependency "activesupport",           "~> 6.0"
+  s.add_runtime_dependency "activesupport",           ">=6.0"
   s.add_runtime_dependency "awesome_spawn",           "~> 1.5"
   s.add_runtime_dependency "aws-sdk-s3",              "~> 1.0"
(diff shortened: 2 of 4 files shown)


manageiq-gems-pending.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
 
   s.required_ruby_version = "> 2.4"
 
-  s.add_runtime_dependency "activesupport",           "~> 6.0"
+  s.add_runtime_dependency "activesupport",           ">=6.0"
```
jrafanie, line 23: I'm debating just keeping this more open ended, allowing us to just test the versions we care about but also making it easier to start on 7.1 without having to update this gem immediately.  The application controls the upper bound for the version constraint so I don't think we need to do `~>` here.

jrafanie:
@agrare since you work on some gems that consume gems pending, what do you think of this PR and most importantly, the note above about loosening the activesupport dependency so the consuming application or gem can dictate the upper bound on the version constraint?

manageiq-gems-pending.gemspec
agrare: Yeah that works for me, we could always add an upper bound if we discover a hard incompatibility with a newer rails version (like we found with manageiq-loggers and Broadcast)

agrare (approved):

Gemfile
```diff
@@ -5,3 +5,13 @@ gemspec
...
+
+minimum_version =
+  case ENV['TEST_RAILS_VERSION']
+  when "7.0"
+    "~>7.0.8"
+  else
+    "~>6.1.4"
+  end
+
+gem "activesupport", minimum_version
```
kbrock, line 17: Do we want to use the TEST_RAILS_VERSION in our test?

Don't like how cryptic my example is but...

```ruby
gem "activesupport", "~>#{ENV.fetch("TEST_RAILS_VERSION", "6.1")}.0"
```

I'm not sure that "~>6.1.0" vs "~>6.1.4" will make a great difference.

jrafanie: yeah, I had that previously in another PR... since we want to just set minimum versions, and default to 6.1, this was the easiest way to do it.  If you set `TEST_RAILS_VERSION=6.1.6.1` your version (and my original rework, that looks liked yours) didn't work correctly.

This version does what we want minimally... set absolute minimum, prevent arbitrary "MINOR" updates, and also defaults to rails 6.1 instead of relying on the gemspec's maximum version constraint to set the default.  This allows us to use rails 7 here without needing bump the gemspec each time.

PR #579, opened 2024-05-06
Print the memory consumed and loaded files for each require line

jrafanie:
This will help us find client libraries or other requires that end of loading a lot of code and consuming time and memory.

It looks like this:

```
      1  <+  | net/ssh (0.207687 seconds) (3964 KB) (118 features loaded)
      1  >   | net/scp/errors
      1  <+  | net/scp/errors (0.000798 seconds) (16 KB) (1 features loaded)
      1  >   | net/scp/upload
      2  >   | | net/scp/errors
      2  <-  | | net/scp/errors (0.000201 seconds) (0 KB) (0 features loaded)
      1  <+  | net/scp/upload (0.001059 seconds) (65 KB) (1 features loaded)
      1  >   | net/scp/download
      2  >   | | net/scp/errors
      2  <-  | | net/scp/errors (0.000222 seconds) (0 KB) (0 features loaded)
      1  <+  | net/scp/download (0.001043 seconds) (98 KB) (1 features loaded)
      0  <+  net/scp (0.217726 seconds) (4194 KB) (122 features loaded)
```

Note, I used this for https://github.com/ManageIQ/manageiq-providers-amazon/pull/856

### DIFF
--- a/lib/gems/pending/util/require_with_logging.rb
+++ b/lib/gems/pending/util/require_with_logging.rb
@@ -10,6 +10,8 @@ $req_log.sync = true
 
 $req_depth = 0
 
+require 'sys-proctable'
+
 module Kernel
   private
 
@@ -24,9 +26,20 @@ module Kernel
   }
   REQ_LOG_TREE = "| "
 
-  def log_require(path, mode, timing = nil)
+  def require_with_metrics
+    initial_memory = Sys::ProcTable.ps(:pid => Process.pid).rss
+    initial_loaded = $LOADED_FEATURES.length
+
+    result = yield
+
+    memory_after = Sys::ProcTable.ps(:pid => Process.pid).rss
+    loaded_after = $LOADED_FEATURES.length
+    return result, (memory_after - initial_memory) / 1000, (loaded_after - initial_loaded)
+  end
+
+  def log_require(path, mode, timing = nil, memory_consumed = nil, features_loaded = nil)
     $req_depth -= 1 if mode == true || mode == false || mode == :fail
-    $req_log.puts "#{$req_depth.to_s.rjust(3)}  #{REQ_LOG_OPERS[mode]}  #{REQ_LOG_TREE * $req_depth}#{path.inspect[1..-2]}#{timing.nil? ? '' : " (#{"%.6f" % timing})"}"
+    $req_log.puts "#{$req_depth.to_s.rjust(3)}  #{REQ_LOG_OPERS[mode]}  #{REQ_LOG_TREE * $req_depth}#{path.inspect[1..-2]}#{timing.nil? ? '' : " (#{"%.6f seconds" % timing})"}#{memory_consumed.nil? ? '' : " (#{memory_consumed} KB)"}#{features_loaded.nil? ? '' : " (#{features_loaded} features loaded)"}"
     $req_depth += 1 if mode == :enter || mode == :reenter
   end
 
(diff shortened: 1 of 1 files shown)


lib/gems/pending/util/require_with_logging.rb
```diff
@@ -24,22 +26,33 @@ module Kernel
...
   REQ_LOG_TREE = "| "
 
-  def log_require(path, mode, timing = nil)
+  def require_with_metrics
+    initial_memory = Sys::ProcTable.ps(:pid => Process.pid).rss
+    initial_loaded = $LOADED_FEATURES.length
+
+    result = yield
+
+    memory_after = Sys::ProcTable.ps(:pid => Process.pid).rss
```
Fryguy, line 35: Should there be some GC calls in here to be more accurate?  Or is this relatively accurate since the calls are so close to one another?

lib/gems/pending/util/require_with_logging.rb
```diff
@@ -24,22 +26,33 @@ module Kernel
...
+    loaded_after = $LOADED_FEATURES.length
+    return result, (memory_after - initial_memory), (loaded_after - initial_loaded)
+  end
+
+  def log_require(path, mode, timing = nil, memory_consumed = nil, files_loaded = nil)
     $req_depth -= 1 if mode == true || mode == false || mode == :fail
-    $req_log.puts "#{$req_depth.to_s.rjust(3)}  #{REQ_LOG_OPERS[mode]}  #{REQ_LOG_TREE * $req_depth}#{path.inspect[1..-2]}#{timing.nil? ? '' : " (#{"%.6f" % timing})"}"
+    $req_log.puts "#{$req_depth.to_s.rjust(3)}  #{REQ_LOG_OPERS[mode]}  #{REQ_LOG_TREE * $req_depth}#{path.inspect[1..-2]}#{timing.nil? ? '' : " (#{"%.6f" % timing})"}#{memory_consumed.nil? ? '' : " (#{memory_consumed})"}#{files_loaded.nil? ? '' : " (#{files_loaded})"}"
```
Fryguy, line 42: I think this needs units now to distinguish the numbers

```suggestion
    $req_log.puts "#{$req_depth.to_s.rjust(3)}  #{REQ_LOG_OPERS[mode]}  #{REQ_LOG_TREE * $req_depth}#{path.inspect[1..-2]}#{timing.nil? ? '' : " (#{"%.6f" % timing}s)"}#{memory_consumed.nil? ? '' : " (#{memory_consumed / 1024.0}MB)"}#{files_loaded.nil? ? '' : " (#{files_loaded} files)"}"
```

Fryguy: I noticed also that this says "files loaded", but earlier it's loaded features - is that the same thing?  Should it say "features loaded" instead?

lib/gems/pending/util/require_with_logging.rb
jrafanie: I considered it but I think it's fairly obvious memory snapshots are supposed to be estimates.  I can if you think it's worth it.

lib/gems/pending/util/require_with_logging.rb
jrafanie: sure, I can do that

lib/gems/pending/util/require_with_logging.rb
Fryguy: Nah it's fine - just thought I'd raise the discussion.

jrafanie:
@Fryguy thanks for the good ideas, I added the units and normalized on "features" instead of files. Note, KB seems more useful to get requires in the hundreds of KB as 500 KB is still pretty awful so it would be useful to grep with 3 or numbers followed by KB as that will capture it without having to parse decimals.  Doing this with numbers like 0.5 MB and 5 MB would be an extra pattern to match.

lib/gems/pending/util/require_with_logging.rb
jrafanie: Note, I tried doing `GC.start` at the start of the method and it's super slow.  I think it's better to assume the memory number is a very rough estimate.

lib/gems/pending/util/require_with_logging.rb
jrafanie: ok, done, used KB, see below.

Fryguy (approved):

PR #581, opened 2024-07-18
Fix MiqXml handling of BOM + handle CVE-2024-39908

Fryguy:
The BOM handling code was incorrectly using `"\xC3\xAF\xC2\xBB\xC2\xBF"`, which is just `"ï»¿"` in UTF-8 encoding. The correct BOM marker is `"\xEF\xBB\xBF"` in binary encoding (which happens to be `"ï»¿"` when encoded as Windows-1252 encoding).

```ruby
> s = "\xC3\xAF\xC2\xBB\xC2\xBF" # UTF-8 encoded
=> "ï»¿"
> s.encode("WINDOWS-1252")
=> "\xEF\xBB\xBF"

> s = "\xEF\xBB\xBF" # UTF-8 encoded
=> "" # BOM is there but hidden in display
> s.force_encoding("WINDOWS-1252")
=> "\xEF\xBB\xBF"
```

I'm not sure why the original code used `"\xC3\xAF\xC2\xBB\xC2\xBF"`, but this code was originally written in 2012 in 9c26f1e4 around the Ruby 1.8/1.9 time frame and was added to handle various encoding problems with Ruby 1.9 (which is when Encoding was introduced).

There have been other commits such as 8209fb6e that have removed some of this older questionable code which should work in modern versions of REXML.

This commit adds some tests ensuring the correct BOM works as expected, and modifies the original BOM test to treat those leading characters more like top-level text nodes, which is how they were being interpreted in modern REXML anyway.

Note that in REXML 3.3.2, the `.write` method no longer prints out extraneous top-level text nodes. REXML 3.3.2 added an Exception for top-level text nodes _after_ the root node, but curiously did not raise an Exception for text nodes _before_ the root node. Nokogiri in STRICT mode will fail on both, so it's curious that REXML does not. Even so, I expect in the future REXML will fail on these leading text nodes, so we may have to revisit this at that time.

This commit also enforces the minimum version to REXML 3.3.2. This is partly because of CVE-2024-39908, and partly because it complicated the test code to support multiple versions of REXML.

@kbrock @jrafanie Please review.

### DIFF
--- a/lib/gems/pending/util/xml/miq_rexml.rb
+++ b/lib/gems/pending/util/xml/miq_rexml.rb
@@ -279,15 +279,6 @@ module REXML
 
     def self.load(data)
       REXML::Document.new(data)
-    rescue => err
-      if err.class == ::Encoding::CompatibilityError
-        data_utf8 = data.dup.force_encoding('UTF-8')
-        # Check for UTF-8 BOM and remove
-        data_utf8 = data_utf8[3..-1] if data_utf8[0, 3] == "\xC3\xAF\xC2\xBB\xC2\xBF".force_encoding("UTF-8")
-        REXML::Document.new(data_utf8)
-      else
-        raise
-      end
     end
 
     def self.loadFile(filename)
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "more_core_extensions",    "~> 4.4"
   s.add_runtime_dependency "net-ftp",                 "~> 0.1.2"
   s.add_runtime_dependency "nokogiri",                "~> 1.14", ">= 1.14.3"
+  s.add_runtime_dependency "rexml",                   ">= 3.3.2"
   s.add_runtime_dependency "sys-proctable",           "~> 1.2.5"
   s.add_runtime_dependency "sys-uname",               "~> 1.2.1"
   s.add_runtime_dependency "win32ole",                "~> 1.8.8" # this gem was extracted in ruby 3 - required if we use wmi on windows
--- a/spec/util/miq-xml_spec.rb
+++ b/spec/util/miq-xml_spec.rb
@@ -6,15 +6,34 @@ describe MiqXml do
     expect(xml.root.elements[1].attributes['attr1']).to eq(attr_string)
   end
 
-  it "handles loaded document with UTF-8 BOM" do
+  it "handles loaded document with top-level text nodes" do
     attr_string = "test string"
-    doc_text = "\xC3\xAF\xC2\xBB\xC2\xBF<test><element_1 attr1='#{attr_string}'/></test>"
+    doc_text = "XXX<test><element_1 attr1='#{attr_string}'/></test>"
+
     xml = MiqXml.load(doc_text)
     expect(xml.root.elements[1].attributes['attr1']).to eq(attr_string)
 
-    expect(xml.to_s[0, 3]).to eq("\xC3\xAF\xC2\xBB\xC2\xBF")
+    expect(xml.to_s).to start_with("XXX<test>")
+
     xml.write(xml_str = '', 1)
-    expect(xml_str[0, 3]).to eq("\xC3\xAF\xC2\xBB\xC2\xBF")
+    expect(xml_str).to start_with("\n<test>")
+  end
+
+  it "handles loaded document with UTF-8 BOM" do
+    bom = "\xEF\xBB\xBF".force_encoding("US-ASCII")
+    attr_string = "test string"
+    doc_text = "#{bom}<test><element_1 attr1='#{attr_string}'/></test>".force_encoding("US-ASCII")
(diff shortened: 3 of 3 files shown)


spec/util/miq-xml_spec.rb
```diff
@@ -6,15 +6,38 @@
...
+    expect(xml.root.elements[1].attributes['attr1']).to eq(attr_string)
+
+    expect(xml.to_s).to start_with("XXX<test>")
+
+    xml.write(xml_str = '', 1)
+    if REXML::VERSION < "3.3.2"
+      expect(xml_str).to start_with("XXX\n<test>")
+    else
+      expect(xml_str).to start_with("\n<test>")
+    end
```
Fryguy, line 23: I'm not sure if we even need this part of the test - the real test is ensuring we can load the document and find the attribute (lines 13-14).

So, I kept it and used a version conditional here, which is gross, but I'm not sure I care too much.

Alternatively, we can _force_ the minimum version of REXML in the gemspec.  There's a CVE anyway, so we may want to do that.  However, when I tried to do that locally I was getting bundler errors and I'm not sure why.

Fryguy: Actually, now that I wrote that - I'm gonna change the .gemspec anyway to set a minimum, because of the CVE.

spec/util/miq-xml_spec.rb
```diff
@@ -6,15 +6,34 @@
...
+    attr_string = "test string"
+    doc_text = "XXX<test><element_1 attr1='#{attr_string}'/></test>"
+
+    xml = MiqXml.load(doc_text)
+    expect(xml.root.elements[1].attributes['attr1']).to eq(attr_string)
+
+    expect(xml.to_s).to start_with("XXX<test>")
+
+    xml.write(xml_str = '', 1)
+    expect(xml_str).to start_with("\n<test>")
```
Fryguy, line 19: This is what I mentioned in the OP about `.write` no longer printing extraneous top-level text nodes.  I think this is because of https://github.com/ruby/rexml/pull/164 (although there are a number of parsing changes [between 3.3.1 and 3.3.2](https://github.com/ruby/rexml/compare/v3.3.1...v3.3.2))

manageiq-gems-pending.gemspec
```diff
@@ -28,6 +28,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "more_core_extensions",    "~> 4.4"
   s.add_runtime_dependency "net-ftp",                 "~> 0.1.2"
   s.add_runtime_dependency "nokogiri",                "~> 1.14", ">= 1.14.3"
+  s.add_runtime_dependency "rexml",                   ">= 3.3.2"
```
Fryguy, line 31: I'd rather not add it as a dependency, because we want to learn towards Nokogiri anyway, but if you look at the [Gemfile.lock.release](https://github.com/ManageIQ/manageiq/blob/7aa4afff3e9614c919fd03f2f6066fbafe4dfcb9/Gemfile.lock.release#L1147), rexml is already there as a dependency of a number of libraries, including google-apis-core, so we might as well enforce a minimum here.

lib/gems/pending/util/xml/miq_rexml.rb
```diff
@@ -279,15 +279,6 @@ def extendXmlDiff; end
...
       REXML::Document.new(data)
-    rescue => err
-      if err.class == ::Encoding::CompatibilityError
-        data_utf8 = data.dup.force_encoding('UTF-8')
-        # Check for UTF-8 BOM and remove
-        data_utf8 = data_utf8[3..-1] if data_utf8[0, 3] == "\xC3\xAF\xC2\xBB\xC2\xBF".force_encoding("UTF-8")
-        REXML::Document.new(data_utf8)
-      else
-        raise
-      end
```
jrafanie, line 290: I'm glad to see this go and I'm glad git blame points to someone else.  ❤️

spec/util/miq-xml_spec.rb
Fryguy: I also commented here about the weird leading blank link - That's probably a bug https://github.com/ruby/rexml/pull/164#discussion_r1683552024

manageiq-gems-pending.gemspec
jrafanie: It's complaining about 

```
metadata['rubygems_mfa_required'] must be set to 'true'.
```

that can be done in a followup

jrafanie (approved):

Fryguy:
Backported to radjabov via merge of master into radjabov branch

Fryguy:
Backported to `quinteros` in commit c9ea1311787617c9e64dc22d3a85721a370ce19d.

```text
commit c9ea1311787617c9e64dc22d3a85721a370ce19d
Date:   Thu Jul 18 18:02:49 2024 -0400

    Merge pull request #581 from Fryguy/fix_improper_bom_handling
    
    Fix MiqXml handling of BOM + handle CVE-2024-39908
    
    (cherry picked from commit 3c0c7bb60759b333aea3dabd6f6f50919b626ae4)
```

PR #582, opened 2024-08-01
Upgrade rexml to 3.3.3+ for CVE-2024-41123 and CVE-2024-41946

Fryguy:
In #581 I wrote

> I expect in the future REXML will fail on these leading text nodes,
> so we may have to revisit this at that time.

and that has now happened in rexml 3.3.3 (https://github.com/ruby/rexml/pull/184), so this commit also removes the spec for leading text nodes, since they can no longer be parsed. The BOM test ensures that the leading BOM marker is not treated as a text node.

@jrafanie Please review.

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |s|
   s.add_runtime_dependency "more_core_extensions",    "~> 4.4"
   s.add_runtime_dependency "net-ftp",                 "~> 0.1.2"
   s.add_runtime_dependency "nokogiri",                "~> 1.14", ">= 1.14.3"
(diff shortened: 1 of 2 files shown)


Fryguy:
Backported to `quinteros` in commit 89c10f4de5f43c625f166beab7b1f1583773d0ae.

```text
commit 89c10f4de5f43c625f166beab7b1f1583773d0ae
Date:   Fri Aug 2 10:39:12 2024 -0400

    Merge pull request #582 from Fryguy/rexml_3_3_4
    
    Upgrade rexml to 3.3.3+ for CVE-2024-41123 and CVE-2024-41946
    
    (cherry picked from commit aa3f5b2f93c3f55426f010ff751ce7728abbc748)
```

PR #592, opened 2024-09-24
Fix broken test matrix  and add ruby 3.2 and 3.3

jrafanie:
### TEST_RAILS_VERSION wasn't set in ENV so matrix was broken
Rails 7.2 will not install in versions of ruby less than 3.1.

Because we weren't setting the ENV variable properly, the test matrix
was not correctly testing against the correct rails version and passing.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -12,13 +12,18 @@ jobs:
     strategy:
       matrix:
         ruby-version:
-        - '3.0'
         - '3.1'
         rails-version:
         - '7.0'
         - '7.1'
         - '7.2'
+        include:
+        - ruby-version: '3.0'
+          rails-version: '7.0'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -12,13 +12,20 @@ jobs:
...
+        - '3.3'
         rails-version:
         - '7.0'
         - '7.1'
         - '7.2'
+        include:
+        - ruby-version: '3.0'
+          rails-version: '7.0'
+        - ruby-version: '3.0'
+          rails-version: '7.1'
```
Fryguy, line 26: I merged, but I think an exclude is better - I'll send up a PR

jrafanie: I like to be inclusive 🤣

PR #593, opened 2024-09-24
Switch the CI matrix from include to exclude

Fryguy:
@jrafanie Please review.


jrafanie (approved):
I like how you merged so you can get a net negative followup PR.  Fine, take your -1 lines.  🤣

kbrock:
Quite a few gems I've been using have gone this way
Or just dropped ruby all together from the matrix
and then 1-off'd the older ruby/rails combinations

PR #610, opened 2025-07-01
nailing down rack version in development

kbrock:
Our style development tool requires rack.
This was triggering some Mend issues.

This change nails down the version

### DIFF
--- a/manageiq-gems-pending.gemspec
+++ b/manageiq-gems-pending.gemspec
@@ -43,6 +43,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ftpd",                      "~> 2.1.0"
   s.add_development_dependency "manageiq-style",            ">= 1.5.4"
 
+  s.add_development_dependency "rack",                      "~> 3.1.16" # this ensures manageiq-style's rack requirement is safe
   s.add_development_dependency "rake",                      ">= 12.3.3"
   s.add_development_dependency "rspec",                     "~> 3.13"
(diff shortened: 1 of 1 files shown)


kbrock:
update:
- rubocop, changed order of gem declarations

manageiq-gems-pending.gemspec
```diff
@@ -37,6 +37,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "ftpd",                      "~> 2.1.0"
   s.add_development_dependency "manageiq-style",            ">= 1.5.4"
 
+  s.add_development_dependency "rack",                      "~> 3.1.16" # this ensures manageiq-style's rack requirement is safe
```
Fryguy, line 40: I thought manageiq core didn't work with rack 3 - maybe put another way - when I bundle update in core I get rack 2

Fryguy:
@miq-bot cross-repo-test ManageIQ/manageiq

manageiq-gems-pending.gemspec
kbrock: This is a development dependency.

This `rack` requirement is only installed when `bundle install` inside `manageiq-gems-pending`. These dependencies are not transient to other projects.

When `bundle update` in `manageiq` core, it will ignore the `rack` requirement from here.
Of course, it will get a def rack requirement from `manageiq-style`, and a runtime requirement from rails. I assume mend won't be too happy in other scanned projects with no `rack` requirement.

PR #620, opened 2025-11-05
Cleanup mount ping

Fryguy:
- Removing Settings lookup which hasn't worked in ages. When then Settings were redesigned this code basically never ran anymore because ::VMDB::CONFIG was renamed to ::Vmdb::Config. Additionally, we've removed the log/collection/ping_* settings in the process of removing log collection.
- Removed use of class vars in favor of ivars
- Modernized ping_depot_options, and made it private

@jrafanie Please review.

Ultimately, I want to drop mount session stuff, but until PxeServer drops the need for it, it's has to stick around.


jrafanie (approved):
Bug circa Date:   Tue Jul 20 22:45:04 2010 +0000.

Nice find.  Yeah, let's kill this broken logic for now and remove it entirely later.

PR #623, opened 2025-12-04
Allow bundler v4.0

agrare:
Allow newer bundler versions while still preventing `2.2.10`

TODO do we need a runtime dependency here at all?  Won't core's bundler dependency supersede this anyway?


Fryguy:
@miq-bot cross-repo-test /all

Fryguy:
This and schema are standalone repos, so some of the dependencies have to be duplicated. Not sure why bundler though, cause I didn't think this repo had bundler specific things.

PR #634, opened 2026-03-25
Drop Rails <7.2, and update to Rails 8

Fryguy:
@jrafanie Please review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,12 +8,12 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 
 minimum_version =
   case ENV['TEST_RAILS_VERSION']
-  when "7.0"
-    "~>7.0.8"
-  when "7.1"
-    "~>7.1.4"
   when "7.2"
     "~>7.2.3"
+  when "8.0"
+    "~>8.0.5"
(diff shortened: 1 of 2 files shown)


Gemfile
```diff
@@ -8,12 +8,10 @@ require File.join(Bundler::Plugin.index.load_paths("bundler-inject")[0], "bundle
 
 minimum_version =
   case ENV['TEST_RAILS_VERSION']
-  when "7.0"
-    "~>7.0.8"
-  when "7.1"
-    "~>7.1.4"
-  when "7.2"
-    "~>7.2.3"
+  when "8.0"
```
jrafanie, line 13: Maybe we should keep current and prior version 7.2.x and 8.0.x until we move to 8.0.x and 8.1.x.

Fryguy: I assumed we were good on 8.0 since we shipped it, but I can add back 7.2

jrafanie: yeah, we're good... I'm fine with it... it just looked weird to have the case statement for one version but I guess it works when we add 8.1

Fryguy: Done

jrafanie (approved):

PR #654, opened 2026-07-02
Set concurrent-ruby minimum to 1.3.7 for CVE-2026-54904

Fryguy:
@agrare Please review.

Should close #650, #651, #652

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,3 +17,6 @@ minimum_version =
   end
 
 gem "activesupport", minimum_version
+
+# security fixes for indirect dependencies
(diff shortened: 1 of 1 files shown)


agrare (approved):

agrare:
@Fryguy should we do this in the core Gemfile also or is that not needed?  `activesupport` only declares `concurrent-ruby (~> 1.0, >= 1.3.1)` and everything else is `~> 1.0`

Fryguy:
yes, Mend is complaining there too. We need this PR anyway though because gems-pending is an "indenpendent" library from core.

